PR #198, opened 2020-07-10
Fixes unit tests to reflect mnesia base change

j4mcs:
This closes #

## Local Testing

```
```

```
$ kubectx dev-bunny
$ make destroy deploy-dev
$ make system-tests
```

### DIFF
--- a/internal/resource/statefulset_test.go
+++ b/internal/resource/statefulset_test.go
@@ -1254,10 +1254,6 @@ var _ = Describe("StatefulSet", func() {
 							Value: "/opt/rabbitmq-secret/username",
 						},
 						corev1.EnvVar{
-							Name:  "RABBITMQ_MNESIA_BASE",
-							Value: "/var/lib/rabbitmq/db",
-						},
-						corev1.EnvVar{
 							Name: "MY_POD_NAME",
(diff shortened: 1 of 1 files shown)


internal/resource/statefulset_test.go
```diff
@@ -1306,7 +1302,7 @@ var _ = Describe("StatefulSet", func() {
 						},
 						corev1.VolumeMount{
 							Name:      "persistence",
-							MountPath: "/var/lib/rabbitmq/db/",
```
michaelklishin, line 1309: FWIW I think `/var/lib/rabbitmq/db/` is a better default. Mnesia will be replaced one day.

Zerpet: Ideally we should just mount `/var/lib/rabbitmq` and give r/w permissions to rabbitmq user. Our challenge is that we already mount another volume in `/var/lib/rabbitmq` for the Erlang cookie and IIRC, we can't mount two different volumes in the same path.

j4mcs: Going to merge because this is just a fix of the tests and the change you're discussing is already merged. But will add it to our next sync up agenda so we don't lose track

PR #205, opened 2020-07-15
Update issue templates

Zerpet:
Remove boilerplate from bug report. Ask for relevant information to reproduce a bug in software.


Zerpet:
> Looks good, and a big improvement over asking for browser version.
> 
> Do we want to ask what container image they're using for RabbitMQ?

Oops, I just read your comment after merging 😅 We could explicitly ask for that, although I think that for issues where the RMQ image used is relevant, we will spot it in the YAML manifests provided as part of the bug report.

PR #206, opened 2020-07-16
Set RabbitMQ cluster name to CR name

### LINKED ISSUE #185: Configure RabbitMQ Cluster name to match Custom Resource Name
The default cluster name in RabbitMQ is the name of the first node to form the cluster. In the Kubernetes context, this name is very long and it breaks the layout of RabbitMQ Management UI, plus it doesn't read very human friendly.

Use the configuration `cluster_name` in `rabbitmq.conf` to pre-configure the cluster name and match, at least, the custom resource name. We could potentially include the namespace as well.

N/A

**Additional context**
This is possible since https://github.com/rabbitmq/rabbitmq-server/pull/2075 got merged in August 2019. Any version of RabbitMQ 3.8.x should allow this property.

Zerpet:
This closes #185 

## Summary Of Changes
The `cluster_name` property will be set with value matching the CR name. Before this change, the cluster name was the name of the first node who formed the cluster, which resulted in very long names.

## Local Testing

Unit and integration tests 🟢

```
```

Deployed the Operator with this changes, created a 3-node cluster and verified in Management UI that cluster name matches the CR name.

### DIFF
--- a/internal/resource/configmap.go
+++ b/internal/resource/configmap.go
@@ -61,10 +61,12 @@ func (builder *ServerConfigMapBuilder) Update(object runtime.Object) error {
 	configMap := object.(*corev1.ConfigMap)
 	configMap.Labels = metadata.GetLabels(builder.Instance.Name, builder.Instance.Labels)
 	configMap.Annotations = metadata.ReconcileAndFilterAnnotations(configMap.GetAnnotations(), builder.Instance.Annotations)
+
 	if configMap.Data == nil {
 		configMap.Data = make(map[string]string)
 	}
-	configMap.Data["rabbitmq.conf"] = defaultRabbitmqConf
+
+	configMap.Data["rabbitmq.conf"] = defaultRabbitmqConf + fmt.Sprintln("cluster_name = "+builder.Instance.Name)
 	if builder.Instance.TLSEnabled() {
 		configMap.Data["rabbitmq.conf"] = configMap.Data["rabbitmq.conf"] + defaultTLSConf
 	}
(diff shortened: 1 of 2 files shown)


internal/resource/configmap.go
```diff
@@ -61,10 +61,12 @@ func (builder *ServerConfigMapBuilder) Update(object runtime.Object) error {
 	configMap := object.(*corev1.ConfigMap)
 	configMap.Labels = metadata.GetLabels(builder.Instance.Name, builder.Instance.Labels)
 	configMap.Annotations = metadata.ReconcileAndFilterAnnotations(configMap.GetAnnotations(), builder.Instance.Annotations)
+
 	if configMap.Data == nil {
 		configMap.Data = make(map[string]string)
 	}
-	configMap.Data["rabbitmq.conf"] = defaultRabbitmqConf
+
+	configMap.Data["rabbitmq.conf"] = defaultRabbitmqConf + fmt.Sprintln("cluster_name = "+builder.Instance.Name)
```
coro, line 69: Is the default config likely to expand in the future, in similar ways to this: requiring builder-specific data, and then appending config? If so, it may be worth moving the `const`s above to their own private method, e.g.

```go
func defaultRabbitmqConf(string instanceName) string { fmt.Sprintf(...) }
```

j4mcs: If we refactor, might as well use a string [Builder](https://www.calhoun.io/concatenating-and-building-strings-in-go/) as well

Zerpet: I'm not sure how likely is that we expand the default conf using builder data. In any case, I think it's good to have an option to do so. I will make this refactors 👌

Zerpet: Implemented these changes in https://github.com/rabbitmq/cluster-operator/commit/a0025394f94d993443f7c8d865fdbb74902b7dd9 🚀

PR #209, opened 2020-07-17
Add example about import definitions

### LINKED ISSUE #104: Create a workflow to automatically import RabbitMQ definitions
As a RabbitMQ user responsible for provisioning RMQ clusters, I want to have a mechanism to deploy RabbitMQ with a common set of vhosts, users, exchanges and queues defined. I want these definitions to be imported at [RabbitMQ startup](https://www.rabbitmq.com/definitions.html#import-on-boot).

I want to provide a schema definition in a ConfigMap of my choice, mount this ConfigMap in RabbitMQ as a volume and make RabbitMQ automatically load my definitions at startup.

Perhaps use a Secret instead of a ConfigMap, if the definitions file contains passwords.

**Additional context**
It is possible to [manually import/export](https://www.rabbitmq.com/backup.html#definitions-export) the definitions file using the Management UI. This process is not automatic on startup.

It is also possible to use [rabbitmqctl](https://www.rabbitmq.com/management.html#load-definitions) to export/import definitions. The way to import the definitions automatically is via `load_definitions` in `rabbitmq.conf`.

Original idea from @ferozjilla, requested by a person in our [public channel](https://pivotal.slack.com/archives/CJ3D7E63Y/p1588293686101200).

ChunyiLyu:
This closes #104

## Summary Of Changes
- add an example that imports definitions json file at node boot leveraging statefulSet Override, and `additionalConfig`

## Additional Context

Exported definitions contains all broker objects, including users and permissions. When importing definitions, the admin user of the cluster will be the one recorded from the definitions file, and not the admin user that the operator generates at rabbitmqcluster creation. I've tried deleting the `users` field from the definitions file when importing, and that resulted in the rabbitmqcluster starting with **no user** at all. This is an interesting side effect that the k8s admin secret is not used when importing definitions, and it would be better if there is an easy way for users to signal that so the operator can skip creating admin secret for such rabbitmqclusters.

## Local Testing

Have verified that the example works against a gke cluster. After importing the definitions, queues and exchanges from previous queue all showed up and including the previous admin user.

### DIFF
--- a/docs/examples/import-definitions/README.md
+++ b/docs/examples/import-definitions/README.md
@@ -1,6 +1,6 @@
 # Import Definitions Example
 
-You can import definitions, which contain definitions of all broker objects from a local file at Node Boot. (Learn more about export and import rabbitmq definitions)[https://www.rabbitmq.com/definitions.html#import].
+You can import definitions, which contain definitions of all broker objects from a local file at Node Boot. [Learn more about export and import rabbitmq definitions](https://www.rabbitmq.com/definitions.html#import).
 
 You can save the export definitions json file in a ConfigMap object like so
 
@@ -14,6 +14,11 @@ data:
 # exported definitions context
 ```
 
-and then leverage the StatefulSet Override to mount this additional ConfigMap `definitions` to your rabbitmqcluster instance. Check out `rabbitmq.yaml` as an example.
+or
+```bash
+kubectl create configmap definitions --from-file='def.json=/my/path/to/definitions.json'
+```
+
+Then, leverage the StatefulSet Override to mount this additional ConfigMap `definitions` to your rabbitmqcluster instance. Check out `rabbitmq.yaml` as an example.
 
 Keep in mind that exported definitions contain all broker objects, including users. This means that the admin user credentials will be imported from the definitions, and will not be the one which is generated at the creation of the deployment as a kubernetes secret object.
--- a/docs/examples/import-definitions/README.md
+++ b/docs/examples/import-definitions/README.md
@@ -1,8 +1,8 @@
 # Import Definitions Example
 
-You can import definitions, which contains definitions of all broker objects from a local file at Node Boot. (Learn more about export and import rabbitmq definitions)[https://www.rabbitmq.com/definitions.html#import].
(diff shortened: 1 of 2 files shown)


docs/examples/import-definitions/README.md
```diff
@@ -0,0 +1,19 @@
+# Import Definitions Example
+
+You can import definitions, which contains definitions of all broker objects from a local file at Node Boot. (Learn more about export and import rabbitmq definitions)[https://www.rabbitmq.com/definitions.html#import].
```
j4mcs, line 3: 'contain' not 'contains'

docs/examples/import-definitions/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+You can save the export definitions json file in a ConfigMap object in like so
```
j4mcs, line 5: 'in a ConfigMap object like so'

j4mcs: actually, the Google technical writing guide would probably suggest removing 'like so' altogether

"You can save the export definitions json file in a ConfigMap object:"

docs/examples/import-definitions/rabbitmq.yaml
```diff
@@ -0,0 +1,23 @@
...
+  replicas: 1
+  override:
+    statefulSet:
+      spec:
+        template:
+          spec:
+            containers:
+            - name: rabbitmq
+              volumeMounts:
+              - mountPath: /tmp/definitions
```
j4mcs, line 15: I think if we were to be suggesting an actual path it should be in `/var` as this file could be expected to grow.

The rabbitMQ docs opt for `/path/to/exported/definitions.json` which saves us having to bother about Linux conventions

ChunyiLyu: Sure

docs/examples/import-definitions/README.md
```diff
@@ -0,0 +1,19 @@
+# Import Definitions Example
+
+You can import definitions, which contain definitions of all broker objects from a local file at Node Boot. (Learn more about export and import rabbitmq definitions)[https://www.rabbitmq.com/definitions.html#import].
```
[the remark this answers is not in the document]
ChunyiLyu: Updated!

PR #213, opened 2020-07-21
Add client service override `spec.override.clientService`

### LINKED ISSUE #142: Add ingress service override

**Additional context**

ChunyiLyu:
This closes #142

## Summary Of Changes

- adds `spec.override.ClientService`
- tests at unit and integration level
- example in `docs/example/additionalPorts`

## Additional Context

Example manifest which opens an additional port on client service and the `rabbitmq` container, using both the client service and the sts override 

```yaml
kind: RabbitmqCluster
metadata:
  name: rabbit
spec:
  override:
    clientService:
      spec:
        ports:
        - name: additional-port # adds an additional port on the client service
          protocol: TCP
          port: 12345
    statefulSetOverwrite:
      spec:
        template:
          spec:
            containers:
              - name: rabbitmq
                ports:
                  - containerPort: 12345 # opens an additional port on the rabbitmq server container
                    name: additional-port
                    protocol: TCP
```

## Local Testing

Have run unit and integration tests.


docs/examples/additionalPorts/rabbitmq.yaml
```diff
@@ -0,0 +1,22 @@
...
+                                                                  name: rabbit
+                                                                spec:
+                                                                  override:
+                                                                    clientService:
+                                                                      spec:
+                                                                        ports:
+                                                                          - name: additional-port # adds an additional port on the client service
+                                                                            protocol: TCP
+                                                                            port: 12345
+                                                                    statefulSetOverwrite:
```
j4mcs, line 13: Isn't this just `statefulSet`?

internal/resource/client_service_test.go
```diff
@@ -402,6 +402,127 @@ var _ = Context("ClientServices", func() {
...
+				instance.Spec.Override.ClientService = &rabbitmqv1beta1.ClientService{
+					Spec: &corev1.ServiceSpec{
+						Type: corev1.ServiceTypeNodePort,
+					},
+				}
+
+				err := serviceBuilder.Update(svc)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(svc.Spec.Type).To(Equal(corev1.ServiceTypeNodePort))
+			})
```
Zerpet, line 524: Would this test cover the case when we try to override a default port e.g. amqp? Or should we add a specific test for that?

internal/resource/client_service.go
```diff
@@ -56,13 +58,51 @@ func (builder *ClientServiceBuilder) Update(object runtime.Object) error {
...
+	}
+
 	if err := controllerutil.SetControllerReference(builder.Instance, service, builder.Scheme); err != nil {
 		return fmt.Errorf("failed setting controller reference: %v", err)
 	}
 
 	return nil
 }
 
+func (builder *ClientServiceBuilder) applySvcOverride(svc *corev1.Service, override *rabbitmqv1beta1.ClientService) error {
```
Zerpet, line 74: Does this need to be a pointer function? I'm not sure if it makes much of a difference in terms of performance or "correctness"?

What do you think @j4mcs ?

internal/resource/client_service_test.go
ChunyiLyu: Any default port (amqp, empd, and management that the operator set) cannot be overwritten by patching because the svc.spec.ports is a list with a patch merge key of `port`. Override can only give the port a new name or a new target port, but not delete it. I'm not sure about the value of testing that behavior here.

ChunyiLyu:
@Zerpet both the svc and the sts override are introducing tons of possibilities for people to misconfigure their deployment, like updating the svc name as you pointed. I am leaning towards not to add any more validations since a failed update will be `aborted` and it won't impact the existing rabbitmq cluster. If we were to say that we should add more validations to prevent people from providing configurations that's invalid, there are much more safeguards we need to add than just this one, and I would rather do it in a separate track of work, to make sure that all validations are done in a consistent and holistic way (it could be a standalone admission webhook)

`spec.replicas` is added to the example manifest. Good catch 👍

internal/resource/client_service.go
ChunyiLyu: ah it does not need to be, I will update the function signature.

internal/resource/client_service_test.go
Zerpet: That makes complete sense, no need to test that case. We should document this tho in our docs, as it's reasonable to think that the merge key is the port name.

Zerpet:
The problem with allowing the fields `name` and `namespace` in `EmbeddedObjectMeta` is that even `create` operations will fail because the mutate function in `CreateOrUpdate()` is not allowed to modify the name/namespace; it's not specific only to an update. I don't see what's the benefit or value of having a name and namespace in the embedded object meta, given that any customisation of those fields will result in a irreconcilable CR. 

I agree that adding safeguards for every possible situation is not realistic because of the endless possibilities. I'm considering that we should delete the fields `name` and `namespace` from the `EmbeddedObjectMeta`. Do they have any functional relevance in the code? 
At a first glance, they only seem to be used in tests and helper functions.

ChunyiLyu:
@Zerpet 
> I agree that adding safeguards for every possible situation is not realistic because of the endless possibilities. I'm considering that we should delete the fields `name` and `namespace` from the `EmbeddedObjectMeta`. Do they have any functional relevance in the code?
> At a first glance, they only seem to be used in tests and helper functions.

Name and namespace are required for PVC configuration. I've added a new type with just the labels and annotations. Client service and statefulSet use the new type which makes the name and namespace not configurable.

Issue #194, opened 2020-07-09
Invstigate different Operator base images


coro:
**Is your feature request related to a problem? Please describe.**
Currently, for the operator image we use a [pinned version of Ubuntu Bionic](https://github.com/rabbitmq/cluster-operator/blob/a8465dd1be35b4f3caf3a3fb3b1c35b04a8987de/Dockerfile#L19) as the base image. While this made sense at the time, this process is no longer suitable: pinning to a particular SHA causes us to miss out on security updates to the base image.

In addition, Ubuntu Docker images are fairly large (64.2MB), compared with some other alternatives available, such as Alpine (5.57MB) or Google Distroless (1.79MB). Larger images contain much more software than the operator needs, and theoretically increases the surface area for security vulnerabilities. In contrast, larger traditional distros such as Debian & Ubuntu have larger communities maintaining them, and this may lead to faster CVE fixing than the likes of Alpine & Distroless.

**Describe the solution you'd like**
At the very least, we should bump to the latest version of the Bionic image to consume the latest fixes for the image.

We could also potentially switch to a different distro like Alpine, or go distroless altogether. Some investigation into what's possible/optimal would be useful.

**Describe alternatives you've considered**
Debian is an alternative to Ubuntu, but appears to be on par in terms of security from what I can tell (happy to stand corrected).

**Additional context**
Kubebuilder, by default, uses Google's Distroless base image in [Dockerfiles it generates](https://github.com/kubernetes-sigs/kubebuilder/blob/6e0d3d4f18ecdc335177d28e1d6cd4d7b2930242/pkg/plugin/v2/scaffolds/internal/templates/dockerfile.go#L62).

Considerations to be made on which image is used:
* Size of image
* Security profile of image
  * Frequency & severity of CVEs
  * Speed with which CVEs are typically fixed
* Whether the image provides / can provide utilities necessary to the operator (~~`useradd` & `groupadd` come to mind~~ looks like you can do this in distroless by providing a custom `/etc/groups` and `/etc/passwd` to the image)
* Performance of the operator running on the image

coro:
Looks like we also had concerns about whether distroless had timezone data & CA certs for TLS. According to [the README](https://github.com/GoogleContainerTools/distroless/blob/master/base/README.md) for the images, it has tzdata and ca-certificates.

coro:
Copying over from a Slack discussion about this:

Requirements (as far as I can tell):

[...]

PR #217, opened 2020-07-21
Bump Dockerfile to latest ubuntu

coro:
Pinning to a specific SHA no longer makes sense with no OSL process.

This addresses #194 - some work still needs to be done to see if another base image is more suitable.

## Local Testing

All of unit, integration & system tests are 💚


Zerpet:
Manually triggered the PR pipeline. If it runs through, I'll merge this PR.

PR #226, opened 2020-07-28
Allow configuration through rabbitmq-env.conf

### LINKED ISSUE #218: Add support for `/etc/rabbitmq/rabbitmq-env.conf`

Erlang runtime tuning (anything that is a vm flag), can only be configured through `rabbitmq-env.conf`

`rabbitmq-env.conf` needs to be at `/etc/rabbitmq/rabbitmq-env.conf`, and the location is **not configurable**

**Additional context**
Implementation wise it is very similar to https://github.com/rabbitmq/cluster-operator/issues/219.

coro:
This closes #218 

## Summary Of Changes

Allow a user to specify any environment variable config to be applied in /etc/rabbitmq/rabbitmq-env.conf. Note that this PR does not contain any validation of the contents of this file yet, however it can be easily tested in the controller using `sh -n /path/to/env.conf`.

## Additional Context

Similar to #221 .

## Local Testing

```
```

```
$ kubectx dev-bunny
$ make destroy deploy-dev
$ make system-tests
```


internal/resource/configmap_test.go
```diff
@@ -137,6 +137,19 @@ my-config-property-1 = better-value`
 			Expect(rabbitmqConf).To(Equal(expectedRabbitmqConf))
 		})
 
+		It("creates and populates a rabbitmq-env.conf when envConfig is provided", func() {
```
harshac, line 162: Do we have an alternative test case where EnvConfig == nil?

config/crd/bases/rabbitmq.com_rabbitmqclusters.yaml
```diff
@@ -4598,6 +4598,10 @@ spec:
                   description: Specify any rabbitmq advanced.config configurations
                   maxLength: 100000
                   type: string
+                envConfig:
+                  description: Modify to add to the rabbitmq-env.conf file. Modify this property on an existing RabbitmqCluster will trigger a StatefulSet rolling restart and will cause rabbitmq downtime.
```
mkuratczyk, line 4602: I believe the second "Modify" should actually be "Modifying"

harshac: Done

internal/resource/configmap.go
```diff
@@ -96,9 +96,14 @@ func (builder *ServerConfigMapBuilder) Update(object runtime.Object) error {
 		}
 	}
 
-	if builder.Instance.Spec.Rabbitmq.AdvancedConfig != "" {
+	if builder.Instance.Spec.Rabbitmq.AdvancedConfig == "" {
+		delete(configMap.Data, "advanced.config")
```
coro, line 100: Ooh I didn't know delete was a no-op if the key doesn't exist! 

Given this is quite a clean pattern for advanced now, would it make sense to do the same in envConfig? That way when you remove the config from the CR spec, both of the config files are removed too (rather than having an empty `rabbitmq-env.conf`)?

internal/resource/configmap.go
```diff
@@ -137,3 +134,11 @@ func AppendIfUnique(a []string, b []rabbitmqv1beta1.Plugin) []string {
 	}
 	return set
 }
+
+func updateProperty(configMapData map[string]string, key string, value string) {
```
harshac, line 138: Can use a better name. Any suggestion @coro?

coro: Can't think of anything better.

harshac: Let's merge it then.

PR #229, opened 2020-08-03
community plugin example

mkuratczyk:
New example

### DIFF
--- a/docs/examples/community-plugins/README.md
+++ b/docs/examples/community-plugins/README.md
@@ -4,6 +4,8 @@
 
 We'll keep investigating alternative options but for the time being, creating a custom image with the required plugins and just enabling them through `additionalPlugins` seems like a better idea.
 
+**NOTE**: Please raise issues related to community plugins with the community - our team does not maintain these plugins.
+
 You can deploy this example like this:
 
 ```shell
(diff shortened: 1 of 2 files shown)


docs/examples/community-plugins/README.md
```diff
@@ -0,0 +1,11 @@
+# Community Plugins Example
+
+**WARNING**: this example proves it is possible to enable a community plugin that is not included in the `rabbitmq` image. However, it relies on being able to download a file on cluster startup. If that URL is unavailable, your cluster won't start. Moreover, the downloaded file is not validated in any way and therefore you could end up loading arbitrary, potentially malicious, file into your cluster.
```
harshac, line 3: Should we also add details around how this would affect commercial support and a list of community plugin that's been approved by the Core team?

mkuratczyk: It's a good point. We don't have an official policy for community plugin support but we should have it (I'll start a separate thread for that).

gerhard: This is a good reference for community plugins that we keep track of: https://www.rabbitmq.com/community-plugins.html

There are other community plugins that that we don't keep track of - e.g. https://github.com/deadtrickster/prometheus_rabbitmq_exporter

Community plugins in the context of the commercial edition is a good private discuss, with the intent of capturing the group consensus publicly. Thanks @mkuratczyk for taking the lead on this 👍

mkuratczyk: Added a short sentence about that. We may update this later but I think it's ok for now.

PR #235, opened 2020-08-06
Check for environment variables in Makefile

ansd:
## Summary Of Changes

Provide better error messages when required environment variables are not set in Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,7 +76,7 @@ deploy-namespace-rbac:
 
 deploy: manifests deploy-namespace-rbac deploy-manager ## Deploy operator in the configured Kubernetes cluster in ~/.kube/config
 
-deploy-dev: docker-build-dev patch-dev manifests deploy-namespace-rbac docker-registry-secret deploy-manager-dev ## Deploy operator in the configured Kubernetes cluster in ~/.kube/config, with local changes
+deploy-dev: check-env-docker-credentials docker-build-dev patch-dev manifests deploy-namespace-rbac docker-registry-secret deploy-manager-dev ## Deploy operator in the configured Kubernetes cluster in ~/.kube/config, with local changes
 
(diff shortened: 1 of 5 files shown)


coro:
Thanks for the PR David! I've been meaning to test this but haven't got around to it yet; but know that I've seen it and appreciate the contribution!

One thing to note - it looks like there's a large amount of changes in this PR because `make manifests` renders all the long lines of text as multilines; I believe there's a reason we're not just checking in those changes at the moment, but I will double check.

coro:
We think it's a controller-gen versioning issue - that some versions of the tool render the multilines, and some don't. There's talk in the team to refactor our tooling to move all the controller-gen work into a container instead of relying on your local system version.

For now, could you revert the change that makes heavy changes to the `make manifests` output?

ansd:
Thank you Connor for reviewing :)
I use version controller-gen `v0.2.5` as defined in https://github.com/rabbitmq/cluster-operator/blob/0d1bd34d1798a4123dd7419a8f0da925fb5cf2ef/Makefile#L155
I reverted that commit for now.

Makefile
```diff
@@ -78,7 +78,7 @@ deploy: manifests deploy-namespace-rbac deploy-manager ## Deploy operator in the
 
 deploy-dev: docker-build-dev patch-dev manifests deploy-namespace-rbac docker-registry-secret deploy-manager-dev ## Deploy operator in the configured Kubernetes cluster in ~/.kube/config, with local changes
```
coro, line 79: I'd consider adding `check-env-docker-repo` to the start of the `deploy-dev` target so it runs before everything else; it's strictly already a prerequisite of this target in `docker-registry-secret`, but given running `make deploy-dev` is such a common operation in the team, I think it's worth it to fail fast in this case where one forgets to set the env vars. Up to you though!

ansd: Done. I like the idea.

PR #236, opened 2020-08-07
WIP: Proposal - Versioning & Release Strategy

### LINKED ISSUE #190: Explore and propose a release and versioning strategy
## Why 

Our current versioning strategy creates a release for every commit in master. This was a good strategy when we were releasing through TanzuNet and wanted to have the ability to promote a build from "the past" into a release candidate.

Given that we won't publish artefacts anymore to TanzuNet, we should re-think our release strategy and versioning. This issue will cover the exploration and write up of a proposal.

## Questions to explore

- Use [semver](https://semver.org/) for version numbers. Do we want to keep the same versioning?
- Publish every build or cut specific releases?
- Publish docker images as part of the release?
- Where to publish release artefacts?

Further considerations if/assuming we pick SemVer:

- Supportability of major and minor versions. For how long can/should we support a minor? For how long can/should we support a major?

The **outcome** of this issue should be a KEP or proposal document.

coro:
This closes #190 

## Summary Of Changes

This PR is currently a WIP. At time of writing, only the versioning sections have been filled out, and I am circing this initial proposal to get early feedback.

Please feel free to leave review comments in the PR, or to message me privately. Thank you!

### DIFF
--- a/docs/proposals/20200806-versioning-release-strategy.md
+++ b/docs/proposals/20200806-versioning-release-strategy.md
@@ -21,35 +21,39 @@ see-also:
 
 ## Table of Contents
 
-A table of contents is helpful for quickly jumping to sections of a proposal and for highlighting
-any additional information provided beyond the standard proposal template.
-[Tools for generating](https://github.com/ekalinin/github-markdown-toc) a table of contents from markdown are available.
+<!--ts-->
+   * [RabbitMQ Cluster Operator Versioning &amp; Release Strategy](#rabbitmq-cluster-operator-versioning--release-strategy)
+      * [Table of Contents](#table-of-contents)
+      * [Glossary](#glossary)
+         * [Preferred / Storage version](#preferred--storage-version)
+      * [Summary](#summary)
+      * [Motivation](#motivation)
+         * [Goals](#goals)
+            * [Discovery](#discovery)
+            * [Comprehension](#comprehension)
+            * [Support](#support)
+            * [Consumption](#consumption)
+         * [Non-Goals/Future Work](#non-goalsfuture-work)
+      * [Versioning of related software &amp; dependencies](#versioning-of-related-software--dependencies)
+         * [RabbitMQ](#rabbitmq)
+         * [Kubernetes &amp; components](#kubernetes--components)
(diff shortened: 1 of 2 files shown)


MirahImage:
Thank you for putting in the work to make such a thorough proposal. 

I'm in agreement that some form of CalVer seems likely to be our best option going forward.

ChunyiLyu:
Note that OLM's `ClusterServiceVersions ` metadata expects operator versions to be semver complaint. If we were to version our operator with `CalVer` and would like to integrate with OLM, then we will have to specify different versions in CSVs of our operator to be OLM complaint.

However, there is currently no plan for the team to adopt OLM as the installation and distribution method.  I am not suggesting that we need to have our operator work with OLM. Just added this as an additional context for our decision.

MirahImage:
> Note that OLM's [`ClusterServiceVersions ` metadata expects operator versions to be semver complaint](https://docs.openshift.com/container-platform/4.2/operators/understanding_olm/olm-understanding-olm.html#olm-csv_olm-understanding-olm).
Interestingly, I think that since calendar years are monotonically increasing (barring anything like calendar resets), technically, the proposal is semver compliant, it just happens to have a very large major version.

ferozjilla:
We discussed the proposal in this PR in a team meeting with the following conclusions: 

1. **Use a non-strict SemVer versioning scheme for the Operator**: We decided this since Tanzu RabbitMQ will be versioned using CalVer and similar CalVer based versioning for the Operator might cause confusion. For example, Tanzu RabbitMQ version 2020-06.11 using the Operator version 2019-10.2 might read a bit strangely and also create more questions around supportability than help us in this case. So a separate versioning scheme - like SemVer, will not cause any confusion of double CalVer versions. We are yet to define "non-strict", and there will be another issue in the backlog to deal with this.

1. **Document our SemVer policy as a REAMDE in the Operator Repository**: This is where we can define the "non-strict" in our versioning policy. Also a suggestion to use the terminology  [RFC 2119 format for this](https://tools.ietf.org/html/rfc2119). This will be continued in a separate story.

1. **Bump major version to 1 when we GA, keep building 0.x.y releases until then.*

PR #243, opened 2020-08-12
Move base operator image to scratch

### LINKED ISSUE #194: Invstigate different Operator base images

coro:
This closes #194 

## Summary Of Changes

Moves the base image from Ubuntu to scratch.

## Additional Context

There's a few additional requirements on the image other than just running the binary:

* Access to tzdata - provided by Golang 1.15 with the `-tags timetzdata` flag
* Access to CA certs for TLS - provided by installing on Alpine, then copying over to the manager image
* Running as a specified User, with a UID / GID of 1000 - provided by creating a custom `/etc/passwd` & `/etc/group` file

## Local Testing

```
```

```
$ kubectx dev-bunny
$ make destroy deploy-dev
$ make system-tests
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM golang:1.13 as builder
+FROM golang:1.15 as builder
 
 WORKDIR /workspace
 # Copy the go source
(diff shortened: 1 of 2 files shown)


MirahImage:
I think the use of alpine, even in a build step, might cause us potential legal issues? I'm not familiar with the problem, but maybe worth using something other than alpine just so we are for sure okay with legal.

PR #246, opened 2020-08-12
Plugins

### LINKED ISSUE #224: Coordinate plugins and configurations update

When list of `enabled_plugins` is updated, our controller enables the plugins through a `podExec`. The `enabled_plugins` value in the server-conf configMap is updated. This means that for an existing cluster, update to any configuration (`rabbitmq.conf` or `advanced.config`) that can only be successfully loaded if an additional corresponding plugin is running, will fail to deploy.

The most straightforward solution I can think of is to separate `enabled_plugins` in a separate configMap, and always keep the configMap up-to-date with the latest enabled plugins.That way, when people are updating `rabbitmq.conf` or `advanced.config` that requires a new plugins to be enabled, statefulSet will be restarted and the `enabled_plugins` configMaps, which is also up-to-date will be reloaded and new plugins will be enabled at node start.

The current implementation on how the controller decides to trigger a statefulSet restart, replies on the fact that the server-conf configMap is only updated with configuration changes. That's the reason I think it's easier to just separate the `enabled_plugins` to a different configMap.

**Additional context**
Here are the steps to reproduce the issue:

1. Have a healthy rabbitmq k8s cluster running
2. Add the following lines to `spec` replacing the username, password and uri
`kubectl edit rabbitmqclusters.rabbitmq.com`

```
rabbitmq:
   additionalPlugins:
   - rabbitmq_shovel
  
   advancedConfig: |
        [{rabbitmq_shovel,
              [ {shovels, [ {my_first_shovel,
                [ {source,
                  [ {protocol, amqp091},
                    {uris, [ "amqp://username:password@uri"]},
                  {declarations, [ {'exchange.declare',
                                     [ {exchange, <<"my_fanout">>},
                                       {type, <<"fanout">>},
                                       durable
                                     ]},
                                   {'queue.declare',
                                     [{arguments,
                                         [{<<"x-message-ttl">>, long, 60000}]}]},
                                   {'queue.bind',
                                     [ {exchange, <<"my_fanout">>},
                                       {queue,    <<>>}
                                     ]}
                                  ]},
                  {queue, <<>>},
                  {prefetch_count, 10} ]},
              {destination,
                [ {protocol, amqp091},
                  {uris, ["amqp://username:password@uri"]},
                  {declarations, [ {'exchange.declare',
                                      [ {exchange, <<"my_direct">>},
                                          {type, <<"direct">>},
                                            durable
                                          ]}
                                  ]},
                  {publish_properties, [ {delivery_mode, 2} ]},
                  {add_forward_headers, true},
                  {publish_fields, [ {exchange, <<"my_direct">>},
                                     {routing_key, <<"from_shovel">>}
                                  ]}
                ]},
              {ack_mode, on_confirm},
              {reconnect_delay, 5}
            ]}
        ]} ]} ].
```

3. The cluster will fail to start with following errors. 
` kubectl logs <pod name>`

```
Error during startup: {error,failed_to_read_advanced_configuration_file}

08:37:24.466 [error] Error during startup: {error,failed_to_read_advanced_configuration_file}
08:37:24.466 [error]
08:37:25.467 [error] Supervisor rabbit_prelaunch_sup had child prelaunch started with rabbit_prelaunch:run_prelaunch_first_phase() at undefined exit with reason failed_to_read_advanced_configuration_file in context start_error
08:37:25.467 [error] CRASH REPORT Process <0.153.0> with 0 neighbours exited with reason: {{shutdown,{failed_to_start_child,prelaunch,failed_to_read_advanced_configuration_file}},{rabbit_prelaunch_app,start,[normal,[]]}} in application_master:init/4 line 138
{"Kernel pid terminated",application_controller,"{application_start_failure,rabbitmq_prelaunch,{{shutdown,{failed_to_start_child,prelaunch,failed_to_read_advanced_configuration_file}},{rabbit_prelaunch_app,start,[normal,[]]}}}"}
Kernel pid terminated (application_controller) ({application_start_failure,rabbitmq_prelaunch,{{shutdown,{failed_to_start_child,prelaunch,failed_to_read_advanced_configuration_file}},{rabbit_prelaunch
```

4. Also, `server-conf` configmap will not contain `rabbitmq_shovel` in the `enabled_plugins` list. 
`kubectl describe configmaps rabbitmq-server-conf`

harshac:
This closes #224, #259

## Summary Of Changes
- Create a new plugins builder
- Add it to the ordered list of builder alongwith the server configmap
- Add new configmap for plugins to sts
- Refactor annotations and add unit tests
- Updates the plugins before updating server conf
- Modifies controller to restart stateful set based on the builder
- Refactors controller tests to remove duplication

## Local Testing

```
```

```
$ kubectx dev-bunny
$ make destroy deploy-dev
$ make system-tests
```

### DIFF
--- a/controllers/rabbitmqcluster_controller.go
+++ b/controllers/rabbitmqcluster_controller.go
@@ -17,7 +17,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
-	"strings"
 	"time"
 
 	"k8s.io/client-go/kubernetes"
@@ -355,10 +354,10 @@ func (r *RabbitmqClusterReconciler) allReplicasReady(ctx context.Context, rmq *r
 // enablePlugins - helper function to set the list of enabled plugins in a given RabbitmqCluster pods
 // `rabbitmq-plugins set` disables plugins that are not in the provided list
 func (r *RabbitmqClusterReconciler) enablePlugins(rmq *rabbitmqv1beta1.RabbitmqCluster) error {
+	plugins := resource.NewRabbitMQPlugins(rmq.Spec.Rabbitmq.AdditionalPlugins)
 	for i := int32(0); i < *rmq.Spec.Replicas; i++ {
 		podName := fmt.Sprintf("%s-%d", rmq.ChildResourceName("server"), i)
-		rabbitCommand := fmt.Sprintf("rabbitmq-plugins set %s",
-			strings.Join(resource.AppendIfUnique(resource.RequiredPlugins, rmq.Spec.Rabbitmq.AdditionalPlugins), " "))
+		rabbitCommand := fmt.Sprintf("rabbitmq-plugins set %s", plugins.AsString(" "))
 
 		stdout, stderr, err := r.exec(rmq.Namespace, podName, "rabbitmq", "sh", "-c", rabbitCommand)
 
--- a/controllers/rabbitmqcluster_controller.go
+++ b/controllers/rabbitmqcluster_controller.go
@@ -17,6 +17,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
+	"strings"
 	"time"
 
 	"k8s.io/client-go/kubernetes"
@@ -422,28 +423,25 @@ func (r *RabbitmqClusterReconciler) exec(namespace, podName, containerName strin
 // logAndRecordOperationResult - helper function to log and record events with message and error
 // it logs and records 'updated' and 'created' OperationResult, and ignores OperationResult 'unchanged'
 func (r *RabbitmqClusterReconciler) logAndRecordOperationResult(rmq runtime.Object, resource runtime.Object, operationResult controllerutil.OperationResult, err error) {
-	if operationResult == controllerutil.OperationResultCreated && err == nil {
-		msg := fmt.Sprintf("created resource %s of Type %T", resource.(metav1.Object).GetName(), resource.(metav1.Object))
+	var operation string
+	if operationResult == controllerutil.OperationResultCreated {
+		operation = "create"
+	}
+
+	if operationResult == controllerutil.OperationResultUpdated {
+		operation = "update"
+	}
+
+	if err == nil {
+		msg := fmt.Sprintf("%sd resource %s of Type %T", operation, resource.(metav1.Object).GetName(), resource.(metav1.Object))
 		r.Log.Info(msg)
-		r.Recorder.Event(rmq, corev1.EventTypeNormal, "SuccessfulCreate", msg)
+		r.Recorder.Event(rmq, corev1.EventTypeNormal, fmt.Sprintf("Successful%s", strings.Title(operation)), msg)
 	}
 
-	if operationResult == controllerutil.OperationResultCreated && err != nil {
-		msg := fmt.Sprintf("failed to create resource %s of Type %T", resource.(metav1.Object).GetName(), resource.(metav1.Object))
+	if err != nil {
+		msg := fmt.Sprintf("failed to %s resource %s of Type %T", operation, resource.(metav1.Object).GetName(), resource.(metav1.Object))
 		r.Log.Error(err, msg)
-		r.Recorder.Event(rmq, corev1.EventTypeWarning, "FailedCreate", msg)
-	}
-
-	if operationResult == controllerutil.OperationResultUpdated && err == nil {
-		msg := fmt.Sprintf("updated resource %s of Type %T", resource.(metav1.Object).GetName(), resource.(metav1.Object))
-		r.Log.Info(msg)
-		r.Recorder.Event(rmq, corev1.EventTypeNormal, "SuccessfulUpdate", msg)
-	}
-
-	if operationResult == controllerutil.OperationResultUpdated && err != nil {
-		msg := fmt.Sprintf("failed to update resource %s of Type %T", resource.(metav1.Object).GetName(), resource.(metav1.Object))
-		r.Log.Error(err, msg)
-		r.Recorder.Event(rmq, corev1.EventTypeWarning, "FailedUpdate", msg)
+		r.Recorder.Event(rmq, corev1.EventTypeWarning, fmt.Sprintf("Failed%s", strings.Title(operation)), msg)
 	}
 }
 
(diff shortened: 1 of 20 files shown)


mkuratczyk:
Unfortunately this PR breaks the feature of enabling plugins without restart. Every time I add or remove a plugin from `additionalPlugins` I see the pod restarted. This should not happen if there are no configuration changes.

Steps:
1. Deploy an instance with no `additionalPlugins`
2. `kubectl edit rabbitmq ...` and add `additionalPlugins: [rabbitmq_shovel]`
3. See the pod being terminated (it shouldn't)
4. Edit again, remove `additionalPlugins`
5. See the pod being terminated (it shouldn't)

controllers/rabbitmqcluster_controller.go
```diff
@@ -312,10 +312,10 @@ func (r *RabbitmqClusterReconciler) setAdminStatus(ctx context.Context, rmq *rab
 	return nil
 }
 
-// restartStatefulSetIfNeeded - helper function that annotates the StatefulSet PodTemplate with current timestamp
-// to trigger a restart of the all pods in the StatefulSet when ConfigMap is updated
-func (r *RabbitmqClusterReconciler) restartStatefulSetIfNeeded(ctx context.Context, resource runtime.Object, operationResult controllerutil.OperationResult, rmq *rabbitmqv1beta1.RabbitmqCluster) {
-	if _, ok := resource.(*corev1.ConfigMap); ok && operationResult == controllerutil.OperationResultUpdated {
+// restartStatefulSetIfNeeded - helper function that annotate the StatefulSet PodTemplate with current timestamp
```
ChunyiLyu, line 315: small: I think it's "annotates"?

harshac: Done

internal/resource/rabbitmq_plugins.go
```diff
@@ -0,0 +1,97 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+var requiredPlugins = []string{
+	"rabbitmq_peer_discovery_k8s", // required for clustering
+	"rabbitmq_prometheus",         // enforce prometheus metrics
+	"rabbitmq_management",
+}
+
+const pluginsConfigMapName = "plugins-conf"
```
ChunyiLyu, line 19: small: may I suggest to call it "plugins" to shorten it?

harshac: Do you mean the name of the constant?

internal/resource/rabbitmq_resource_builder.go
```diff
@@ -22,6 +22,7 @@ type RabbitmqResourceBuilder struct {
 type ResourceBuilder interface {
 	Update(runtime.Object) error
 	Build() (runtime.Object, error)
+	RequiresClusterRestart() bool
```
ChunyiLyu, line 25: Small: I find the name of this method a bit confusing. Without the context, the name suggests that it should be the single method needed to be called to determine whether to restart the cluster or not. However, when using the method, we still need to check the operational result.

Maybe rename it to "UpdateRequiresStsRestart"? so that it's a bit more specific? Your call.

internal/resource/rabbitmq_plugins.go
```diff
@@ -0,0 +1,97 @@
...
+		}
+	}
+	return enabledPlugins
+}
+
+func (r *RabbitMQPlugins) AsString(sep string) string {
+	return strings.Join(r.DesiredPlugins(), sep)
+}
+
+func (builder *RabbitmqResourceBuilder) RabbitmqPluginsConfigMapBuilder() *RabbitmqPluginsConfigMapBuilder {
```
ChunyiLyu, line 60: Small: method name can be renamed to `RabbitmqPluginsConfigMap` so it's less redundant and consistent with other resource builder methods.

internal/resource/rabbitmq_plugins.go
harshac: I will rename it to pluginsConfig to indicate that it's a configuration.

ChunyiLyu:
```
const pluginsConfigMapName = "plugins-conf"
```
to 
```
const pluginsConfigMapName = "plugins"
```

So that the name of the configMap is shorted

internal/resource/rabbitmq_plugins.go
```diff
@@ -0,0 +1,97 @@
...
+	return RabbitMQPlugins{
+		requiredPlugins:   requiredPlugins,
+		additionalPlugins: additionalPlugins,
+	}
+}
+
+func (r *RabbitMQPlugins) DesiredPlugins() []string {
+	allPlugins := append(r.requiredPlugins, r.additionalPlugins...)
+
+	check := make(map[string]bool)
```
ChunyiLyu, line 45: I see that you are deduping the list of plugins for the value for running the command. Why not do the same thing for the values in the configMap? Am I missing something? You can extract the dedup logic and use it in `pluginsValue()`

harshac: The config map also takes in the deduped value. The asString in plugins returns deduped list as string. Maybe should name it better to make it clearer. 

I can't think of a good name for `pluginsValue`. Can you suggest something?

controllers/rabbitmqcluster_controller_test.go
```diff
@@ -918,23 +758,30 @@ var _ = Describe("RabbitmqclusterController", func() {
...
 			Expect(allEventMsgs).To(ContainSubstring(fmt.Sprintf("updated resource %s of Type *v1.ConfigMap", rabbitmqCluster.ChildResourceName("server-conf"))))
+			Expect(allEventMsgs).To(ContainSubstring(fmt.Sprintf("updated resource %s of Type *v1.ConfigMap", rabbitmqCluster.ChildResourceName("plugins-conf"))))
 			Expect(allEventMsgs).To(ContainSubstring(fmt.Sprintf("updated resource %s of Type *v1.Secret", rabbitmqCluster.ChildResourceName("erlang-cookie"))))
 			Expect(allEventMsgs).To(ContainSubstring(fmt.Sprintf("updated resource %s of Type *v1.Secret", rabbitmqCluster.ChildResourceName("admin"))))
 			Expect(allEventMsgs).To(ContainSubstring(fmt.Sprintf("updated resource %s of Type *v1.ServiceAccount", rabbitmqCluster.ChildResourceName("server"))))
 			Expect(allEventMsgs).To(ContainSubstring(fmt.Sprintf("updated resource %s of Type *v1.Role", rabbitmqCluster.ChildResourceName("endpoint-discovery"))))
-			Expect(allEventMsgs).To(ContainSubstring(fmt.Sprintf("updated resource %s of Type *v1.RoleBinding", rabbitmqCluster.ChildResourceName("server"))))
+			Expect(allEventMsgs).To(ContainSubstring("combined from similar events"))
```
ChunyiLyu, line 784: Is this enough to make the test not flaky? I am not sure whether the problem will only happy for `v1.RoleBinding`. If this happens to `v1.Role` or any other types, we will still have flakes. And I would like the test to reliably pass in the future even if we create a new child resource.

The test here is to make sure that child resources annotations has been updated. I would suggest either to get these object to verify that updated annotations are there, or, to assert on the operator logs.

ChunyiLyu (commented):
I am happy with all the functional changes. Please see my comments regarding the flaky tests that we worked on yesterday. I don't think not asserting on the events from `v1.RoleBinding` is really going to make it not flaky anymore, since this is more of a random problem that happens to the last event published. In the future, when we add a new child resource, we will likely to have this issue again. I would either assert on the operator logs, or get all the resources to see that annotations updates are reflected.

The branch now has almost 20 commits. To keep the history clear and concise, I suggest squash the commits before merging. Maybe one commit for all the feature work on plugins, a commit for `internal/metadata` refactor, and an integration tests refactor.

internal/resource/rabbitmq_plugins.go
ChunyiLyu: Sorry I missed it! I think the name is fine. I need to be more careful when reviewing the code.

ChunyiLyu (approved):
Manually triggered charts and system tests in PR pipeline, and all tests passed. 

Lets merge it in 👍

PR #260, opened 2020-08-19
Use ini library to create rabbitmq.conf

### LINKED ISSUE #256: Use an ini library for creating rabbitmq.conf
Currently, for creating the RabbitMQ config file, we use multi line strings and string builders as done [here](https://github.com/rabbitmq/cluster-operator/blob/920f147e6d61f62b120f6e6ff23e7f6b50dd32e3/internal/resource/configmap.go#L25-L100).

Since the new RabbitMQ config uses an ini / TOML like format, we might want to replace this manual string formatting using a well tested and simple to use ini library like https://github.com/go-ini/ini.

ansd:
This closes #256 

**Advantages:**
* no need to manually format rabbitmq.conf (using StringBuilder and taking care of newlines etc.)
* conf gets pretty formatted
* when additional conf is provided in the spec, it not only appends but actually overwrites duplicate keys (which is then less confusing to people looking into the conf file)
* basic validation for the user provided `spec.rabbitmq.additionalConfig` is now happening at an earlier stage during reconciliation.  For example, when the user enters an invalid additional conf like `= invalid`, the pod won't even get created and the rabbitmqcluster status includes a well formatted error message:
```
Status:
  Conditions:
    Message:               failed to append spec.rabbitmq.additionalConfig: error creating new key: empty key name
    Reason:                Error
    Status:                False
    Type:                  ReconcileSuccess
```
Ideally, we'll move that part to a validating webhook in future.

**Disadvantages:**
* Additional dependency to external library. However, https://github.com/go-ini/ini seems to be widely used, well tested and maintained.


internal/resource/configmap.go
```diff
@@ -63,41 +60,47 @@ func (builder *ServerConfigMapBuilder) Update(object runtime.Object) error {
...
+	if _, err := cfg.WriteTo(&buffer); err != nil {
+		return err
+	}
+
+	if configMap.Data == nil {
+		configMap.Data = make(map[string]string)
 	}
 
-	configMap.Data["rabbitmq.conf"] = rmqConfBuilder.String()
+	configMap.Data["rabbitmq.conf"] = buffer.String()
```
harshac, line 103: Small: Rename the `buffer` to maybe `rmqConfBuffer`.

ansd: Done

harshac (approved):

ferozjilla (approved):
LGTM

ferozjilla:
Just a note to squash the commits into a single commit and merge :)

PR #262, opened 2020-08-20
Fix condition AllReplicasReady

### LINKED ISSUE #261: AllReplicasReady outputs True although no pod created
## Describe the bug
RabbitMQ cluster condition `AllReplicasReady` outputs `True` although no pod got created.

## To Reproduce
1. `make deploy-kind`
2. `kubectl apply -f cr-example.yaml` where cf-example.yaml contains
```
kind: RabbitmqCluster
metadata:
  name: example
  namespace: default
spec:
  affinity: {}
  image: rabbitmq
  imagePullSecret: my-secret
  persistence:
    storage: 10Gi
  replicas: 1
  resources:
    limits:
      cpu: "2"
      memory: 2Gi
    requests:
      cpu: "1"
      memory: 2Gi
  service:
    type: ClusterIP
  tolerations:
  - {}
``` 
3. `kubectl describe statefulsets.apps example-rabbitmq-server`
will output the following event since no pod could be created:
```
Warning  FailedCreate      31s (x15 over 113s)  statefulset-controller  create Pod example-rabbitmq-server-0 in StatefulSet example-rabbitmq-server failed error: Pod "example-rabbitmq-server-0" is invalid: spec.tolerations[0].operator: Invalid value: "": operator must be Exists when `key` is empty, which means "match all values and all keys"
```

```
kubectl get pods
No resources found in default namespace.
```

However, `kubectl describe rabbitmqcluster example`
outputs
```
  Conditions:
    Last Transition Time:  2020-08-20T13:39:55Z
    Reason:                AllPodsAreReady
    Status:                 True
    Type:                  AllReplicasReady
```

**Expected behavior**
`kubectl describe rabbitmqcluster example`
should output
```
  Conditions:
    Last Transition Time:  2020-08-20T13:39:55Z
    Message:             0/1 Pods ready
    Reason:                NotAllPodsReady
    Status:                False
    Type:                  AllReplicasReady
```

## Version information
- RabbitMQ: 3.8.5
- RabbitMQ Cluster Operator: https://github.com/rabbitmq/cluster-operator/commit/cccfcb4f724a08472a216f51a948923f524b16c0
- Kubernetes: 1.17.5
 
## Additional context
This bug occurs because https://github.com/rabbitmq/cluster-operator/blob/fba823feaf43b4419511c6f8872dcfeb4d0f3da8/internal/status/all_replicas_ready.go#L40 compares `Status.Replicas` with  `Status.ReadyReplicas`. However, the former "is the number of Pods created by the StatefulSet controller". What we want instead is `StatefulSetSpec.Replicas` which is the desired number of replicas.

ansd:
This closes #261

### DIFF
--- a/internal/status/all_replicas_ready_test.go
+++ b/internal/status/all_replicas_ready_test.go
@@ -23,16 +23,16 @@ import (
 
 var _ = Describe("AllReplicasReady", func() {
 	var (
-		sts               *appsv1.StatefulSet
-		existingCondition *rabbitmqstatus.RabbitmqClusterCondition
-		desiredReplicas   int32 = 5
+		sts             *appsv1.StatefulSet
+		oldCondition    *rabbitmqstatus.RabbitmqClusterCondition
+		desiredReplicas int32 = 5
 	)
 
 	BeforeEach(func() {
 		sts = &appsv1.StatefulSet{
 			Status: appsv1.StatefulSetStatus{},
 		}
-		existingCondition = nil
+		oldCondition = nil
 
 	})
 
@@ -44,7 +44,7 @@ var _ = Describe("AllReplicasReady", func() {
 			})
 
 			It("returns the expected condition", func() {
-				condition := rabbitmqstatus.AllReplicasReadyCondition([]runtime.Object{&corev1.Endpoints{}, sts}, existingCondition)
+				condition := rabbitmqstatus.AllReplicasReadyCondition([]runtime.Object{&corev1.Endpoints{}, sts}, oldCondition)
 
 				By("having status true and reason message", func() {
 					Expect(condition.Status).To(Equal(corev1.ConditionTrue))
@@ -60,7 +60,7 @@ var _ = Describe("AllReplicasReady", func() {
 			})
 
 			It("returns a condition with state false", func() {
-				condition := rabbitmqstatus.AllReplicasReadyCondition([]runtime.Object{sts}, existingCondition)
+				condition := rabbitmqstatus.AllReplicasReadyCondition([]runtime.Object{sts}, oldCondition)
 
 				By("having status false and reason", func() {
 					Expect(condition.Status).To(Equal(corev1.ConditionFalse))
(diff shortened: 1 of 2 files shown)


internal/status/all_replicas_ready_test.go
```diff
@@ -107,23 +108,23 @@ var _ = Describe("AllReplicasReady", func() {
...
 				})
 
 				It("updates the transition time", func() {
-					condition := rabbitmqstatus.AllReplicasReadyCondition([]runtime.Object{&corev1.Endpoints{}, sts}, existingCondition)
+					condition := rabbitmqstatus.AllReplicasReadyCondition([]runtime.Object{&corev1.Endpoints{}, sts}, oldCondition)
 					Expect(condition.LastTransitionTime).ToNot(Equal(emptyTime))
 				})
 			})
 
 			When("transitions to false", func() {
```
harshac, line 120: I know it's not a change made by this PR but a suggestion. -
The transition to false or true does not really seem to have a previous state. The test blocks are a bit confusing. We can merge these with the `When` blocks above where another `It` could be verifying that the transition time is set. It's upto you.

internal/status/all_replicas_ready_test.go
```diff
@@ -153,32 +154,32 @@ var _ = Describe("AllReplicasReady", func() {
...
+					condition := rabbitmqstatus.AllReplicasReadyCondition([]runtime.Object{sts}, oldCondition)
 
-					Expect(existingCondition).NotTo(BeNil())
-					existingConditionTime := existingCondition.LastTransitionTime.DeepCopy()
+					Expect(oldCondition).NotTo(BeNil())
+					oldConditionTime := oldCondition.LastTransitionTime.DeepCopy()
 
-					Expect(condition.LastTransitionTime.Equal(existingConditionTime)).To(BeFalse())
-					Expect(condition.LastTransitionTime.Before(existingConditionTime)).To(BeFalse())
+					Expect(condition.LastTransitionTime.Equal(oldConditionTime)).To(BeFalse())
```
harshac, line 166: Can we replace the matcher to use Equal instead of BeFalse and BeTrue? Is there a reason for not doing that?

ChunyiLyu (approved):
LGTM

PR #281, opened 2020-08-26
Guidelines for non-strict semver (#265)

### LINKED ISSUE #265: Define guidelines for non-strict SemVer
In https://github.com/rabbitmq/cluster-operator/pull/236, we decided to use non-strict SemVer as our versioning strategy for the Operator. 

For our own reference, and for the reference of our users, let us define and document the guidelines for our non-strict SemVer versioning scheme.

A README in this repository that defines the different types of changes where we would bump major, minor or patch. The README may follow the language from the [RFC 2119 format](https://tools.ietf.org/html/rfc2119). There is a discussion on the different types of changes in https://github.com/rabbitmq/cluster-operator/pull/236 already, and this will be a good place to start.

**Additional context**
The reasoning behind the decision to go with non-strict SemVer is covered in https://github.com/rabbitmq/cluster-operator/pull/236.

harshac:
This closes #265

## Local Testing

```
```

```
$ kubectx dev-bunny
$ make destroy deploy-dev
$ make system-tests
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ The operator deploys RabbitMQ `3.8.5`, and requires a Kubernetes cluster of `1.1
 
 ## Versioning
 
-RabbitMQ Cluster Kubernetes Operator follows non-strict [semver. More details about other versioning strategies that were considered can be found
+RabbitMQ Cluster Kubernetes Operator follows non-strict [semver](https://semver.org/). More details about other versioning strategies that were considered can be found
 [here](https://github.com/rabbitmq/cluster-operator/blob/main/docs/proposals/20200806-versioning-release-strategy.md). 
 
 [Here](version_guidelines.md) are some guidelines on how we implement non-strict semver. The version number MAY or MAY NOT follow the semver rules. Hence, we highly recommend to read
--- a/README.md
+++ b/README.md
@@ -8,6 +8,14 @@ Manage [RabbitMQ](https://www.rabbitmq.com/) clusters deployed to [Kubernetes](h
 
 The operator deploys RabbitMQ `3.8.5`, and requires a Kubernetes cluster of `1.16` or above.
 
+## Versioning
+
+RabbitMQ Cluster Kubernetes Operator follows non-strict [semver. More details about other versioning strategies that were considered can be found
+[here](https://github.com/rabbitmq/cluster-operator/blob/main/docs/proposals/20200806-versioning-release-strategy.md). 
+
+[Here](version_guidelines.md) are some guidelines on how we implement non-strict semver. The version number MAY or MAY NOT follow the semver rules. Hence, we highly recommend to read
+the release notes to understand the changes and their potential impact for any release.
+
 ## Quickstart
 
 ### Deploying on KinD
--- a/version_guidelines.md
+++ b/version_guidelines.md
@@ -1,0 +1,33 @@
+### RabbitMQ Cluster Kubernetes Operator Versioning Scheme
+---
+RabbitMQ Cluster Kubernetes Operator follows non-strict semver. More details about other versioning strategies that were considered can be found
+[here](https://github.com/rabbitmq/cluster-operator/blob/main/docs/proposals/20200806-versioning-release-strategy.md). This document explains the non-strict semver versioning scheme used by RabbitMQ Cluster Kubernetes Operator.
+
+_**Operator** refers to *RabbitMQ Cluster Kubernetes Operator* in the document henceforth._
+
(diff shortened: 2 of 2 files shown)


MirahImage (commented):
The use of RFC 2119 feels a bit like overkill, given that we're not following strict semver, so the version is more guidance rather than a strict statement. While we will attempt to generally follow semver, we cannot guarantee that we will not ship what ends up being a breaking change in a minor or patch release.

ChunyiLyu (approved):
LGTM

harshac:
> The use of RFC 2119 feels a bit like overkill, given that we're not following strict semver, so the version is more guidance rather than a strict statement. While we will attempt to generally follow semver, we cannot guarantee that we will not ship what ends up being a breaking change in a minor or patch release.

I agree that we are not following strict semver. However, the guidelines here can be more or less followed. I think use of RFC 2119 is ok unless it's otherwise not conveying the intent.

MirahImage (approved):

README.md
```diff
@@ -8,6 +8,14 @@ Manage [RabbitMQ](https://www.rabbitmq.com/) clusters deployed to [Kubernetes](h
 
 The operator deploys RabbitMQ `3.8.5`, and requires a Kubernetes cluster of `1.16` or above.
 
+## Versioning
+
+RabbitMQ Cluster Kubernetes Operator follows non-strict [semver](https://semver.org/). More details about other versioning strategies that were considered can be found
+[here](https://github.com/rabbitmq/cluster-operator/blob/main/docs/proposals/20200806-versioning-release-strategy.md).
```
ansd, line 14: I would delete the second sentence since it's already mentioned in version_guidelines.md and is not that relevant for the main README.

README.md
```diff
@@ -8,6 +8,14 @@ Manage [RabbitMQ](https://www.rabbitmq.com/) clusters deployed to [Kubernetes](h
 
 The operator deploys RabbitMQ `3.8.5`, and requires a Kubernetes cluster of `1.16` or above.
 
+## Versioning
+
+RabbitMQ Cluster Kubernetes Operator follows non-strict [semver. More details about other versioning strategies that were considered can be found
+[here](https://github.com/rabbitmq/cluster-operator/blob/main/docs/proposals/20200806-versioning-release-strategy.md). 
+
+[Here](version_guidelines.md) are some guidelines on how we implement non-strict semver. The version number MAY or MAY NOT follow the semver rules. Hence, we highly recommend to read
```
ferozjilla, line 16:
```suggestion
[The versioning guidelines document](version_guidelines.md) contains some guidelines on how we implement non-strict semver. The version number MAY or MAY NOT follow the semver rules. Hence, we highly recommend to read
```

README.md
```diff
@@ -8,6 +8,14 @@ Manage [RabbitMQ](https://www.rabbitmq.com/) clusters deployed to [Kubernetes](h
 
 The operator deploys RabbitMQ `3.8.5`, and requires a Kubernetes cluster of `1.16` or above.
 
+## Versioning
+
+RabbitMQ Cluster Kubernetes Operator follows non-strict [semver. More details about other versioning strategies that were considered can be found
```
ferozjilla:
```suggestion
RabbitMQ Cluster Kubernetes Operator follows non-strict [semver](https://semver.org/). More details about other versioning strategies that were considered can be found
```

ferozjilla (changes requested):

ansd (changes requested):

version_guidelines.md
```diff
@@ -0,0 +1,33 @@
...
+
+1. A normal version number MUST take the form X.Y.Z where X, Y, and Z are non-negative integers, and MUST NOT contain leading zeroes. 
+X is the major version, Y is the minor version, and Z is the patch version. Each element MUST increase numerically. For instance: 1.9.0 -> 1.10.0 -> 1.11.0.
+
+2. Major version zero (0.y.z) is for initial development. This is a Beta version and anything MAY change at anytime. The Operator SHOULD NOT be considered stable.
+
+3. Version 1.0.0 defines the GA/stable version of Operator. The way in which the version number is incremented after this release is dependent 
+on how the Operator changes.
+
+4. Patch version Z (x.y.Z | x > 0) MUST be incremented if only backwards compatible bug fixes and/or CVE fixes are introduced.
```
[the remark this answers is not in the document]
harshac: I think the MUST here is correct. I do not see a scenario where we will have a bug fix but not increment the patch version, unless it's being released in a minor or a major which is covered in those sections. Am I missing something?

Issue #228, opened 2020-07-31
1-step installation


Zerpet:
**Is your feature request related to a problem? Please describe.**
Our current [installation guide](https://www.rabbitmq.com/kubernetes/operator/install-operator.html#using-kubectl-yaml) has a number of steps that are not "beginner friendly", for example building your own Docker image or clonning the repo. These steps assume developer experience with Git and with Docker, which might not be true for some users of this project, who might have a more "IT-ops" kind of background. We should make the installation experience as simple as possible.

**Describe the solution you'd like**
Inspired by project [Antrea](https://github.com/vmware-tanzu/antrea/blob/master/docs/getting-started.md#installation), we should aim to have an installation method as simple as `kubectl apply -f <github-url/to/a/file.yaml>`. Kubectl already supports pointing to URLs and it has a built-in module for `kustomize` files, which can be invoked using `kubectl apply -k <github-url>//path/to/some/folder/`. We might need to create a new kustomization folder, or change the current layout (dangerous) to achieve this.

Examples of the above working for this repo. Please note that the double forward slash is intended and required:

```bash
kubectl create -k github.com/rabbitmq/cluster-operator//config/namespace/
kubectl create -k github.com/rabbitmq/cluster-operator//config/crd/
```

**Describe alternatives you've considered**
Using pure `kustomize` is also posible e.g. `kustomize build github.com/rabbitmq/cluster-operator//config/rbac/`, however the adoption and use of `kustomize` is not guaranteed and some users might not have this tool installed. It is almost guaranteed that a Kubernetes user will have `kubectl` installed, therefore we should favour the use `kubectl` when possible.

**Additional context**
We can't invoke `kubectl create -k github.com/rabbitmq/cluster-operator//config/default/base/` because it will try to load files from `../some/other/location`, which won't exist if we use this method..

Git should be aware of symlinks, it's worth trying out and exploring if a "central" folder with symlinks to other resources can be a solution.

coro:
Seems like there's a tradeoff here - since we're using kustomize either way (either packaged in kubectl or directly), we should consider whether we're happy having kustomize integration pinned to v2.0.1 (packaged in kubectl) and potentially not function with the latest kustomize (we've been using v3.7.0). Seems like if we go down the `kubectl create -k` route then we're adopting the former.

coro:
[...]

PR #296, opened 2020-08-28
Remove operator Helm chart

ansd:
We implement one step installation of the operator in #228.

In addition, `kubectl rabbitmq install-cluster-operator` can be used to install the operator.

We don't want to support another installation method via helm because it requires maintenance and suffers from CRD upgrade issues.

The rabbitmq chart is still kept, in particular for KSM integration.


ChunyiLyu:
what about this: https://github.com/rabbitmq/cluster-operator/tree/1160858f895e467c5f267833861d369f14924d4b/config/default/overlays/helm

ChunyiLyu (changes requested):
What about this kustomize folder: https://github.com/rabbitmq/cluster-operator/tree/1160858f895e467c5f267833861d369f14924d4b/config/default/overlays/helm

I think the charts/Makefile can use come clean up as well: https://github.com/rabbitmq/cluster-operator/blob/main/charts/Makefile

ChunyiLyu (approved):
LGTM

PR #297, opened 2020-08-28
Generate single manifest for installation #(228)

### LINKED ISSUE #228: 1-step installation

harshac:
This relates #228 

## Summary Of Changes

Adapted `generate-installation-manifest` make target to generate a single-file manifest with all the resources required to deploy Cluster Operator. The image name and tag is hard-coded, however it can be (and it's expected to) changed using `kustomize edit set image <..>` command. The pipeline will run a similar `kustomize` command to generate release specific, versioned, manifest files.

## Additional Context

This change is the first step to accomplish 1-step install. Second step is to adapt the pipeline to leverage this change and produce versioned manifest files.

## Local Testing

N/A no changes to code or to how we deploy.

Edit: @Zerpet picked this PR and added some changes.


config/installation/kustomization.yaml
```diff
@@ -7,10 +7,16 @@
...
 namespace: rabbitmq-system
+images:
+- name: controller
+  newName: rabbitmqoperator/rabbitmq-cluster-kubernetes-operator-dev
+  newTag: latest
 
-patches:
-- manager_deployment_patch.yaml
+resources:
+- generated/namespace.yaml
```
harshac, line 19: Check if we can directly use the sibling directory resource here for namespace and crd

Makefile
```diff
@@ -90,19 +90,22 @@ deploy-kind: check-env-docker-repo git-commit-sha patch-kind manifests deploy-na
...
 	kustomize build config/namespace/base/ > charts/operator/templates/namespace.yaml
 	kustomize build config/crd/ > charts/operator/templates/crd.yaml
 	kustomize build config/rbac/ > charts/operator/templates/rbac.yaml
 	kustomize build config/default/overlays/helm/ > charts/operator/templates/deployment.yaml
 
+install_dir = config/installation/generated
+generate-installation-manifest:
+	mkdir -p $(install_dir)
+	mkdir -p releases
+	cp config/namespace/base/namespace.yaml $(install_dir)/namespace.yaml
```
harshac, line 103: Remove the cp if we can use sibling resources in kustomization.yml under config/installation

Makefile
```diff
@@ -90,19 +90,22 @@ deploy-kind: check-env-docker-repo git-commit-sha patch-kind manifests deploy-na
...
-	kustomize build config/installation > installation/operator.yaml
-
 generate-helm-manifests:
 	kustomize build config/namespace/base/ > charts/operator/templates/namespace.yaml
 	kustomize build config/crd/ > charts/operator/templates/crd.yaml
 	kustomize build config/rbac/ > charts/operator/templates/rbac.yaml
 	kustomize build config/default/overlays/helm/ > charts/operator/templates/deployment.yaml
 
+install_dir = config/installation/generated
+generate-installation-manifest:
```
harshac, line 100: Should accept a docker-image-url, if missing points to latest.

Makefile
```diff
@@ -90,19 +90,22 @@ deploy-kind: check-env-docker-repo git-commit-sha patch-kind manifests deploy-na
...
+	mkdir -p $(install_dir)
+	mkdir -p releases
+	cp config/namespace/base/namespace.yaml $(install_dir)/namespace.yaml
+	cp config/crd/bases/rabbitmq.com_rabbitmqclusters.yaml $(install_dir)/rabbitmq.com_rabbitmqclusters.yaml
+	kustomize build config/rbac/ > $(install_dir)/rbac.yaml
+	kustomize build config/manager/ > $(install_dir)/manager.yaml
+	kustomize build config/installation/ > releases/rabbitmq-cluster-operator.yaml
```
harshac, line 107: Replace the docker image url before building the rabbitmq-cluser-operator.yaml if provided.

config/installation/kustomization.yaml
```diff
@@ -7,10 +7,16 @@
...
 
-resources:
-- ../default/base
-
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
 namespace: rabbitmq-system
+images:
+- name: controller
+  newName: rabbitmqoperator/rabbitmq-cluster-kubernetes-operator-dev
```
harshac, line 15: Change the image name to match the dockerhub image

harshac (commented):
* Also figure out how to override the docker image url without cloning the repo in case of private registry
* Modify the docs

coro (approved):

mkuratczyk (approved):
Looks good!

Zerpet:
Checks are green 🟢 and with two approvals ✅ I'm merging this PR 🚀

PR #300, opened 2020-08-31
Avoid dirty git state

ansd:
Before this commit, when running `make deploy-dev` or `make deploy-kind` the files `config/default/overlays/dev/manager_image_patch.yaml` and `config/default/overlays/kind/manager_image_patch.yaml` showed local changes.

After this commit the git state won't be dirty anymore.

### DIFF
--- a/config/crd/bases/rabbitmq.com_rabbitmqclusters.yaml
+++ b/config/crd/bases/rabbitmq.com_rabbitmqclusters.yaml
@@ -1,3 +1,13 @@
+# RabbitMQ Cluster Operator
+#
+# Copyright 2020 VMware, Inc. All Rights Reserved.
+#
+# This product is licensed to you under the Mozilla Public license, Version 2.0 (the "License").  You may not use this product except in compliance with the Mozilla Public License.
+#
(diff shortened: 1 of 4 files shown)


config/crd/bases/rabbitmq.com_rabbitmqclusters.yaml
```diff
@@ -1,13 +1,3 @@
-# RabbitMQ Cluster Operator
-#
-# Copyright 2020 VMware, Inc. All Rights Reserved.
-#
-# This product is licensed to you under the Mozilla Public license, Version 2.0 (the "License").  You may not use this product except in compliance with the Mozilla Public License.
-#
-# This product may include a number of subcomponents with separate copyright notices and license terms. Your use of these subcomponents is subject to the terms and conditions of the subcomponent's license, as noted in the LICENSE file.
-
-
----
```
ferozjilla, line 10: Looks like an inadvertent change. Let's roll it back 🗞️

ferozjilla (approved):
Looks good! Please just see the comment on the inadvertent change.

config/crd/bases/rabbitmq.com_rabbitmqclusters.yaml
ansd: Added the notice.
It got removed because `make manifests` removed it. Fixed in #305.

PR #305, opened 2020-09-02
Preserve copyright notice

ansd:
Before this commit `make manifests` removed the copyright notice from `config/crd/bases/rabbitmq.com_rabbitmqclusters.yaml`.

After this commit, the copyright notice is preserved.


ferozjilla (approved):
Looks good! Tried it out on the branch and on `main` too :)

PR #308, opened 2020-09-02
Do not exec into pods during rolling update

### LINKED ISSUE #304: Reconcile Error after change the Config Map
## Describe the bug
When the ` definition-rabbitmq-server-conf ` config map is changed the Operator raises an exception:
```golang
2020-09-01T21:32:07.332+0200    ERROR   rabbitmqcluster-controller      Failed to enable plugins on pod definition-rabbitmq-server-0 in namespace default, running command rabbitmq-plugins set rabbitmq_peer_discovery_k8s rabbitmq_prometheus rabbitmq_management with output:   {"error": "unable to upgrade connection: container not found (\"rabbitmq\")"}
github.com/go-logr/zapr.(*zapLogger).Error
        /home/gas/go/pkg/mod/github.com/go-logr/zapr@v0.1.1/zapr.go:128
github.com/rabbitmq/cluster-operator/controllers.(*RabbitmqClusterReconciler).enablePlugins
        /home/gas/go/src/github.com/rabbitmq/cluster-operator/controllers/rabbitmqcluster_controller.go:367
github.com/rabbitmq/cluster-operator/controllers.(*RabbitmqClusterReconciler).Reconcile
```
And the config map won't be changed. 
 
## To Reproduce
1. Install the operator
2. Deploy the definition:
```yaml
kind: RabbitmqCluster
metadata:
  name: definition
spec:
  replicas: 3
```
 
3. Try to change the config map:

Screenshot from 2020-09-01 21-22-58

I was trying to change the `log.file.level = debug`
4. See error:
```
2020-09-01T21:32:07.332+0200    ERROR   rabbitmqcluster-controller      Failed to enable plugins on pod definition-rabbitmq-server-0 in namespace default, running command rabbitmq-plugins set rabbitmq_peer_discovery_k8s rabbitmq_prometheus 
```

**Expected behavior**
Change the config map without errors.

## Version information
- RabbitMQ:  3.8.5
- RabbitMQ Cluster Operator: building from source 3e52b749330693483a9205d2c73a0bee7cd36f9c
- Kubernetes: Kind 1.17.5 
 
## Additional context
When the config map is changed, the reconcile restarts the pod, and when it tries to [enable the plugins](https://github.com/rabbitmq/cluster-operator/blob/main/controllers/rabbitmqcluster_controller.go#L363)  gets the error since the POD isn't ready yet.

ansd:
This closes #304

Exec into pods only if StatefulSet is ready and up to date.

Before this commit, we observed in #304 that the controller tried to exec into pods at the same time as the pods got updated due to a StatefulSet restart resulting in connection errors.

The main change is to not set plugins if
```
sts.Status.ReadyReplicas < desiredReplicas || sts.Status.UpdatedReplicas < desiredReplicas
```
because during a StatefulSet restart, it happened multiple times in #304 that since `sts.Status.ReadyReplicas == desiredReplicas`, the `exec` commands got executed and the connection got interrupted because the pod got updated.

The main finding is that `sts.Status.ReadyReplicas == desiredReplicas` can be `true` although the StatefulSet rolling update is still ongoing:
```
$ while true; do  kubectl get statefulsets.apps definition-rabbitmq-server --template={{.status}}; echo "" ; sleep 1; done

map[collisionCount:0 currentReplicas:2 currentRevision:definition-rabbitmq-server-56b65fdc4f observedGeneration:5 readyReplicas:3 replicas:3 updateRevision:definition-rabbitmq-server-6f667f5f7d]
…
map[collisionCount:0 currentReplicas:2 currentRevision:definition-rabbitmq-server-56b65fdc4f observedGeneration:5 readyReplicas:2 replicas:3 updateRevision:definition-rabbitmq-server-6f667f5f7d]
…
map[collisionCount:0 currentReplicas:2 currentRevision:definition-rabbitmq-server-56b65fdc4f observedGeneration:5 readyReplicas:2 replicas:3 updateRevision:definition-rabbitmq-server-6f667f5f7d updatedReplicas:1]
…
map[collisionCount:0 currentReplicas:1 currentRevision:definition-rabbitmq-server-56b65fdc4f observedGeneration:5 readyReplicas:3 replicas:3 updateRevision:definition-rabbitmq-server-6f667f5f7d updatedReplicas:1]
…
map[collisionCount:0 currentReplicas:1 currentRevision:definition-rabbitmq-server-56b65fdc4f observedGeneration:5 readyReplicas:2 replicas:3 updateRevision:definition-rabbitmq-server-6f667f5f7d updatedReplicas:1]
…
map[collisionCount:0 currentReplicas:1 currentRevision:definition-rabbitmq-server-56b65fdc4f observedGeneration:5 readyReplicas:2 replicas:3 updateRevision:definition-rabbitmq-server-6f667f5f7d updatedReplicas:2]
…
```

Thank you @Gsantomaggio for helping troubleshooting 🙂

### DIFF
--- a/controllers/rabbitmqcluster_controller.go
+++ b/controllers/rabbitmqcluster_controller.go
@@ -198,7 +198,9 @@ func (r *RabbitmqClusterReconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
 			return ctrl.Result{}, err
 		}
 
-		r.restartStatefulSetIfNeeded(ctx, builder, operationResult, rabbitmqCluster)
+		if restarted := r.restartStatefulSetIfNeeded(ctx, builder, operationResult, rabbitmqCluster); restarted {
+			return ctrl.Result{RequeueAfter: time.Second * 10}, nil
+		}
 	}
 
 	// Set ReconcileSuccess to true here because all CRUD operations to Kube API related
@@ -214,7 +216,7 @@ func (r *RabbitmqClusterReconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
 		return ctrl.Result{}, err
 	}
 
-	if ok, err := r.allReplicasReady(ctx, rabbitmqCluster); !ok {
+	if ok, err := r.allReplicasReadyAndUpdated(ctx, rabbitmqCluster); !ok {
 		// only enable plugins when all pods of the StatefulSet become ready
 		// requeue request after 10 seconds without error
 		logger.Info("Not all replicas ready yet; requeuing request to enable plugins on RabbitmqCluster",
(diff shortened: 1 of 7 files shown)


mkuratczyk:
Why do we even try to run `rabbitmq-plugins set` if the cluster is restarting?  Since https://github.com/rabbitmq/cluster-operator/issues/224 is done, we should only exec into the container to enable/disable plugins if the plugins ConfigMap is changed while the cluster is running. If the cluster configuration is changed (a different ConfigMap), we should restart the cluster and it should load plugins on startup based on the plugins ConfigMap - no need to run `rabbitmq-plugins set` at all. What am I missing? :)

Gsantomaggio:
It doesn't crash anymore, but I think there is still some problem.

I tried to change the config map, by changing some value:
Screenshot from 2020-09-03 18-40-27

Changed:
```
queue_master_locator                                     = random
cluster_formation.node_cleanup.interval         = 50
```

but the config map does not change
```bash
 ~ kubectl get configmaps definition-rabbitmq-server-conf -o yaml | grep queue_master_locator
queue_master_locator                            = min-masters
```

ansd:
@mkuratczyk as we discussed this morning, you're right, we should only exec into the container to enable/disable plugins if the plugins ConfigMap is changed. We can and should still use the same guards as in this PR to only exec if all replicas are ready and not in the middle of a rolling update.

ansd:
@Gsantomaggio that's desired behaviour.
You changed the configmap directly (e.g. via `kubectl edit configmap definition-rabbitmq-server-conf`).
You should instead apply the RabbitMQ Custom Resource setting the `additionalConfig` like:
```yaml
kind: RabbitmqCluster
...
spec:
  rabbitmq:
    additionalConfig: |
      queue_master_locator = random
      cluster_formation.node_cleanup.interval = 50
```

Gsantomaggio:
Ok, thank you @ansd 

So why does the operator reboot the pod when the config map is changed?
I mean, if the original configuration is restored, the operator should not reboot the pods, right? Or I am missing something :)!

ansd:
Hey @Gsantomaggio 🙂

Whenever the configuration is updated, the operator restarts the StatefulSet.

That's desired behaviour because https://www.rabbitmq.com/configure.html states that
> rabbitmq.conf and advanced.config changes take effect after a node restart

This logic happens in https://github.com/rabbitmq/cluster-operator/blob/585c609edc9c74b5d2adbc4e2c9c4e5863acb5ad/controllers/rabbitmqcluster_controller.go#L318 and https://github.com/rabbitmq/cluster-operator/blob/585c609edc9c74b5d2adbc4e2c9c4e5863acb5ad/internal/resource/configmap.go#L53

In your specific case, you're right that the operator unnecessarily restarts the StatefulSet. So that's a side effect of changing the configuration in the wrong way.

coro (approved):

Zerpet:
I made the change in the system tests to not import the internal library and use a test constant. Given that we are 3 people at the moment and 2/3 have reviewed this, I will merge the PR when the checks pass.

PR #317, opened 2020-09-07
Use new rabbitmq-upgrade commands in preStop hook

### LINKED ISSUE #255: Use new rabbitmq-upgrade commands
[RabbitMQ 3.8.4](https://github.com/rabbitmq/rabbitmq-server/releases/tag/v3.8.4) introduces the two new commands
`rabbitmq-upgrade await_online_quorum_plus_one` and
`rabbitmq-upgrade await_online_synchronized_mirror`.
Both commands support a timeout flag.

Let's simplify our code by using these two new commands instead of using our own waiting logic in [here](https://github.com/rabbitmq/cluster-operator/blob/7b7b09e33d119ed68d50108ee1d9c50e16f2abe1/internal/resource/statefulset.go#L558-L562).

ChunyiLyu:
This closes #255

## Summary Of Changes

Use `rabbitmq-upgrade` commands to check synchronization and quorum status, instead of `rabbitmq-queues` commands.
Timeout is set to the same value as the default for grace period timeout.

## Local Testing

### DIFF
--- a/internal/resource/statefulset.go
+++ b/internal/resource/statefulset.go
@@ -594,13 +594,10 @@ func (builder *StatefulSetBuilder) podTemplateSpec(annotations, labels map[strin
 					Lifecycle: &corev1.Lifecycle{
 						PreStop: &corev1.Handler{
 							Exec: &corev1.ExecAction{
-								Command: []string{
-									"/bin/bash", "-c", fmt.Sprintf("if [ ! -z \"$(cat /etc/pod-info/%s)\" ]; then exit 0; fi;", DeletionMarker) +
-										" while true; do rabbitmq-queues check_if_node_is_quorum_critical" +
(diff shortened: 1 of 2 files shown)


man
```diff
@@ -0,0 +1,7 @@
+apiVersion: rabbitmq.com/v1beta1
+kind: RabbitmqCluster
+metadata:
+  name: sample0
+  namespace: rabbitmq-system
+spec:
+  replicas: 5
```
ansd, line 7: Is that file accidentally committed?

ansd (changes requested):

man
ChunyiLyu: Yes, I will remove it. Thanks for catching it.

ansd (approved):

MirahImage (approved):

PR #325, opened 2020-09-09
Controller gen update

### LINKED ISSUE #306: controller-gen: Bump and use go tools
Currently, the controller-gen version is hard-coded in https://github.com/rabbitmq/cluster-operator/blob/458514b5f8063b2586efc6f87ab2881d691c576f/Makefile#L145

This issue is about:
1. Bump controller-gen to latest
2. Use the [go tools approach](https://marcofranssen.nl/manage-go-tools-via-go-modules/).
This has two advantages. The first advantage is that in future, we won't need to manually bump the controller-gen version anymore. Depandabot will do it. The second advantage is that we will all use the same `controller-gen` version when developing on our workstations.
3. Include kustomize in the `tools.go` file.

IlyaAlexin:
This closes #306 

## Summary Of Changes

- Removed hardcoded controller-gen installation from Makefile
- Implemented controller-gen installation via tools.go
- As a result changes **config/crd/bases/rabbitmq.com_rabbitmqclusters.yaml**

## Additional Context
The implementation is similar to the solution in the article
[Manage go tools via go modules article](https://marcofranssen.nl/manage-go-tools-via-go-modules/)


ansd:
Hey @IlyaAlexin I squashed your commits and pushed one more commit with some more go tools that we're using. I also added a new make target `install-tools`.

ansd (approved):

ChunyiLyu (approved):
LGTM

PR #345, opened 2020-09-16
Short names for RabbitmqCluster resource

Zerpet:
This PR has no related issue.

## Summary Of Changes
Provides short names `rmq` and `rmqc` as short version of `rabbitmqcluster`. 
The short hand `rc` is already taken by native object `ReplicationController`.

## Additional Context

Ever typo'ed `rabbitmqclsuter` or `rabbitqmclsuter`? Me too.
This PR allows to do `kubectl get rmq` or `kubectl get rmqc`. Never typo again!

## Local Testing

N/A no behavioural change, no change to how we deploy.

### DIFF
--- a/api/v1beta1/rabbitmqcluster_types.go
+++ b/api/v1beta1/rabbitmqcluster_types.go
@@ -37,7 +37,7 @@ const (
 // +kubebuilder:subresource:status
 // +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
 // +kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.clusterStatus"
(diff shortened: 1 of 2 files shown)


ansd (approved):

MirahImage:
Does it make sense to have both `rmq` and `rmqc` as short names? Perhaps just a single short name would be sufficient. If so, I would vote for `rmq`

Zerpet:
I opted to have both since it's "free" and I expect very few to none collisions with other CRDs. Other RabbitMQ CRDs might use the short name `rmq`. I don't see others using `rmqc`.

Zerpet:
Changed to just use `rmq` becase we are *the* RabbitMQ company 🐰

MirahImage (approved):

coro (approved):

Issue #340, opened 2020-09-15
Investigate our coupling to RabbitMQ Community docker entry point script


Zerpet:
## Problem statement

We rely on some behaviour implemented in [RabbitMQ Community](https://hub.docker.com/_/rabbitmq) Docker [entry point script](https://github.com/docker-library/rabbitmq/blob/66ddebbb7fc778c481da06856e350ceae3e02113/3.8/ubuntu/docker-entrypoint.sh). One example is the templating of the default username and password via environment variables e.g. `RABBITMQ_DEFAULT_PASS_FILE`. By relying on this script, we (and any user) are unable to switch to a different RabbitMQ docker image.

## Expected outcome

In this issue we will track the investigation to determine how much and on what points we rely on the docker entry point script, and propose changes to address this coupling. If the changes required are small enough, they can be implemented as part of this issue.

gerhard:
FWIW, the Docker entry point script will be going away: https://github.com/docker-library/rabbitmq/pull/424

[This is how we handle the default user in the "other" RabbitMQ Docker image](https://github.com/rabbitmq/infrastructure/blob/master/commercial-edition/docker/rmq/default_user.conf). There is a [related TODO](https://github.com/rabbitmq/infrastructure/commit/c9cd2d382f59ac73d4dbeaeb777452458b532098#diff-53cb85b79278630e8a6eefd00e690f0fR149-R150) which @coro just reminded me to follow-up on. Thanks 👍🏻

coro:
I've investigated the entrypoint script, and the only behaviour we rely on from it is the default user/password configuration. It should be fairly simple to transition this to a mounted configMap/`.conf` file, and so I'll do this as part of this issue.

gerhard:
It would be great for the public RabbitMQ Docker image to have the same behaviour. Is this potential PR to https://github.com/docker-library/rabbitmq/ in scope?

For more context, read why now is a great time to start breaking down docker-entrypoint.sh functionality into sysctl config fragments: https://github.com/docker-library/rabbitmq/pull/422#issuecomment-650074731

A separate but somewhat related functionality that will require a different approach to using env vars in the Docker image: https://github.com/rabbitmq/rabbitmq-cli/issues/443. Hint: `--erlang-cookie`

cc @tianon @michaelklishin

Zerpet:
[...]

PR #346, opened 2020-09-16
Configure default user/password through conf.d

coro:
## Summary Of Changes

This commit stops the operator configuring the default username and
password for the cluster through environment variables. Instead, these
credentials are configured through `/etc/rabbitmq/conf.d/`. This is for
two reasons:

* RabbitMQ is moving away from configuration via env vars
* The operator should not be coupled to functionality in the entrypoint script in the docker-library/rabbitmq Docker image, as this prevents other RabbitMQ images being used

This effectively removes support for RabbitMQ images pre 3.8.4, where
the ability to configure RabbitMQ via sysctl-style `*.conf` files in
the conf.d directory.

See #340 

## Local Testing

Unit, Integration & System tests passed. Also tested on kind with `rabbitmqctl authenticate_user`.

```
```

```
$ kubectx dev-bunny
$ make destroy deploy-dev
$ make system-tests
```

### DIFF
--- a/controllers/rabbitmqcluster_controller_test.go
+++ b/controllers/rabbitmqcluster_controller_test.go
@@ -1218,19 +1218,15 @@ var _ = Describe("RabbitmqclusterController", func() {
 					},
 				},
 				corev1.Volume{
-					Name: "rabbitmq-admin",
+					Name: "rabbitmq-confd",
 					VolumeSource: corev1.VolumeSource{
 						Secret: &corev1.SecretVolumeSource{
 							DefaultMode: &defaultMode,
 							SecretName:  "rabbitmq-sts-override-rabbitmq-admin",
 							Items: []corev1.KeyToPath{
 								{
-									Key:  "username",
-									Path: "username",
-								},
-								{
-									Key:  "password",
-									Path: "password",
+									Key:  "default_user.conf",
+									Path: "default_user.conf",
 								},
 							},
 						},
--- a/controllers/rabbitmqcluster_controller_test.go
+++ b/controllers/rabbitmqcluster_controller_test.go
@@ -1207,6 +1207,21 @@ var _ = Describe("RabbitmqclusterController", func() {
 			Expect(sts.Spec.Template.Spec.HostNetwork).To(BeFalse())
 			Expect(sts.Spec.Template.Spec.Volumes).To(ConsistOf(
 				corev1.Volume{
+					Name: "rabbitmq-admin",
+					VolumeSource: corev1.VolumeSource{
+						Secret: &corev1.SecretVolumeSource{
+							DefaultMode: &defaultMode,
+							SecretName:  "rabbitmq-sts-override-rabbitmq-admin",
+							Items: []corev1.KeyToPath{
+								{
+									Key:  "default_user.conf",
+									Path: "default_user.conf",
+								},
+							},
+						},
+					},
+				},
+				corev1.Volume{
 					Name: "additional-config",
 					VolumeSource: corev1.VolumeSource{
 						ConfigMap: &corev1.ConfigMapVolumeSource{
@@ -1220,16 +1235,7 @@ var _ = Describe("RabbitmqclusterController", func() {
 				corev1.Volume{
 					Name: "rabbitmq-confd",
 					VolumeSource: corev1.VolumeSource{
-						Secret: &corev1.SecretVolumeSource{
-							DefaultMode: &defaultMode,
-							SecretName:  "rabbitmq-sts-override-rabbitmq-admin",
-							Items: []corev1.KeyToPath{
-								{
-									Key:  "default_user.conf",
-									Path: "default_user.conf",
-								},
-							},
-						},
+						EmptyDir: &corev1.EmptyDirVolumeSource{},
 					},
 				},
 				corev1.Volume{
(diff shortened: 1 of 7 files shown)


controllers/rabbitmqcluster_controller_test.go
```diff
@@ -1206,6 +1206,21 @@ var _ = Describe("RabbitmqclusterController", func() {
 
 			Expect(sts.Spec.Template.Spec.HostNetwork).To(BeFalse())
 			Expect(sts.Spec.Template.Spec.Volumes).To(ConsistOf(
+				corev1.Volume{
+					Name: "rabbitmq-admin",
```
michaelklishin, line 1210: Unless this user is always supposed to exist and always have administrative privileges, I'd suggest naming this "rabbitmq-default-user".

Zerpet (changes requested):
I would like to see the improvement around projecting the Secret key, instead of copying using the init-container. The refactoring in the tests look good and the changes in the secret resource builder look good.

controllers/rabbitmqcluster_controller_test.go
gerhard: `rabbitmq` with password `rabbitmq` sounds most reasonable to me. I never really understood the reasoning behind `guest`. Do you know where that name came from @michaelklishin?

internal/resource/statefulset.go
```diff
@@ -302,6 +298,12 @@ func (builder *StatefulSetBuilder) podTemplateSpec(annotations, labels map[strin
 				EmptyDir: &corev1.EmptyDirVolumeSource{},
 			},
 		},
+		{
+			Name: "rabbitmq-confd",
+			VolumeSource: corev1.VolumeSource{
+				EmptyDir: &corev1.EmptyDirVolumeSource{},
+			},
+		},
```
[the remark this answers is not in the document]
coro: > Our security context in the template spec ensures that RabbitMQ can read the file, so there is no need to change the ownership in the init container.

Whoa, that's huge. Can this be applied to any of our other initContainer prep? I will try this out.

controllers/rabbitmqcluster_controller_test.go
coro: This user will be admin by default as per the default config for RabbitMQ of `default_user_tags.administrator = true`. As for whether it should be admin by design for the operator, I'd say it's the default user first, that happens to be admin by default unless otherwise configured.

internal/resource/statefulset.go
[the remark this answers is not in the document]
Zerpet: Probably we could apply this to some parts of the init container setup. It **can't** be applied to the Erlang cookie because the Erlang VM expects the cookie to be readable only by the user i.e. `0600` permissions, and given how Kubernetes mounts files (user `root`, group as per security context or `root` if none), we have to copy the cookie and set the expected ownership and permissions.

controllers/rabbitmqcluster_controller_test.go
michaelklishin: Oh, I now understand what the value really is.

From the protocol committee back in the day. "guest" is a reasonable name and is so widely known and documented in the RabbitMQ community that I'd keep it.

So yeah, I'd go with `guest`/`guest` or `administrator`/`administrator` if we intend this user to be the "root" for the cluster.

michaelklishin: If this is the default user, which is there for convenience and testability, I'd say we can go with `guest`/`guest` as widely known credentials specifically for RabbitMQ. The risk is, well, as with all well known credentials :/

coro: Currently in the operator, we override `guest`/`guest` with randomly generated strings. They're stored in a K8s Secret suffixed with `-admin`.

michaelklishin: That sounds good.

coro: You raise a valid point though. The only reason this user is an admin is by the default value of `default_user_tags.administrator`. There's no cluster-specific reason you'd need this user to be an admin; while you may need an admin to create resources or other users on the cluster, you could also just as easily make the default user not an admin through configuration. If you did this, the Status field on the CR would report the 'Admin Credentials' for a user that isn't an admin, and the secret containing the now-non-admin user would be called `XXXXXX-admin`.

I'll raise an issue to switch to referring it as `default_user`, and maybe displaying its permissions in the CR Status field, and we can discuss whether we want to play the story.

coro:
FWIW, I checked the other elements in the initContainer, to see if we could mount them with the correct owner/permissions using projectedVolumes like we did here. Unfortunately it's not suitable for anything that needs to be more than read-only, such as any of our config files. Once https://github.com/kubernetes/kubernetes/issues/81089 is addressed, then it should be possible to mount these correctly without needing an initContainer.

Zerpet (commented):
The latest changes are looking good. I've noticed that we are testing the new behaviour of `controllers/rabbitmqcluster_controller_test.go` in the "Override" section. Functionally this will work, however it doesn't seem to be the right place, as this new behaviour is not specific to the override feature. I think this would fit better in, perhaps, the "it works" section, although that might require some refactor to that test case.

Zerpet (approved):
I (re)reviewed the code and now I see that we have to add an extra case for the `rabbitmq-confd` volume in the override case because we exact-match the elements in the Volume array 🙂 

The feature is tested at unit level in the resource builder, and the integration of the resource builders with the reconcile loop is already tested at integration level. I think this is sufficiently tested and ready to merge ✅

MirahImage (approved):

Issue #338, opened 2020-09-14
Repo still refers to old Docker image in places


coro:
## Describe the bug
The repo references the old Docker image `rabbitmqoperator/rabbitmq-cluster-kubernetes-operator-dev` and `rabbitmqoperator/rabbitmq-cluster-kubernetes-operator` in places, which were replaced by `rabbitmqoperator/cluster-operator-dev` and `rabbitmqoperator/cluster-operator`, respectively.

**Expected behavior**
I would expect both of these image names to be replaced with the new image name. It would be worth investigating where in our repo & tooling the old image is referenced, then updating to the new image name.

## Additional context

Once these deletions are made, we can remove the old images from Dockerhub.

ChunyiLyu:
Remember to update this file as it uses the docker image as a placeholder string when using `kustomize edit image`: https://github.com/pivotal/rabbitmq-for-kubernetes-ci/blob/master/tasks/operator-system-tests/task.yml

Zerpet:
After https://github.com/pivotal/rabbitmq-for-kubernetes-ci/commit/a0a8cd994494048873add181f82390934bcab46e, this issue can be closed.

PR #347, opened 2020-09-17
Update registry references to new name

Zerpet:
This relates to #338 

## Summary Of Changes
We renamed our DockerHub registry and have to reflect the same in the manifests.

It also simplifies the manifest structure by removing some patches that were always applied. If a patch is always applied, it should be part of the resource itself.

Images are not set using kustomize 'image' field, instead of using a patch. Overlays still need to use a patch because they change the image pull policy.

## Additional Context

This PR requires a change into the pipeline to avoid failures in the `operator-system-tests` job. Before merging this PR, mentioned change should occur first.

## Local Testing

✅ Unit tests
✅ System tests


coro (approved):

ChunyiLyu (approved):
LGTM

PR #364, opened 2020-09-28
New command to enable all feature flags

mkuratczyk:
`rabbitmqctl enable_feature_flags` currently doesn't have a `--all` flag. To make it easier to enable all feature flags after an upgrade, I've added a new command to the `kubectl` plugin.


coro (commented):
I like this, but I feel there should be an accompanying enhancement made upstream to make the `--all` flag, rather than it being kubernetes-exclusive behaviour. Do you have any indication of whether the core team are going to introduce this flag? (Or alternatively do you feel like making that PR? ;) )

mkuratczyk:
Yes and yes. The issue is here: https://github.com/rabbitmq/rabbitmq-cli/issues/455
I do intend to PR this.

PR #378, opened 2020-10-01
Put nodes into maintenance mode in preStop hook

### LINKED ISSUE #320: Use maintenance mode for termination
rabbitmq `3.8.8` ships with a newly added maintenance mode. It makes senses for us to use it before shutting down pods for upgrade, or even before any kind of pod restart.

Is this something we only need to run before during upgrade? Should we run it in `preStop` hook for any pod termination?

[Maintenance mode feature PR](https://github.com/rabbitmq/rabbitmq-server/issues/2321)

**Additional context**

coro:
This commit causes nodes to go into maintenance mode before the
controller runtime terminates the rabbitmq-server process on a pod stop.
This allows for a more controlled shutdown of pods in a cluster during
rolling restart, or upgrade.

Using https://github.com/Vanlightly/RabbitTestTool, I was able to show that with
these changes, no data loss, or serious periods of cluster
unavailability, were caused during a rolling restart. However, this is
the case without these changes; I attempted to find a demonstrable case
where this change leads to an improvement in rolling restart
reliability, however was not able to improve this beyond its current
reliability.

I still put this PR forward for review, as the core team recommend that
nodes are put into maintenance mode before being stopped, to allow more
predictable transfer of responsibility to other nodes. The
`rabbitmq-upgrade drain` command is also able to be expanded in the
future if there are further improvements that can be made in the future.

This closes #320.

## Additional Context
All tests pass. Also ran some manual testing with `kubectl rollout restart statefulset <>`

## Local Testing

```
```

```
$ kubectx dev-bunny
$ make destroy deploy-dev
$ make system-tests
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ under `site/kubernetes`.
 
 ## Supported Versions
 
-The operator deploys RabbitMQ `3.8.8` by default, and supports versions from `3.8.4` upwards. The operator requires Kubernetes `1.16` or newer.
+The operator deploys RabbitMQ `3.8.8` by default, and supports versions from `3.8.8` upwards. The operator requires Kubernetes `1.16` or newer.
 
 ## Versioning
 
--- a/internal/resource/statefulset.go
+++ b/internal/resource/statefulset.go
@@ -620,7 +620,8 @@ func (builder *StatefulSetBuilder) podTemplateSpec(annotations, labels map[strin
 								Command: []string{"/bin/bash", "-c",
 									fmt.Sprintf("if [ ! -z \"$(cat /etc/pod-info/%s)\" ]; then exit 0; fi;", DeletionMarker) +
 										fmt.Sprintf(" rabbitmq-upgrade await_online_quorum_plus_one -t %d;"+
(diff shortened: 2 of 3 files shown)


mkuratczyk (approved):
Great stuff. My only comment, and I'm not saying we need to do anything about it, is that we started with a week-long termination timeout but now we pass that value to three different commands so we basically have a 3 week timeout now. I don't think that matters really though - a week basically means infinity and 3*infinity is still infinity ;)

MirahImage (approved):

Zerpet (approved):
Changes look good and tests are passing in my machine ✅

Zerpet:
> Great stuff. My only comment, and I'm not saying we need to do anything about it, is that we started with a week-long termination timeout but now we pass that value to three different commands so we basically have a 3 week timeout now. I don't think that matters really though - a week basically means infinity and 3*infinity is still infinity ;)

The container will receive a [SIGKILL if the grace period expires](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination), even if it's executing the preStop hook; the longest we will wait will be one week (infinity).

PR #379, opened 2020-10-02
Remove embedded descriptions only

### LINKED ISSUE #369: Reintroduce property descriptions for not embedded types
We should  raise an issue and a PR to `controller-gen` to implement description trimming annotation as proposed here:
https://github.com/kubernetes/kubernetes/issues/82292#issuecomment-601851309

Once such a feature is available, we should reintroduce our descriptions removed in https://github.com/rabbitmq/cluster-operator/issues/368

mkuratczyk:
Previously we used maxDescLen to remove all descriptions (to limit the
size of the CRD). This approach keeps descriptions of our properties
while removing them from the `override` part of the CRD which contains
large embedded core types. Size of the CRD is a bit larger than with no
descriptions but still significantly smaller than with all descriptions.

Explanation of what the `jq` command does (I hope ;) ):

* `delpaths()` delete all paths (as in JSONPath) passed as the argument

* `..` convert input (single large object) into a series of entries (each node of the YAML structure is a separate input now)

* `paths(scalars)` give me the paths of the inputs that are scalar values (basicaly leaf nodes of the YAML structure); a path is something like .foo.bar but returned as an array: [“foo”, “bar”]

* `select(contains([“spec”,“versions”,0,“schema”,“openAPIV3Schema”,“properties”,“spec”,“properties”,“override”,“description”]))` select only the paths that contain all the elements listed

Closes #369

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,11 +19,11 @@ integration-tests: install-tools generate fmt vet manifests ## Run integration t
 
 manifests: install-tools ## Generate manifests e.g. CRD, RBAC etc.
 	controller-gen $(CRD_OPTIONS) rbac:roleName=operator-role paths="./api/...;./controllers/..." output:crd:artifacts:config=config/crd/bases
+	./hack/remove-override-descriptions.sh
 	./hack/add-notice-to-yaml.sh config/rbac/role.yaml
 # this is temporary workaround due to issue https://github.com/kubernetes/kubernetes/issues/91395
 # the hack ensures that "protocal" is a required value where this field is listed as x-kubernetes-list-map-keys
 # without the hack, our crd doesn't install on k8s 1.18 because of the issue above
-	./hack/remove-override-descriptions.sh
 	./hack/patch-crd.sh
 	./hack/add-notice-to-yaml.sh config/crd/bases/rabbitmq.com_rabbitmqclusters.yaml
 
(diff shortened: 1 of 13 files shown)


Makefile
```diff
@@ -23,6 +23,7 @@ manifests: install-tools ## Generate manifests e.g. CRD, RBAC etc.
 # this is temporary workaround due to issue https://github.com/kubernetes/kubernetes/issues/91395
 # the hack ensures that "protocal" is a required value where this field is listed as x-kubernetes-list-map-keys
 # without the hack, our crd doesn't install on k8s 1.18 because of the issue above
+	./hack/remove-override-descriptions.sh
```
ChunyiLyu, line 26: The comment no longer matches the command it's for. Would you mind moving `./hack/patch-crd.sh` above `./hack/remove-override-descriptions.sh`?

ChunyiLyu (approved):
One minor thing about comments in Makefile.

Makefile
mkuratczyk: Sure. Done

Zerpet (approved):

Zerpet:
I suspect the error for the system tests in the actions lies around not having `yq` or the expected version of `yq` i.e. python vs golang, hinted by this: https://github.com/rabbitmq/cluster-operator/pull/379/checks?check_run_id=1198945098#step:3:39

PR #386, opened 2020-10-08
Remove expectation on token timeout

Zerpet:

## Summary Of Changes
Remove the expectations around token wait with timeout. This wait is inside the "retry loop", therefore
this function is expected to return false in some cases, which is fine because the loop will retry
shortly after. The expectation around the token error right after the retry loop is correct because
we don't know from the loop if the connection has been successful, therefore we have to assert on the
token success right after the retry loop to guarantee that the connection was a success (or fail the test otherwise).

## Local Testing
System tests have changed one line. They should be green.


coro:
Could this entire loop be replaced with a gomega `Eventually` block? - https://onsi.github.io/gomega/#eventually

system_tests/utils.go
```diff
@@ -749,7 +749,7 @@ func publishAndConsumeMQTTMsg(hostname, nodePort, username, password string, ove
 		fmt.Printf("Attempt #%d to connect using MQTT\n", retry)
 		token = c.Connect()
 		// Waits for the network request to reach the destination and receive a response
-		Expect(token.WaitTimeout(3 * time.Second)).To(BeTrue())
+		token.WaitTimeout(3 * time.Second)
```
coro, line 752: Could this entire loop just be replaced with a Gomega `Eventually` Statement? - https://onsi.github.io/gomega/#eventually

coro (commented):
Could this

system_tests/utils.go
Zerpet: Sure it can 😄 4863a88

coro (approved):

system_tests/utils.go
```diff
@@ -745,18 +745,16 @@ func publishAndConsumeMQTTMsg(hostname, nodePort, username, password string, ove
...
-		fmt.Printf("Attempt #%d to connect using MQTT\n", retry)
+	Eventually(func() bool {
 		token = c.Connect()
 		// Waits for the network request to reach the destination and receive a response
-		Expect(token.WaitTimeout(3 * time.Second)).To(BeTrue())
+		token.WaitTimeout(3 * time.Second)
 
 		if err := token.Error(); err == nil {
-			break
+			return true
```
ansd, line 754: https://github.com/eclipse/paho.mqtt.golang/blob/ba85050a1f239f4e954dc95920213db51f937df1/token.go#L38-L42 states

> In the case of a timeout the Token does not have an error set

Therefore, with the current logic, this expectation would succeed after a 3 seconds timeout.

ansd (changes requested):

system_tests/utils.go
Zerpet: Good catch! I just pushed a new commit checking the return state of `WaitTimeout()`.

ChunyiLyu (approved):
🦖  nice!

PR #388, opened 2020-10-09
Rename admin to default user

### LINKED ISSUE #350: Present the 'admin' user as 'default user with admin permissions'
The operator creates a default user for a RabbitMQ Cluster, with a random username/password. This is done through the configuration in `/etc/rabbitmq/conf.d/default_user.conf` as of #346.

The Secret containing the credentials of this user is presented in the Status field of the RabbitmqCluster CR:
```bash
$ k describe rabbitmqclusters.rabbitmq.com bunny

...
Status:
  Admin:
    Secret Reference:
      Keys:
        Password:  password
        Username:  username
      Name:        config-rabbit-rabbitmq-admin
      Namespace:   rabbitmq-system
   Service Reference:
     Name:       config-rabbit-rabbitmq-client
     Namespace:  rabbitmq-system
```

The issue is, this user doesn't have to be an admin. A user can specify that the default user, for whatever reason, is not admin, by setting `default_user_tags.administrator = false` in their `additionalConfig`. Usually, the administrator tag is present:
```
$ k exec bunny-rabbitmq-server-0 rabbitmqctl list_users
Listing users ...
user    tags
kYbsLQFdSA-Z4ZY6BBitcFdazPeLPowo        [administrator]
```
However, creating a cluster with the following spec:
```yaml
kind: RabbitmqCluster       
metadata:
  name: foo-cluster   
spec:
  replicas: 1    
  rabbitmq:
    additionalConfig: "default_user_tags.administrator = false"
```
creates an 'admin' user that isn't actually an admin!  
```
$ k exec foo-cluster-rabbitmq-server-0 rabbitmqctl list_users
Listing users ...
user    tags
7PD4ho7fjS8EPAoHmn8aLL8pfb8HVK9h        []
```
The status for this CR still reports the creds under 'Admin', however.

Any default user configuration should still be displayed in the Status of the CR, but representing the user (and optionally its permissions / tags) as a default, rather than an admin. For a rough example:

```bash
$ k describe rabbitmqclusters.rabbitmq.com bunny

...
Status:
  Default User:
    Permissions:
      Configure: .*
      Read: .*
      Write: .*
    Tags:
    - administrator
    - management
    - policymaker
    Secret Reference:
      Keys:
        Password:  password
        Username:  username
      Name:        config-rabbit-rabbitmq-default-user
      Namespace:   rabbitmq-system
   Service Reference:
     Name:       config-rabbit-rabbitmq-client
     Namespace:  rabbitmq-system
```
Note the secret name has been changed to have a `-default-user` suffix rather than `-admin`.

ansd:
This closes #350.


MirahImage (approved):
I'm glad we've fixed the URI here too.

ChunyiLyu (approved):
LGTM. I will merge it now!

PR #393, opened 2020-10-19
Use default for replicas

ansd:
Before this PR, when `replicas` is not specified, the following error occurs:
```
error: error validating "cr-example.yaml": error validating data: ValidationError(RabbitmqCluster.spec): missing required field "replicas" in com.rabbitmq.v1beta1.RabbitmqCluster.spec; if you choose to ignore these errors, turn validation off with --validate=false
```
After this PR, the error does not occur and the default specified in https://github.com/rabbitmq/cluster-operator/blob/df0c63074f1f865487222bb1e494c03befd7cab7/api/v1beta1/rabbitmqcluster_types.go#L42 gets used.

### DIFF
--- a/api/v1beta1/rabbitmqcluster_types.go
+++ b/api/v1beta1/rabbitmqcluster_types.go
@@ -38,6 +38,7 @@ type RabbitmqCluster struct {
 // Spec is the desired state of the RabbitmqCluster Custom Resource.
 type RabbitmqClusterSpec struct {
 	// Replicas is the number of nodes in the RabbitMQ cluster. Each node is deployed as a Replica in a StatefulSet. Only 1, 3, 5 replicas clusters are tested.
+	// +optional
 	// +kubebuilder:validation:Minimum:=0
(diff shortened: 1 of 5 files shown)


ferozjilla (approved):
LGTM

Zerpet (approved):

mkuratczyk:
This does really work for me. If you delete these 3 lines https://github.com/rabbitmq/cluster-operator/blob/main/api/v1beta1/rabbitmqcluster_types_test.go#L157-L159 and run `make unit-tests`, this test will actually fail with `RabbitmqCluster.rabbitmq.com "rabbitmq-defaults" is invalid: spec.replicas: Invalid value: "null": spec.replicas in body must be of type integer: "null"` while I would expect it to create a single replica instance. Same result when deployed in a real env (I tried 1.17 and 1.18).

ansd:
@mkuratczyk this is fixed in https://github.com/rabbitmq/cluster-operator/pull/405

PR #394, opened 2020-10-19
Configuration updates

### LINKED ISSUE #225: Better solution for updating configurations

At the moment, the controller uses configMap update as the signal to decide to trigger a statefulSet restart. This approach is very cheap, but has it's downside. If the statefulSet restart was not successful at first reconcile time, even if we requeue the same request, configMap no long needs an update and the controller will not try to restart the statefulSet. We need a better solution so that the restart can be trigger reliably.

ChunyiLyu:
This closes #225 

## Summary Of Changes

- adds a timestamped annotation to the server-conf configmap "rabbitmq.com/serverConfUpdatedAt" when the configMap is updated
- controller compares the timestamp of the server-conf configMap and the sts restart timestamp (if exists)
- if the sts restart timestamp is before the server conf timestamp, or if the server-conf has the timestamp, and the sts doesn't, the controller restarts the sts by annotating the podTemplate (how we trigger the sts restart is the same as before)
- randomized specs for unit tests.
- deleted unused method "UpdateRequiresStsRestart()" from all the resource builders; it was used previously in the controller for sts updates.
- added test cases in integration to ensure annotations are added for sts and server-conf configmap for all rabbitmq configurations updates.
- no new system tests since the use case is the same as before and already covered. 

## Local Testing

Have run unit, integration, and system tests locally.

### DIFF
--- a/controllers/rabbitmqcluster_controller.go
+++ b/controllers/rabbitmqcluster_controller.go
@@ -346,7 +346,7 @@ func (r *RabbitmqClusterReconciler) runPostDeployStepsIfNeeded(ctx context.Conte
 	}
 
 	// Retrieve the plugins config map, if it exists.
-	pluginsConfig, err := r.configMap(ctx, rmq, rmq.ChildResourceName(resource.PluginsConfig))
+	pluginsConfig, err := r.configMap(ctx, rmq, rmq.ChildResourceName(resource.PluginsConfigName))
 	if client.IgnoreNotFound(err) != nil {
 		return 0, err
 	}
@@ -431,7 +431,7 @@ func (r *RabbitmqClusterReconciler) runSetPluginsCommand(ctx context.Context, rm
 // it compares annotation "rabbitmq.com/serverConfUpdatedAt" from server-conf configMap and annotation "rabbitmq.com/lastRestartAt" from sts
 // to determine whether to restart sts
 func (r *RabbitmqClusterReconciler) restartStatefulSetIfNeeded(ctx context.Context, rmq *rabbitmqv1beta1.RabbitmqCluster) (time.Duration, error) {
-	serverConf, err := r.configMap(ctx, rmq, rmq.ChildResourceName(resource.ServerConfigMap))
+	serverConf, err := r.configMap(ctx, rmq, rmq.ChildResourceName(resource.ServerConfigMapName))
 	if err != nil {
 		// requeue request after 10s if unable to find server-conf configmap, else return the error
 		return 10 * time.Second, client.IgnoreNotFound(err)
@@ -506,10 +506,10 @@ func (r *RabbitmqClusterReconciler) annotateConfigMapIfUpdated(ctx context.Conte
 	var configMap, annotationKey string
 	switch builder.(type) {
 	case *resource.RabbitmqPluginsConfigMapBuilder:
-		configMap = rmq.ChildResourceName(resource.PluginsConfig)
+		configMap = rmq.ChildResourceName(resource.PluginsConfigName)
 		annotationKey = pluginsUpdateAnnotation
 	case *resource.ServerConfigMapBuilder:
-		configMap = rmq.ChildResourceName(resource.ServerConfigMap)
+		configMap = rmq.ChildResourceName(resource.ServerConfigMapName)
 		annotationKey = serverConfAnnotation
 	default:
 		return nil
--- a/controllers/rabbitmqcluster_controller.go
+++ b/controllers/rabbitmqcluster_controller.go
@@ -103,13 +103,17 @@ func (r *RabbitmqClusterReconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
 		return ctrl.Result{}, nil
 	}
 
-	// Resource has been marked for deletion
+	// Check if the resource has been marked for deletion
 	if !rabbitmqCluster.ObjectMeta.DeletionTimestamp.IsZero() {
 		logger.Info("Deleting RabbitmqCluster",
 			"namespace", rabbitmqCluster.Namespace,
 			"name", rabbitmqCluster.Name)
-		// Stop reconciliation as the item is being deleted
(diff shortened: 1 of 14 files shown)


controllers/rabbitmqcluster_controller.go
```diff
@@ -419,16 +427,32 @@ func (r *RabbitmqClusterReconciler) runSetPluginsCommand(ctx context.Context, rm
...
-	rmq *rabbitmqv1beta1.RabbitmqCluster) (restarted bool) {
+// Adds an arbitrary annotation to the sts PodTemplate to trigger a sts restart
+// it compares annotation "rabbitmq.com/serverConfUpdatedAt" from server-conf configMap and annotation "rabbitmq.com/lastRestartAt" from sts
+// to determine whether to restart sts
+func (r *RabbitmqClusterReconciler) restartStatefulSetIfNeeded(ctx context.Context, rmq *rabbitmqv1beta1.RabbitmqCluster) (time.Duration, error) {
+	serverConf, err := r.configMap(ctx, rmq, rmq.ChildResourceName(resource.ServerConfigMap))
+	if err != nil {
+		// requeue request after 10s if unable to find server-conf configmap, else return the error
+		return 10 * time.Second, client.IgnoreNotFound(err)
+	}
```
Zerpet, line 438: I'm thinking that we could rewrite this block with `retry.OnError()` [function](https://godoc.org/k8s.io/client-go/util/retry#OnError). It would require to rewrite the helper function so that it only returns an error, and gets the config map into a pointer argument. The main benefit I observe is avoiding a complete requeue of the reconcile function in the scenario where the ConfigMap is taking a bit of time to come up in the Informer, which I think it's not necessary in this case. In the event of an error, we can return the error in the result and the reconcile will requeue automatically.

controllers/rabbitmqcluster_controller.go
```diff
@@ -419,16 +427,32 @@ func (r *RabbitmqClusterReconciler) runSetPluginsCommand(ctx context.Context, rm
...
+	if !ok {
+		// server-conf configmap hasn't been updated; no need to restart sts
+		return 0, nil
+	}
+
+	sts, err := r.statefulSet(ctx, rmq)
+	if err != nil {
+		// requeue request after 10s if unable to find sts, else return the error
+		return 10 * time.Second, client.IgnoreNotFound(err)
+	}
```
Zerpet, line 450: Same as with the ConfigMap ☝️

internal/resource/configmap.go
```diff
@@ -24,7 +24,7 @@ import (
 )
 
 const (
-	serverConfigMapName = "server-conf"
+	ServerConfigMap     = "server-conf"
```
Zerpet, line 27: I'm not sure I agree with this name change. The content of the variable is the server-configuration config map name, not the config map itself.

Zerpet (commented):
Looking good. I made a recommendation around the decision to requeue. Feel free to comment if you disagree with the recommendation.

controllers/rabbitmqcluster_controller_test.go
```diff
@@ -1638,6 +1638,87 @@ var _ = Describe("RabbitmqClusterController", func() {
...
+			assertRabbitmqServerUpdates("additional-config")
+		})
+
+		When("spec.rabbitmq.advancedConfig is updated", func() {
+			assertRabbitmqServerUpdates("advanced-config")
+		})
+
+		When("spec.rabbitmq.envConfig is updated", func() {
+			assertRabbitmqServerUpdates("env-config")
+		})
```
ansd, line 1720: Could use `DescribeTable`

ansd (approved):
Very good!
I like that `UpdateRequiresStsRestart` is gone 🙂

controllers/rabbitmqcluster_controller.go
ChunyiLyu: I see your point Aitor. I think a requeue is a rather cheap thing here compared to adding retry logic. Sure we will jump back to the beginning of the reconcile loop, however it's the same request that we are requeuing, and the request was already processed before. I think adding retry logic is more overhead in this case considering that objects records should be created instantly, so this Get is very unlikely to fail anyways. Thoughts?

internal/resource/configmap.go
ChunyiLyu: Good point. I will change it!

controllers/rabbitmqcluster_controller_test.go
ChunyiLyu: Good catch! Updated

controllers/rabbitmqcluster_controller.go
Zerpet: I guess it depends how we define 'cheap'. If we assume every function/check has cost 1 and running the complete `Reconcile()` function in its happy path (no errors) has cost N, then a requeue here will imply a cost of aproximately 2N, or 2x(3/4 x N) since we are almost at the end of the function. If we add retry logic and every retry is, guesstimate cost of 5, and in the worst case, the retry succeeds in the last attempt e.g. 4th, then the cost of the function would be N + 20, which is equivalent to N, or N + M; in this case, N + M is better than 2N if M < N, which I think it's the case.

ChunyiLyu: > I believe that going through the "update Status" and "resource builders" part is more expensive than retrying a Get() a number of times.

I agree that requeuing is going to be more expensive. The question here is what's the odds of the Get request to fail in this scenario? I would say that Get is **very unlikely** to fail with the error Not Found. The controller can successfully Get/List resources from the k8s api after creating them is an assumption in our controller code. If we decide that the api is unreliable to return resources, then we should have retries in all the Get request that we are performing. I don't see why we should optimize for a very unlikely scenario here.

Zerpet: I don't feel strong about it. I simply think that if we can recover from an error by retrying instead of requeuing, it makes sense to do so. In any case, the outcome will be the same so 🤷‍♂️

Zerpet (approved):
Ginkgo's DescribeTables are awesome! 🙌

PR #403, opened 2020-10-21
Add rabbit memory limit

### LINKED ISSUE #392: Set memory override value in RabbitMQ configuration
### Is your feature request related to a problem?
The community RabbitMQ image in DockerHub sets the total memory available to the node in the [entry point script](https://github.com/docker-library/rabbitmq/blob/master/3.8/ubuntu/docker-entrypoint.sh#L300-L321), matching the memory defined by the cgroups. We want to stop relying on any behaviour that is implemented via this entry point script and make any relevant configuration part of the Operator configuration. This [PR from rabbitmq-server](https://github.com/rabbitmq/rabbitmq-server/pull/1234) delivered the ability to set the property `total_memory_available_override_value` to configure how much memory is available to RabbitMQ. By default, RabbitMQ will assume all the memory available in the current host. This becomes a problem in the context of containers because RabbitMQ will, incorrectly, assume that it has all the memory available in the host, however the memory will be limited by either cgroups or Kubernetes resource limits.

### Describe the solution you'd like
Automatically configure the value of `total_memory_available_override_value` based on the value in `.spec.resources.limits.memory`. If this property is explicitly set to empty object i.e. unlimited, we will skip setting it and let the container assume it has all the memory available in the kubernetes node.

MirahImage:
This closes #392

**Note to reviewers:** the commits should be squashed before merging

## Summary Of Changes
Sets `total_memory_available_override_value` in rabbitmq.conf if there is a memory resource limit.

### DIFF
--- a/api/v1beta1/rabbitmqcluster_types.go
+++ b/api/v1beta1/rabbitmqcluster_types.go
@@ -314,6 +314,10 @@ func (cluster *RabbitmqCluster) MutualTLSEnabled() bool {
 	return cluster.TLSEnabled() && cluster.Spec.TLS.CaSecretName != ""
 }
 
+func (cluster *RabbitmqCluster) MemoryLimitProvided() bool {
+	return cluster.Spec.Resources != nil && cluster.Spec.Resources.Limits != nil && !cluster.Spec.Resources.Limits.Memory().IsZero()
+}
+
 func (cluster *RabbitmqCluster) SingleTLSSecret() bool {
 	return cluster.MutualTLSEnabled() && cluster.Spec.TLS.CaSecretName == cluster.Spec.TLS.SecretName
 }
--- a/api/v1beta1/rabbitmqcluster_types.go
+++ b/api/v1beta1/rabbitmqcluster_types.go
@@ -314,7 +314,7 @@ func (cluster *RabbitmqCluster) MutualTLSEnabled() bool {
 	return cluster.TLSEnabled() && cluster.Spec.TLS.CaSecretName != ""
 }
 
-func (cluster *RabbitmqCluster) MemoryLimitProvided() bool {
+func (cluster *RabbitmqCluster) MemoryLimited() bool {
 	return cluster.Spec.Resources != nil && cluster.Spec.Resources.Limits != nil && !cluster.Spec.Resources.Limits.Memory().IsZero()
 }
 
--- a/internal/resource/configmap.go
+++ b/internal/resource/configmap.go
@@ -12,6 +12,7 @@ package resource
 import (
 	"bytes"
 	"fmt"
+	"strconv"
 	"strings"
 
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
@@ -90,8 +91,11 @@ func (builder *ServerConfigMapBuilder) Update(object runtime.Object) error {
 	}
 
 	if builder.Instance.MemoryLimitProvided() {
-		if _, err := defaultSection.NewKey("total_memory_available_override_value",
-			formatMemory(builder.Instance.Spec.Resources.Limits.Memory().String())); err != nil {
+		memLimit, err := formatMemory(builder.Instance.Spec.Resources.Limits.Memory().String())
+		if err != nil {
+			return err
+		}
+		if _, err := defaultSection.NewKey("total_memory_available_override_value", memLimit); err != nil {
 			return err
 		}
 	}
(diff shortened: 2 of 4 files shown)


MirahImage:
Kubernetes supports E, P, T, G, M, and K as decimal units and Ei, Pi, Ti, Gi, Mi, and Ki as base 2 units for memory, per https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-memory

RabbitMQ supports kB, MB, and GB as decimal units and kiB, MiB, and GiB as base 2 units for memory, as per https://github.com/rabbitmq/rabbitmq-server/blob/1b976358b62db552f70a0f6c273431895e3cddb0/docs/rabbitmq.config.example#L230

Due to the limited RabbitMQ support for large units, and the unusual capitalization of kB and kiB, the formatting function is longer than would otherwise be idea.

mkuratczyk:
I was talking to Gerhard today and learned that Erlang is still not playing nicely with container memory limits. He suggested that the value set for Erlang should be lower than what is assigned to the pod to allow Erlang to exceed what is set in rabbitmq.conf temporarily, without getting OOMkilled.

MirahImage:
I don't think we should really be building Erlang VM workarounds into the cluster operator, especially since it would involve SIGNIFICANT string manipulations.

gerhard:
This is a grey area that costs some number of grey hairs each time it is discussed. I will keep it short & simple.

A high message ingress rate (steady or sudden), will make RabbitMQ require more memory, which in turn makes the Erlang VM request memory from the OS. Erlang VM requests memory in stages, it can happen multiple times in the span of a few seconds, and the requests use growth stages, each stage requesting more memory than the previous one.

RabbitMQ has this component called the memory monitor which runs every 5 seconds and reads the RSS memory that the Erlang system process uses. If it uses more than the memory limit, the memory alarm gets triggered and a bunch of things happen, including stop reading from all TCP sockets that publish messages. This is normally sufficient to flush as much to disk as possible, and bring the memory usage down. In reality, it is more complicated than this, but there is nothing that we can do about this in this context.

The question becomes: how much extra memory should we allow the pod to use, vs what we tell RabbitMQ that it is allowed to use, so that we account for Erlang requesting memory that it shouldn't, and RabbitMQ noticing the actual memory usage, which may be delayed by at most 5 seconds.

To put it in a picture, we want to prevent the pod getting OOMkilled when this happens:
image

This sounds like a good first strategy (we will most certainly refine it): **limit RabbitMQ memory to 80% of what the pod is allowed to use, but no more than 2GB**. This is effectively memory headroom, so somewhat wasted, but it's the only thing that stands between a killed RabbitMQ node, and a node that can recover from a sudden (or steady) high message ingress.

MirahImage:
Implementing the headroom for the erlang VM actually drove me to a much simpler implementation of the memory limit, which is nice and makes everything much more readable. Additionally, it will be easy to remove if this problem ever goes away.

internal/resource/configmap.go
```diff
@@ -129,3 +136,11 @@ func updateProperty(configMapData map[string]string, key string, value string) {
 		configMapData[key] = value
 	}
 }
+
+func removeHeadroom(memLimit int64) int64 {
+	const GiB int64 = 1073741824
+	if memLimit/5 > 2*GiB {
+		return memLimit - 2*GiB
+	}
+	return memLimit - memLimit/5
```
Zerpet, line 145: Another neat pick: perhaps add a comment here stating that headroom is lowest between 20% of the amount passed in as argument or 2GiB?
I had to look at this for a bit to figure out why we were dividing by 5 😄

Zerpet (commented):
Left two neat picks there. Everything else looks good.

api/v1beta1/rabbitmqcluster_types.go
```diff
@@ -314,6 +314,10 @@ func (cluster *RabbitmqCluster) MutualTLSEnabled() bool {
 	return cluster.TLSEnabled() && cluster.Spec.TLS.CaSecretName != ""
 }
 
+func (cluster *RabbitmqCluster) MemoryLimitProvided() bool {
+	return cluster.Spec.Resources != nil && cluster.Spec.Resources.Limits != nil && !cluster.Spec.Resources.Limits.Memory().IsZero()
+}
```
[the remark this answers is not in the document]
MirahImage: I followed the convention of the other helper methods on this object, none of which follow the `Is` convention

internal/resource/configmap.go
MirahImage: Have added a comment to clarify why this exists and what it does

mkuratczyk (approved):

ChunyiLyu (approved):
Looks good! I suggest squash all commits when merging since it's related to one feature.

MirahImage:
> Looks good! I suggest squash all commits when merging since it's related to one feature.

Yep, that was the plan, especially as the entire implementation changed halfway through the PR

internal/resource/configmap.go
ktzsolt: Late to the party, but this "smaller amount of 20% memory or 2GiB" headroom should be documented, it was driving me crazy where my precious 20% of memory went :)

mkuratczyk: It is documented:
https://www.rabbitmq.com/kubernetes/operator/using-operator#resource-reqs

ktzsolt: Ah I see now, couldn't find it, I was searching in the memory settings of rabbitmq docs but I also used the search bar to search for "20%" but that also didn't find this. Thanks you!

PR #405, opened 2020-10-22
Omit replicas from encoding when empty

### LINKED ISSUE #404: spec.replicas defaulting does not work
## Describe the bug

The CRD has default value set for `spec.replicas` and it is an optional value. However, you cannot create a RabbitmqCluster with no `spec.replicas` specified.

## To Reproduce

`kubectl apply ` with a RabbitmqCluster manifest with no spec.replicas set

**Expected behavior**

1) If the value is set to be `optional`, we should be able to create RabbitmqCluster without specifying it.
2) We change it back to a required value. Q: why do we have default value for this required property?

ansd:
Close #404

### DIFF
--- a/api/v1beta1/rabbitmqcluster_types.go
+++ b/api/v1beta1/rabbitmqcluster_types.go
@@ -41,7 +41,7 @@ type RabbitmqClusterSpec struct {
 	// +optional
 	// +kubebuilder:validation:Minimum:=0
 	// +kubebuilder:default:=1
-	Replicas *int32 `json:"replicas"`
+	Replicas *int32 `json:"replicas,omitempty"`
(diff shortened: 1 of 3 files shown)


coro (approved):
LGTM. Theoretical fix, or have you tested live?

ansd:
I tested it :)
Actually, good point, I'll push a commit where we can remove https://github.com/rabbitmq/cluster-operator/blob/396e5c6d16abc13e1b430802de512e9e7b18fb1c/api/v1beta1/rabbitmqcluster_types_test.go#L167-L169

Zerpet (approved):
:shipitparrot:

Zerpet (approved):

PR #408, opened 2020-10-23
Set topologySpreadConstraints by default

### LINKED ISSUE #398: Set `topologySpreadConstraints` by default

ChunyiLyu:
This closes #398

## Summary Of Changes

- "topology.kubernetes.io/zone" is a popular label set by kubelet on worker nodes
- "ScheduleAnyways" makes sure that this constraint is treated as a soft requirement, and pods still get scheduled without meeting the requirement
- using hostname as topologyKey is also a valid contraint. However, the official docs recommends not to set multiple contraints if possible to aviod conflicts

## Local Testing

Added new test in unit tests to ensure on the templating level. Have run unit, integration, and system tests

### DIFF
--- a/internal/resource/statefulset.go
+++ b/internal/resource/statefulset.go
@@ -491,6 +491,19 @@ func (builder *StatefulSetBuilder) podTemplateSpec(annotations, labels map[strin
 			Labels:      labels,
 		},
 		Spec: corev1.PodSpec{
+			TopologySpreadConstraints: []corev1.TopologySpreadConstraint{
+				{
+					MaxSkew: 1,
+					// "topology.kubernetes.io/zone" is a well-known label.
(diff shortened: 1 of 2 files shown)


MirahImage (approved):

mkuratczyk (approved):
Thanks!

Issue #414, opened 2020-10-27
Enable all feature flags when deploying a new cluster


mkuratczyk:
Some feature flags are disabled by default because RabbitMQ itself cannot tell whether it's starting a fresh node or not at the time it needs to decide the state of a feature flag. To be on the safe side - those flags are disabled. The operator however, can distinguish between a fresh cluster deployment and update/upgrade and should therefore enable all feature flags when deploying a new cluster so that all features are available by default. At this time, these disabled-by-default feature flags are `drop_unroutable_metric` and `empty_basic_get_metric` but ideally, we should make the implementation generic so that we always enable all feature flags when deploying a new cluster.

PR #418, opened 2020-10-27
Break down rabbitmqcluster_controller.go

ChunyiLyu:

## Summary Of Changes

This is a refactor PR which separates logics of controllers/rabbitmqcluster_controller.go. I extracted helper functions according to different concepts/objects that the `Reconcile()` is operating on. 

- reconcile_finalizer.go
- reconcile_queue_rebalance.go
- reconcile_rabbitmq_configurations.go
- reconcile_status_.go
- reconcile_tls.go

Each of the file has a corresponding `reconcile_*_test.go` which contains its integration tests. I add an additional `utils.go` which contained general helper functions that were used in multiple of the above files. `PodExecutor` interface is extracted to its own file as well. `rabbitmqcluster_controller.go` still containers the `Reconcile()`, `SetupWithManager()`, and helper function `addResourceToIndex`, `getRabbitmqCluster`, `getChildResources` and `logAndRecordOperationResult`.

## Pros and Cons

I think the controller logic became easier to navigate with this refactor. When we are modifying `rabbitmqcluster_controller.go`, most of the time we are changing the logic of how the `Reconcile()` deals with one object/aspects. With this refactor, it should be easier to see where to make changes. Each of these files can grow or shrink themselves without impacting the main controller logic.

The obvious cons will be losing history behind the helper function. However, I would argue that we will eventually do a refactor and the earlier we do it, the more history we can preserve.

## Next Step

This is "shift and life" PR that does not change any helper function or any test. I think the logical step next will be to refactor all helper reconcile to have the same contract, and to further move certain logic from `Reconcile()` to the helpers. Before I head there, I would like to check in with the team first.

## Local Testing

Have run unit and integration tests.


MirahImage:
I'm really happy we're doing this, the controller codebase was getting difficult to navigate.

MirahImage (approved):

mkuratczyk (approved):
Great stuff. I feel like this queue rebalancing could be done in a more generic way (it doesn't really reconciles anything and it's not clear how another post-deploy command would fit into this). But I'm totally happy to take this PR as-is and worry about that when working on https://github.com/rabbitmq/cluster-operator/issues/414, which is another post deploy command (but only executed on fresh deployments).

ferozjilla (approved):
LGTM

ablease:
Thanks for this Chunyi. This solves some of the navigation issues I had with the controller also.

ablease (approved):

ansd (approved):

Zerpet (approved):

coro (approved):
Thanks for this refactor!

PR #424, opened 2020-10-27
Make TerminationGracePeriodSeconds configurable

### LINKED ISSUE #419: Make terminationGracePeriodSeconds configuration at rabbitmqcluster.spec
**Context**

As discussed in sync-up, it's an important context that we should make people aware that we intentionally set terminationGracePeriodSeconds to a week long to ensure that no data loss during a restart of pods. We decided to 1) document this week long timeout and make people aware that this could cause long termination for pods 2) make this timeout configurable at the top level since we are aware that not all people want this feature.

```yaml
kind: RabbitmqCluster
spec:
  terminationGracePeriodSeconds: # configurable and [defaults to 604800](https://github.com/rabbitmq/cluster-operator/blob/main/internal/resource/statefulset.go#L30)
```

**Additional context**

ChunyiLyu:
This closes #419 

## Summary Of Changes

- make TerminationGracePeriodSeconds configurable
- refactor to use "k8s.io/utils/pointer" library everywhere

## Local Testing

Have run unit, integration, and system tests


ansd:
@ChunyiLyu could you add more details to the PR or issue as to why this value is made configurable?

ChunyiLyu:
@ansd added more context in the issue.

internal/resource/statefulset_test.go
```diff
@@ -1011,12 +1011,19 @@ var _ = Describe("StatefulSet", func() {
 			}))
 		})
 
-		It("adds the required terminationGracePeriodSeconds", func() {
+		It("sets TerminationGracePeriodSeconds in podTemplate as provided in instance spec", func() {
+			ten := int64(10)
+			instance.Spec.TerminationGracePeriodSeconds = &ten
```
ansd, line 1016: There are a couple of places in this PR where you could use the `"k8s.io/utils/pointer"` package if you want to as done in https://github.com/rabbitmq/cluster-operator/blob/01b368312c24625bb74efbbe5bd860534b9d43a1/internal/resource/statefulset.go#L524

ansd (approved):

internal/resource/statefulset_test.go
ChunyiLyu: Good point. On it!

api/v1beta1/rabbitmqcluster_types.go
```diff
@@ -63,6 +63,12 @@ type RabbitmqClusterSpec struct {
 	// Has no effect if the cluster only consists of one node.
 	// For more information, see https://www.rabbitmq.com/rabbitmq-queues.8.html#rebalance
 	SkipPostDeploySteps bool `json:"skipPostDeploySteps,omitempty"`
+	// TerminationGracePeriodSeconds is the timeout that each rabbitmqcluster pod will have to terminate gracefully
+	// It defaults to 604800 seconds ( a week long) to ensure that the container preStop lifecycle hook can finish running
+	// For more information, see: https://github.com/rabbitmq/cluster-operator/blob/main/docs/design/20200520-graceful-pod-termination.md
+	// +kubebuilder:validation:Minimum:=0
+	// +kubebuilder:default:=604800
+	TerminationGracePeriodSeconds *int64 `json:"terminationGracePeriodSeconds,omitempty"`
```
Zerpet, line 71: Does `omitempty` have any effect? Given that we set a default and value 0 is also an acceptable value, I think this field will never be empty.

Zerpet (commented):
I'm thinking that we should adapt the timeout in the preStop hooks to match the termination grace period i.e. `rabbitmq-upgrade await_online_quorum_plus_one -t 604800`. And perhaps remove the preStop hook altogether if the grace period is set to 0. 

What do you think?

api/v1beta1/rabbitmqcluster_types.go
ChunyiLyu: Without `omitempty`, defaulting won't work here. See [commit](https://github.com/rabbitmq/cluster-operator/pull/405/commits/f564031c706c2b536976f03bfe183a285062d5fd)

ChunyiLyu:
Merge conflicts in this branch is hard to resolve. Moved to PR #430 

Closing

PR #429, opened 2020-10-29
Format expected config file with ini pkg

mkuratczyk:
This way the tests are not sensitive to whitespace differences since they compare actual config
file with the expected value after formatting using the same method as the actual code

To test, you can add/remove some spaces in `expectedRabbitmqConf` and the tests should still pass.

### DIFF
--- a/internal/resource/configmap_test.go
+++ b/internal/resource/configmap_test.go
@@ -382,6 +382,6 @@ func iniString(input string) string {
 	var output bytes.Buffer
 	cfg, _ := ini.Load([]byte(input))
 	_, err := cfg.WriteTo(&output)
-	Expect(err).NotTo(HaveOccurred())
+	ExpectWithOffset(1, err).NotTo(HaveOccurred())
 	return output.String()
 }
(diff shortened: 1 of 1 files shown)


MirahImage (approved):

internal/resource/configmap_test.go
```diff
@@ -362,3 +375,13 @@ ssl_options.verify                       = verify_peer`)))
...
 	})
 })
+
+// iniString formats the input string using "gopkg.in/ini.v1"
+func iniString(input string) string {
+	ini.PrettySection = false
+	var output bytes.Buffer
+	cfg, _ := ini.Load([]byte(input))
+	_, err := cfg.WriteTo(&output)
+	Expect(err).NotTo(HaveOccurred())
```
ChunyiLyu, line 385: You can use [`ExpectWithOffset(1,...)` ](https://godoc.org/github.com/onsi/gomega#ExpectWithOffset) here so that the stack trace can point to where this helper is called :)

ChunyiLyu (approved):
Great! Just one small thing with the helper function assertion.

PR #431, opened 2020-10-29
Renaming

### LINKED ISSUE #415: Drop "headless" from the service name
We've heard feedback that the word "headless" in the service name is confusing - it's a very technical aspect of the service that leaks into many places, including all names in the Prometheus metrics. It also makes all names quite a bit longer (again, that mostly affects for monitoring). Unless we see a good reason to keep it, we should just drop "headless" from the service name.

ChunyiLyu:
This closes #415 

**Note to reviewers:** Don't mind the messy commit history, all commits will be squashed when merging

## Summary Of Changes

- Rename the suffix "-headless" for the headless svc to `-nodes` since the suffix is unnecessary and exposes implementation details to users
- The new commit also drops "rabbitmq" from child resources suffix. We are already labelling all resources with the CR name and the component name was `rabbitmq`. This suffix seems rather redundant
- Drop the suffix `-client` for the external rmq svc

## Local Testing

Have run unit, integration, and system tests locally


MirahImage (approved):

ChunyiLyu:
@Zerpet I was thinking about the same. Maybe "name-rabbitmq" as a svc name is confusing to people. I will ask in standup Monday.

Zerpet:
After giving a read to the [StatefulSet docs](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#stable-network-id) about the Pod's network ID, I think it's ok to leave the service name as `<cr-name>-rabbitmq` because 1) it's short and unique and 2) does not leak implementation detail e.g. clustering, headless, etc. The examples in the linked doc also show examples with no suffixes, just short and right on point. In our case, we need the uniqueness provided by prefixing the CR name.

Perhaps we could append `-cluster`, so it would result in `<cr-name>-rabbitmq-cluster`. Given that the purpose of this service is to provide network identifiers for RabbitMQ Pods, I think this name makes its clear that is related to the RabbitMQ cluster, and alongside with the Client service, it won't be confused as "ingress" for the RabbitMQ cluster.

mkuratczyk (approved):
I like these changes. If we can come up with a better name for the client service, that would be awesome but that could be a separate PR anyway

ansd (approved):

ChunyiLyu:
Closing and created #447

PR #434, opened 2020-10-30
Enable all feature flags for new clusters

### LINKED ISSUE #414: Enable all feature flags when deploying a new cluster

ansd:
Closes #414 

**Notes:**
* The controller identifies a new RabbitMQ cluster by the operation result `created` of the StatefulSet.
* A new annotation `rabbitmq.com/createdAt` is added onto the created StatefulSet.
* Once all pods are ready, CLI command `rabbitmqctl enable_feature_flag` is used on the first pod of the StatefulSet to enable all stable and previously disabled feature flags. (In future, we could add feature flag methods to https://github.com/michaelklishin/rabbit-hole/ to not exec into the pod anymore.)

### DIFF
--- a/controllers/rabbitmqcluster_controller.go
+++ b/controllers/rabbitmqcluster_controller.go
@@ -14,14 +14,15 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"reflect"
+	"strings"
+
 	"github.com/rabbitmq/cluster-operator/internal/resource"
 	"github.com/rabbitmq/cluster-operator/internal/status"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/record"
-	"reflect"
-	"strings"
 
 	"k8s.io/apimachinery/pkg/types"
 
@@ -182,7 +183,7 @@ func (r *RabbitmqClusterReconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
 			return ctrl.Result{}, err
 		}
 
-		if err = r.annotateConfigMapIfUpdated(ctx, builder, operationResult, rabbitmqCluster); err != nil {
+		if err = r.annotateIfNeeded(ctx, builder, operationResult, rabbitmqCluster); err != nil {
 			return ctrl.Result{}, err
 		}
 	}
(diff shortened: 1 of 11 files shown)


MirahImage (approved):

Zerpet (commented):
I'm not sure if we should move the set plugins functionality to "post deploy", given that there are cases where the plugins might be updated without a (re)deploy. For example, by updating the CR with `rabbitmq_federation`, the plugin is enabled and the cluster should not be restarted.

coro (approved):
Great changes, and thanks for the great refactoring along the way!

ansd:
@Zerpet good point, previously we ran "post deploy" at the end of every successful reconcile loop.
I added https://github.com/rabbitmq/cluster-operator/pull/434/commits/63a5ae3117cbf0222d2a610ad6456237364befbe  which renames the file and the function from "post deploy" to "CLI" because at the end of every reconcile loop we want to check whether there are any annotations which tell us that we need to run RabbitMQ CLI commands (such as setting plugins, rebalancing queues (which is a "post deploy" step), enabling feature flags). 

The post deploy flag https://github.com/rabbitmq/cluster-operator/blob/7244faa67827efe12607572dcea48fa952843b96/api/v1beta1/rabbitmqcluster_types.go#L65 still exists and can be extended with other RabbitMQ CLI commands in future.

mkuratczyk:
I think there is something wrong with the functionality.

It works if I just `deploy-dev` from the PR branch and create some clusters - feature flags get enabled.

However, I wanted to make sure it would not enable feature flags for pre-existing clusters so I deployed from `main`, created a cluster and then deployed from the branch. In that case feature flags were not enabled for **either** of the clusters - even a newly created cluster did not get them enabled.

mkuratczyk:
Please ignore my comment above. Because we removed double prefix in `main`, I ended up with two operator running (the old one did the reconciliation)

Zerpet (approved):
Latest changes look good :shipit:

PR #443, opened 2020-11-03
Use RabbitMQ management image variant

### LINKED ISSUE #441: Switch to RabbitMQ management docker image variant
The non-management image variant disabled management agent metric collection. This configuration is set in `conf.d` and it takes precedence over any configuration set in `rabbitmq.conf`. This means that we can't override this configuration using the `RabbitmqCluster` spec.

The image variant `3.8-management` comes with the management plugin enabled and it does not disable the metric collection. Given that we want to always provide the management UI, it makes sense that we use this image instead. We should set this image by default in the Operator.

N/A

**Additional context**
Discussed in weekly sync up ✌️

Zerpet:
This closes #441

## Summary Of Changes

- Changed the default RabbitMQ image to use tag `3.8.9-management`

## Additional Context

The non-management image variant disables metrics collection by the management agent, resulting in very little information
shown in the management UI. We want to always enable the management UI as it is a common entry point to RabbitMQ
administration. It's also not straightforward to re-enable the metric collection in the non-management variant.

## Local Testing

- Made changes to unit tests. Those are passing.
- No changes to integration or system tests. Those are also green.

### DIFF
--- a/api/v1beta1/rabbitmqcluster_types.go
+++ b/api/v1beta1/rabbitmqcluster_types.go
@@ -43,7 +43,7 @@ type RabbitmqClusterSpec struct {
 	// +kubebuilder:default:=1
 	Replicas *int32 `json:"replicas,omitempty"`
 	// Image is the name of the RabbitMQ docker image to use for RabbitMQ nodes in the RabbitmqCluster.
-	// +kubebuilder:default:="rabbitmq:3.8.9"
+	// +kubebuilder:default:="rabbitmq:3.8.9-management"
 	Image string `json:"image,omitempty"`
 	// List of Secret resource containing access credentials to the registry for the RabbitMQ image. Required if the docker registry is private.
 	ImagePullSecrets []corev1.LocalObjectReference `json:"imagePullSecrets,omitempty"`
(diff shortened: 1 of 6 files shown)


ChunyiLyu (approved):
Looks good. One small thing: this is something worth mentioning in README and docs? Your call.

ansd (approved):
We could also modify the example in https://github.com/rabbitmq/cluster-operator/blob/d4b61299d2b5b7674c77a7dc1997ba37edb58d53/charts/rabbitmq/example-configurations.yaml#L20

Zerpet:
We don't mention the management UI in the Readme or in the [docs](https://www.rabbitmq.com/kubernetes/operator/using-operator.html#images). In the docs we actually make a vague reference to the "community RabbitMQ image on dockerhub". I don't think we need to be explicit because of this change.

@ansd I blame `ag rabbitmq:3.8.9` for comming out empty 😎  Will do this change.

ablease:
No need to squash IMO. One commit for default image change, another for helm chart.

Zerpet:
Two approvals, checks are green. Merging! 🚀

Issue #440, opened 2020-11-03
Release and test `kubectl-rabbitmq`


ChunyiLyu:
**Is your feature request related to a problem? Please describe.**

We would like users of rabbitmqcluster to start using the cli tool that we have developed. We would like to test the cli tool, and making sure that it is compatible with the release for usability.

**Describe the solution you'd like**

- `kubectl-rabbitmq version` should return the version of the operator that the cli is tested and released with.
-  cli is tested in the CI
-  cli is released as an artifact in the release

**Describe alternatives you've considered**

**Additional context**

ansd:
Done:
* Add tests https://github.com/rabbitmq/cluster-operator/pull/445
* Run tests in CI https://github.com/rabbitmq/cluster-operator/pull/445
* Release https://github.com/kubernetes-sigs/krew-index/pull/895

Todo:
* Release automation - add GitHub action, see https://krew.sigs.k8s.io/docs/developer-guide/release/automating-updates/
* Add version command to plugin

PR #445, opened 2020-11-06
Add tests for kubectl plugin

ansd:
Relates to #440.

### DIFF
--- a/bin/kubectl-rabbitmq
+++ b/bin/kubectl-rabbitmq
@@ -73,9 +73,9 @@ END
 
 get_instance_details() {
     instance=${1}
-    username=$(kubectl get secret "${instance}-rabbitmq-default-user" -o jsonpath="{.data.username}" | base64 --decode)
-    password=$(kubectl get secret "${instance}-rabbitmq-default-user" -o jsonpath="{.data.password}" | base64 --decode)
-    service=${instance}-rabbitmq-client
+    username=$(kubectl get secret "${instance}-default-user" -o jsonpath="{.data.username}" | base64 --decode)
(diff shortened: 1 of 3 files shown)


bin/kubectl-rabbitmq
```diff
@@ -286,13 +290,21 @@ main() {
...
     "install-cluster-operator")
         shift 1
-        if [[ "$#" -gt 1 ]]; then
+        if [[ "$#" -gt 0 ]]; then
             usage
             exit 1
         fi
-        install_cluster_operator "${1:-}"
+        install_cluster_operator
+        ;;
```
Zerpet, line 298: I think this was used to install the Operator using a custom/relocated image. Are we removing support for that feature?

ansd: This feature was removed in https://github.com/rabbitmq/cluster-operator/commit/d5444f53103d3bba91050390b907c6222338eb32. If we want that feature, we could add it back after this PR has been merged?

coro (approved):
LGTM. Thanks for adding in the extra options for helptext - that always annoys me when one is missing. Also thanks for the note about GH Actions CPU and the `--unlimited` flag, it's really good to keep that context in the test.

bin/kubectl-rabbitmq
ansd: @Zerpet, I created https://github.com/rabbitmq/cluster-operator/issues/446 which would require this flag. So, we can add it back when working on that issue.

Zerpet: Ah! I wasn't aware of d5444f53103d3bba91050390b907c6222338eb32, my bad 🙂 I'm happy to conside re-adding this feature as part of #446.

Zerpet (approved):

PR #447, opened 2020-11-06
Rename

### LINKED ISSUE #415: Drop "headless" from the service name

ChunyiLyu:
This closes #415

It's a duplicate of #431, re created this branch because of the messy history

## Summary Of Changes

- Rename the suffix "-headless" for the headless svc to -nodes since the suffix is unnecessary and exposes implementation details to users
- Drop "rabbitmq" from child resources suffix. We are already labelling all resources with the CR name - and the component name was rabbitmq. This suffix seems rather redundant
- Drop the suffix -client for the external rmq svc. Also rename client_service.go to service.go, since I find it confusing. override.clientService is renamed to override.service as well.

## Local Testing

Have run unit, integration, and system tests


mkuratczyk (approved):
Awesome. Thanks!

ansd (approved):

PR #449, opened 2020-11-09
Stop triggering pod restarts for CR labels/annotations updates

### LINKED ISSUE #442: Setting labels via metadata.label is incompatible with Flux
We use Flux for our GitOps (https://github.com/fluxcd/flux2) pipeline. Flux sets a couple of labels for each object/resource under its control. One of it is the current git commit id which changes with every Flux reconciliation that gets triggered by changes in the source repository. Currently, this triggers a restart of all RabbitMQ pods.

I'd propose to set the labels that affect the ReplicaSet/Pods under `spec.template` and not directly under `metadata`.

ChunyiLyu:
This closes #442 

## Summary Of Changes
-  Stop populate CR labels/annotations to pods. Instance labels and annotations are still populated and updated for statefulSet.
- Stop updating label/annotation on server conf configMap, because updates to server-conf configMap triggers pod restart. Instance labels and annotations are still applied at creation.
Stop updating label/annotation on plugins configmap, because updates to the plugins configMap triggers unnecessary `rabbitmq-plugins set`.

## Local Testing

### DIFF
--- a/controllers/rabbitmqcluster_controller_test.go
+++ b/controllers/rabbitmqcluster_controller_test.go
@@ -534,34 +534,13 @@ var _ = Describe("RabbitmqClusterController", func() {
 				}, 3).Should(HaveKeyWithValue(annotationKey, annotationValue))
 			})
 
-			It("updates annotations for config maps", func() {
-				Eventually(func() map[string]string {
-					roleBinding, err := clientSet.CoreV1().ConfigMaps(cluster.Namespace).Get(ctx, cluster.ChildResourceName("server-conf"), metav1.GetOptions{})
-					Expect(err).NotTo(HaveOccurred())
-					return roleBinding.Annotations
-				}, 3).Should(HaveKeyWithValue(annotationKey, annotationValue))
-
+			It("updates annotations for plugins config map", func() {
 				Eventually(func() map[string]string {
 					roleBinding, err := clientSet.CoreV1().ConfigMaps(cluster.Namespace).Get(ctx, cluster.ChildResourceName("plugins-conf"), metav1.GetOptions{})
 					Expect(err).NotTo(HaveOccurred())
 					return roleBinding.Annotations
 				}, 3).Should(HaveKeyWithValue(annotationKey, annotationValue))
 			})
-
-			It("updates annotations for config maps", func() {
-				Eventually(func() map[string]string {
-					roleBinding, err := clientSet.CoreV1().ConfigMaps(cluster.Namespace).Get(ctx, cluster.ChildResourceName("server-conf"), metav1.GetOptions{})
-					Expect(err).NotTo(HaveOccurred())
-					return roleBinding.Annotations
-				}, 3).Should(HaveKeyWithValue(annotationKey, annotationValue))
-
-				Eventually(func() map[string]string {
-					roleBinding, err := clientSet.CoreV1().ConfigMaps(cluster.Namespace).Get(ctx, cluster.ChildResourceName("plugins-conf"), metav1.GetOptions{})
-					Expect(err).NotTo(HaveOccurred())
-					return roleBinding.Annotations
-				}, 3).Should(HaveKeyWithValue(annotationKey, annotationValue))
-			})
-
 		})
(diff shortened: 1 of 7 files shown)


mkuratczyk (approved):
As discussed - applying the labels initially but not reconciling them later feels wrong. We might want to change that in the future. But for now, at least we don't cause cluster restarts when adding labels.

controllers/rabbitmqcluster_controller_test.go
```diff
@@ -534,34 +534,13 @@ var _ = Describe("RabbitmqClusterController", func() {
...
-			})
-
-			It("updates annotations for config maps", func() {
-				Eventually(func() map[string]string {
-					roleBinding, err := clientSet.CoreV1().ConfigMaps(cluster.Namespace).Get(ctx, cluster.ChildResourceName("server-conf"), metav1.GetOptions{})
-					Expect(err).NotTo(HaveOccurred())
-					return roleBinding.Annotations
-				}, 3).Should(HaveKeyWithValue(annotationKey, annotationValue))
-
+			It("updates annotations for plugins config map", func() {
```
ansd, line 537: I'm wondering whether we should not update plugins config map either since a label gets added to that config map and `rabbitmq plugins-set` get executed on every pod of the StatefulSet for every Flux reconciliation.

internal/resource/configmap.go
```diff
@@ -55,16 +55,16 @@ func (builder *RabbitmqResourceBuilder) ServerConfigMap() *ServerConfigMapBuilde
 func (builder *ServerConfigMapBuilder) Build() (runtime.Object, error) {
 	return &corev1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      builder.Instance.ChildResourceName(ServerConfigMapName),
-			Namespace: builder.Instance.Namespace,
+			Name:        builder.Instance.ChildResourceName(ServerConfigMapName),
+			Namespace:   builder.Instance.Namespace,
+			Labels:      metadata.GetLabels(builder.Instance.Name, builder.Instance.Labels),
+			Annotations: metadata.ReconcileAndFilterAnnotations(nil, builder.Instance.Annotations),
```
ansd, line 61: Nice solution!

internal/resource/statefulset_test.go
```diff
@@ -401,11 +401,21 @@ var _ = Describe("StatefulSet", func() {
...
 
-			It("has the labels from the instance on the pod", func() {
+			It("adds default labels to pods but does not populate labels from the instance onto pods", func() {
 				stsBuilder := builder.StatefulSet()
 				Expect(stsBuilder.Update(statefulSet)).To(Succeed())
-				podTemplate := statefulSet.Spec.Template
-				testLabels(podTemplate.Labels)
+
+				labels := statefulSet.Spec.Template.ObjectMeta.Labels
+				Expect(len(labels)).To(Equal(3))
```
ansd, line 409: Could use `HaveLen` matcher

internal/resource/statefulset_test.go
```diff
@@ -523,28 +521,24 @@ var _ = Describe("StatefulSet", func() {
...
-						"prometheus.io/port":             "15692",
-						"my-annotation":                  "i-like-this",
-						"app.kubernetes.io/part-of":      "rabbitmq-pod",
-						"this-was-the-previous-pod-anno": "should-be-preserved",
-						"app.k8s.io/something":           "something-amazing-on-pod",
-					}
-
-					Expect(statefulSet.Spec.Template.Annotations).To(Equal(expectedAnnotations))
+					annotations := statefulSet.Spec.Template.Annotations
+					Expect(len(annotations)).To(Equal(3))
```
ansd, line 532: same here

controllers/rabbitmqcluster_controller_test.go
ChunyiLyu: I think you are right. I am really not a fan of having special behaviors for the configmaps, but I updated it. Nice catch David!

ansd (approved):

Issue #267, opened 2020-08-21
Share TLS config between RabbitMQ plugins by default


ferozjilla:
**Is your feature request related to a problem? Please describe.**

TLS for AMQP can be enabled by following -  https://www.rabbitmq.com/kubernetes/operator/using-operator.html#tls-conf.

TLS for other RabbitMQ plugins - the management plugin for example, must be done through a few CR changes. For example, by configuring the TLS properties for the plugin through `additionalConfig` and opening the TLS port through the `statefulSetOverride`.

The extra work for enabling TLS for plugins may not be justified given that in **most cases**:
1. When a user enables TLS, they wish to enable TLS for all plugins and not just AMQP
1. The same TLS credentials are used across plugins anyway

This *most cases* belief came up in our public slack channel and was confirmed by a member of the core team, and a platform architect. 

Given this, we can extend the AMQP TLS setting to all plugins. 

```gherkin
Scenario: TLS is configured for all enabled plugins
Given I deploy a RabbitMQ cluster with `spec.tls` set, and the plugins - management and stomp
When I try with contact RabbitMQ over the default TLS ports for each of the plugins - amqp, management, stomp
Then the connection succeeds and uses TLS

Scenario: Adding a plugin when TLS is enabled
Given I have a RabbitMQ cluster with TLS enabled
And I update the cluster by adding the MQTT plugin in the `additionalPlugins` section
When I try to connect to RabbitMQ using MQTTS over the default TLS port for MQTT
Then the connection succeeds and uses TLS

Scenario: Overriding the default TLS settings
Given I have a RabbitMQ cluster with TLS enabled and the management plugin
And I update the cluster to change the TLS port for management to 15678
By adding `management.ssl.port = 15678` to the `additionalConfig`
By opening port 15678 using the `statefulSetOverride`
When I try to connect over the new port
Then the connection succeeds and uses TLS

Scenario: Disable non-ssl listeners
Given I have a RabbitMQ cluster with TLS enabled and the management plugin
When I try to connect to the non-TLS management port (15672)
Then I see that the connection fails
```

**Additional Context**

RabbitMQ ships with some [tier-1 plugins](https://www.rabbitmq.com/plugins.html#tier1-plugins). We will need to decide which of these plugins to auto-enable TLS for. For example, enabling TLS for the management plugin involves much the same work as enabling TLS for AMQP. However, for the shovel and federation plugins, since another cluster is involved, this may be trickier. 

[...]

PR #451, opened 2020-11-09
Share TLS config with management, amqp 1.0, (web) mqtt and (web) stomp plugins

ferozjilla:
Relates to https://github.com/rabbitmq/cluster-operator/issues/267

## Summary Of Changes

Add TLS support for 
- management
- amqp 1.0
- mqtt and
- stomp plugins 

when TLS support is specified in the custom resource.

This is done by
- opening tls ports for each plugin on the RabbitMQ container
- exposing tls ports for each plugin on the service (formerly the client service)
- extending the `rabbitmq.conf` configmap to enable TLS for each plugin

Also included in these changes:
- renaming `management` to `http`

## Additional Context

The list of plugins considered to share the TLS config have been documented [in the issue](https://github.com/rabbitmq/cluster-operator/issues/267#issuecomment-723028006).

## Testing

Added a system test that checks connectivity of the http API over TLS.


ferozjilla:
TODO on this PR:
- add system tests for tls over mqtt and stomp

internal/resource/configmap_test.go
```diff
@@ -252,13 +296,66 @@ listeners.ssl.default = 5671
...
 ssl_options.verify     = verify_peer
+management.ssl.cacertfile = /etc/rabbitmq-tls/ca.certificate
 `)
 
 				Expect(configMapBuilder.Update(configMap)).To(Succeed())
 				Expect(configMap.Data).To(HaveKeyWithValue("rabbitmq.conf", expectedRabbitmqConf))
 			})
+
+			When("additional plugins are enabled", func() {
+				It("adds TLS config for the additional plugins", func() {
```
ChunyiLyu, line 314: We can have the description a bit clearer that this test is about testing the tls config for `rabbitmq_web_mqtt` and `rabbitmq_web_stomp`. Right now the spec description reads the same as the one from line `239`.

ablease: Ah good spot. I agree we can be more clear here, thanks Chunyi!

internal/resource/configmap_test.go
```diff
@@ -252,13 +296,66 @@ listeners.ssl.default = 5671
...
+management.ssl.port       = 15671
+
+mqtt.listeners.ssl.default = 8883
+
+stomp.listeners.ssl.1 = 61614
+
+ssl_options.cacertfile = /etc/rabbitmq-tls/ca.certificate
+ssl_options.verify     = verify_peer
+management.ssl.cacertfile = /etc/rabbitmq-tls/ca.certificate
+`)
```
Zerpet, line 353: How are we testing Web STOMP and Web MQTT here?

Following [RabbitMQ docs](https://www.rabbitmq.com/web-stomp.html#tls), I was expecting to see the keys `web_stomp.ssl.certfile` and `web_stomp.ssl.keyfile`. Same story for [MQTT](https://www.rabbitmq.com/web-mqtt.html#tls).

internal/resource/configmap_test.go
```diff
@@ -226,11 +226,55 @@ CONSOLE_LOG=new`
...
+							},
+							Rabbitmq: rabbitmqv1beta1.RabbitmqClusterConfigurationSpec{
+								AdditionalPlugins: []rabbitmqv1beta1.Plugin{
+									"rabbitmq_mqtt",
+									"rabbitmq_stomp",
+									"rabbitmq_amqp_1_0",
+								},
+							},
+						},
+					}
```
Zerpet, line 257: Neat pick: perhaps we can use `generateRabbitmqCluster()` and modify only the bits an pieces we want to test?

https://github.com/rabbitmq/cluster-operator/blob/911ef0e91ce9028599ee6539529af09397613ae9/internal/resource/statefulset_test.go#L1572-L1579

Something in the lines of:

```golang
instance = generateRabbitmqCluster()
instance.spec.ObjectMeta.Name = "rabbit-tls"
instance.spec.TLS.SecretName = "a-secret"
instance.spec.Rabbitmq.AdditionalPlugins = []rabbitmqv1beta1.Plugin{ ... }
```

Zerpet (changes requested):
The test for secure connections over web sockets doesn't look right to me. Let's double check how to configure those.

internal/resource/configmap_test.go
ferozjilla: Web mqtt and web stomp are not implemented yet. The decision to add them too was reached at sync-up and is one of the actions (along with removing the tls property) that we need to do.

internal/resource/configmap_test.go
ferozjilla: That's a good idea!

ChunyiLyu:
What's the particular reason to rename port 'management' to 'http'? I think @mkuratczyk  has just recently PR the rename from 'http' to 'management' in the statefulSet: https://github.com/rabbitmq/cluster-operator/pull/444

internal/resource/configmap_test.go
ablease: Hi @Zerpet. configmap additions for web_mqtt and web_stomp have been added here https://github.com/rabbitmq/cluster-operator/pull/451/commits/bc28103a2ba49af6a2d62e6facc1cf054203357c. Please take a look and let me know if it looks right.

internal/resource/configmap_test.go
ablease: I like this idea too. In the interest of time I will create another issue to refactor the config map tests. Thanks Aitor

ablease:
> #444

Ah well spotted. I will opt to rename ports from http to management. What would be a good name for 'https' ports? 'management-tls' perhaps?

ablease:
> What's the particular reason to rename port 'management' to 'http'? I think @mkuratczyk has just recently PR the rename from 'http' to 'management' in the statefulSet: #444

Ah well spotted. I will opt to rename ports from http to management. What would be a good name for 'https' ports? 'management-tls' perhaps?

ChunyiLyu:
> Ah well spotted. I will opt to rename ports from http to management. What would be a good name for 'https' ports? 'management-tls' perhaps?

Sounds good. I see that we are naming tls port as "web-mqtt-tls" and "web-stomp-tls", so lets keep it in the same format.

internal/resource/statefulset.go
```diff
@@ -436,12 +436,32 @@ func (builder *StatefulSetBuilder) podTemplateSpec(annotations, labels map[strin
 	}
 
 	tlsSpec := builder.Instance.Spec.TLS
-	if tlsSpec.SecretName != "" {
-		// add tls port
+	if builder.Instance.TLSEnabled() {
+		// add tls port, should we disable non tls ports here?
```
ChunyiLyu, line 440: Lets remove the comment here since we've already created an issue for this

internal/resource/service.go
```diff
@@ -184,6 +184,22 @@ func (builder *ServiceBuilder) updatePorts(servicePorts []corev1.ServicePort) []
 			}
 		}
 	}
+	if builder.Instance.TLSEnabled() && builder.Instance.MutualTLSEnabled() {
```
Zerpet, line 187: The function `MutualTLSEnabled()` already checks if `TLSEnabled()`. Don't we get the same result if we just check for `MutualTLSEnabled()` ?

Zerpet (approved):

internal/resource/statefulset.go
ablease: Makes sense, thanks for pointing that out! https://github.com/rabbitmq/cluster-operator/pull/451/commits/b737ca4a59b2e05ac2ba106efb485dbe6f603fcd

internal/resource/service.go
ablease: Ah so it does! Will just use the single check here instead. Thanks for pointing that out! Change made here https://github.com/rabbitmq/cluster-operator/pull/451/commits/b737ca4a59b2e05ac2ba106efb485dbe6f603fcd.

internal/resource/configmap.go
```diff
@@ -81,15 +86,57 @@ func (builder *ServerConfigMapBuilder) Update(object runtime.Object) error {
...
+			}
+			if _, err := defaultSection.NewKey("web_mqtt.ssl.keyfile", "/etc/rabbitmq-tls/tls.key"); err != nil {
+				return err
+			}
+		}
+		if builder.Instance.AdditionalPluginEnabled("rabbitmq_web_stomp") {
+			if _, err := defaultSection.NewKey("web_stomp.ssl.port", "15673"); err != nil {
+				return err
+			}
+			if _, err := defaultSection.NewKey("web_stomp.ssl.cacertfile", "/etc/rabbitmq-tls/ca.crt"); err != nil {
```
ansd, line 130: Should we extract some of these paths into variables?
`/etc/rabbitmq-tls/ca.crt` is used 4 times here.

system_tests/utils.go
```diff
@@ -741,6 +764,13 @@ func publishAndConsumeMQTTMsg(hostname, nodePort, username, password string, ove
 		SetClientID("system tests MQTT plugin").
 		SetProtocolVersion(4) // RabbitMQ MQTT plugin targets MQTT 3.1.1
 
+	if tlsConfig != nil {
+		url = fmt.Sprintf("ssl://%s:%s", hostname, nodePort)
+		opts = opts.
+			AddBroker(url).
+			SetTLSConfig(tlsConfig)
+	}
+
```
ansd, line 773: Can we also add a test for WebMQTT and TLS, i.e. testing `wss`?

ansd (changes requested):

internal/resource/service.go
```diff
@@ -163,6 +163,42 @@ func (builder *ServiceBuilder) updatePorts(servicePorts []corev1.ServicePort) []
...
+			servicePortsMap["stomps"] = corev1.ServicePort{
+				Protocol: corev1.ProtocolTCP,
+				Port:     15673,
+				Name:     "web-stomp-tls",
+			}
+		}
+		if builder.Instance.AdditionalPluginEnabled("rabbitmq_web_mqtt") {
+			servicePortsMap["mqtts"] = corev1.ServicePort{
+				Protocol: corev1.ProtocolTCP,
+				Port:     15676,
```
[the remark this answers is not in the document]
ChunyiLyu: updated. Good catch david!

controllers/reconcile_tls.go
```diff
@@ -47,12 +48,12 @@ func (r *RabbitmqClusterReconciler) checkTLSSecrets(ctx context.Context, rabbitm
...
 		// Mutual TLS: verify that CA certificate is present in secret
-		_, hasCaCert := secret.Data[rabbitmqCluster.Spec.TLS.CaCertName]
+		_, hasCaCert := secret.Data["ca.crt"]
 		if !hasCaCert {
 			r.Recorder.Event(rabbitmqCluster, corev1.EventTypeWarning, "TLSError",
-				fmt.Sprintf("The TLS secret %v in namespace %v must have the field %v", rabbitmqCluster.Spec.TLS.CaSecretName, rabbitmqCluster.Namespace, rabbitmqCluster.Spec.TLS.CaCertName))
+				fmt.Sprintf("The TLS secret %v in namespace %v must have the field ca.crt", rabbitmqCluster.Spec.TLS.CaSecretName, rabbitmqCluster.Namespace))
 
-			return ctrl.Result{}, errors.NewBadRequest(fmt.Sprintf("The TLS secret must have the field %s", rabbitmqCluster.Spec.TLS.CaCertName))
+			return ctrl.Result{}, errors.NewBadRequest("The TLS secret must have the field ca.crt")
```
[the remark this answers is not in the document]
ChunyiLyu: The error should already contain the secret name and namespace it tries to get but failed.

internal/resource/configmap.go
ChunyiLyu: updated. Good suggestion!

controllers/reconcile_tls.go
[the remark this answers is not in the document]
ansd: Where is that happening?
The Event contains secret name and namespace, but the error which is returned doesn't?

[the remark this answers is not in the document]
ChunyiLyu: @ansd updated. Could you take a look again?

system_tests/utils.go
ChunyiLyu: I don't see too much value in us testing every single plugins protocol with TLS. I think the coverage we have currently is enough, and would prefer to leave the system tests as it is at least for now.

controllers/reconcile_tls.go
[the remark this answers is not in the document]
ansd: looks good thanks

system_tests/utils.go
ansd: The additional value is that we configured MQTT Web Stomp TLS correctly, i.e. the changes in https://github.com/rabbitmq/cluster-operator/pull/451/files#diff-8963e26fa961496c5fbf37e06487c020529c5cee6f6ff953ed8d63c018d1c5efR115-R128 and that we expose the port. We do have unit tests, but an end-to-end test adds value from my point of view that this feature actually works.

However, as discussed in standup, let's not block merging the PR on this test. @ablease said he will add a test once this is merged.

ansd (approved):

PR #462, opened 2020-11-10
Add `kubectl rabbitmq tail` subcommand

mkuratczyk:
New command to tail logs from all nodes.

### DIFF
--- a/bin/kubectl-rabbitmq
+++ b/bin/kubectl-rabbitmq
@@ -43,7 +43,7 @@ USAGE:
   Set log level to 'debug' on all nodes
     kubectl rabbitmq debug INSTANCE
 
-  Tail logs from all all nodes
+  Tail logs from all nodes
     kubectl rabbitmq tail INSTANCE
 
   Run 'rabbitmq-diagnostics observer' on a specific INSTANCE NODE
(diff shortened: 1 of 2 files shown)


ansd (commented):
Can we achieve the same without the dependency to the external `tail` plugin by running:
```
kubectl logs --follow=true --prefix=true -lapp.kubernetes.io/name=<INSTANCE> -lapp.kubernetes.io/part-of=rabbitmq -lapp.kubernetes.io/component=rabbitmq
```
This would have the advantage that users wouldn't need to install the `tail` plugin.

Furthermore, we wouldn't rely on external plugins which aren't maintained very well. The krew index points to v0.10.1 (see https://github.com/kubernetes-sigs/krew-index/blob/b093153e8476f4dabc6515593f854bd0fa45b6d5/plugins/tail.yaml#L6) but the latest plugin version is v0.15.0.

mkuratczyk:
The command you suggested disconnects when a container disappears. For monitoring a cluster restart/upgrade, the plugin is better, unless there are some additional `kubectl` flags that would improve the behaviour.

ansd (approved):

MirahImage (approved):

PR #477, opened 2020-11-18
Add spec.tls.disableNonTLSListeners property

### LINKED ISSUE #459: Close non-TLS service and container ports when `listeners.tcp = none` is set in the `rabbitmq.conf` configmap
Users may [disable non-TLS listeners](https://www.rabbitmq.com/networking.html#single-stack-ipv4) by adding the `listeners.tcp = none` property to the [additional config](https://www.rabbitmq.com/kubernetes/operator/using-operator.html#additional-config). This will result in the property being added to `rabbitmq.conf` which will result RabbitMQ not listening over non-TLS ports . The service and RabbitMQ container however will still expose the ports. This may be confusing for users and they may wonder if their intended change - disable non-TLS for RabbitMQ, has actually worked.

**Implementation Options**
1. The `rabbitmq.conf` configmap can detect when the property (`listeners.tcp = none`) is being set. If being set, we should remove all non-TLS ports from the service and the statefulset pod spec. Our resources do not have a way to affect each other at the moment, so we will have to figure out how to do this.

2. An alternative would be to add a top level property to the custom resource (something like `disableNonTLSListeners: true`). This would make it easy to identify when to change all affected objects (`rabbitmq.conf` configmap, service ports, container ports).

ChunyiLyu:
This closes #459 

## Summary Of Changes

- Add `spec.tls.disableNonTLSListeners` property, when set to true, the operator disables non TLS listeners in given RabbitmqCluster
- ` spec.tls.disableNonTLSListeners` also means tcp readiness probe would use port `amqps` instead of `amqp`
- The property can only configure to true if TLS is enabled, or else there will be errors logged and published to events
- when ` spec.tls.disableNonTLSListeners` set to true, the epmd 4369 and the prometheus 15692 ports remains open on the container level for clustering and for prometheus scraping (neither of them are exposed in the outfacing service object)
- tcp listeners disabled for any enabled plugins as well by setting plugins specific tcp config:

```
mqtt.listeners.tcp   = none
stomp.listeners.tcp   = none
rabbitmq_web_mqtt.tcp_config = none
rabbitmq_web_stomp.tcp_config = none
```

- extracted func `reconcileTLS()`

## Additional Context

Some additional refactoring in system tests when asserting content of `rabbitmq.conf` using `kubectl` and ini library.

### DIFF
--- a/controllers/reconcile_tls.go
+++ b/controllers/reconcile_tls.go
@@ -8,10 +8,25 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
-	ctrl "sigs.k8s.io/controller-runtime"
 )
 
-func (r *RabbitmqClusterReconciler) checkTLSSecrets(ctx context.Context, rabbitmqCluster *rabbitmqv1beta1.RabbitmqCluster) (ctrl.Result, error) {
+func (r *RabbitmqClusterReconciler) reconcileTLS(ctx context.Context, rabbitmqCluster *rabbitmqv1beta1.RabbitmqCluster) error {
+	if rabbitmqCluster.DisableNonTLSListeners() && !rabbitmqCluster.TLSEnabled() {
+		err := errors.NewBadRequest("TLS must be enabled if disableNonTLSListeners is set to true")
+		r.Recorder.Event(rabbitmqCluster, corev1.EventTypeWarning, "TLSError", err.Error())
+		r.Log.Error(err, "Error setting up TLS", "namespace", rabbitmqCluster.Namespace, "name", rabbitmqCluster.Name)
+		return err
+	}
+
+	if rabbitmqCluster.TLSEnabled() {
+		if err := r.checkTLSSecrets(ctx, rabbitmqCluster); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (r *RabbitmqClusterReconciler) checkTLSSecrets(ctx context.Context, rabbitmqCluster *rabbitmqv1beta1.RabbitmqCluster) error {
 	secretName := rabbitmqCluster.Spec.TLS.SecretName
 	r.Log.Info("TLS enabled, looking for secret", "secret", secretName, "namespace", rabbitmqCluster.Namespace)
 
@@ -21,7 +36,7 @@ func (r *RabbitmqClusterReconciler) checkTLSSecrets(ctx context.Context, rabbitm
 		r.Recorder.Event(rabbitmqCluster, corev1.EventTypeWarning, "TLSError",
 			fmt.Sprintf("Failed to get TLS secret %s in namespace %s: %v", secretName, rabbitmqCluster.Namespace, err.Error()))
 		r.Log.Error(err, "Error setting up TLS", "namespace", rabbitmqCluster.Namespace, "name", rabbitmqCluster.Name)
-		return ctrl.Result{}, err
+		return err
 	}
 	// check if secret has the right keys
 	_, hasTLSKey := secret.Data["tls.key"]
@@ -30,7 +45,7 @@ func (r *RabbitmqClusterReconciler) checkTLSSecrets(ctx context.Context, rabbitm
 		err := errors.NewBadRequest(fmt.Sprintf("TLS secret %s in namespace %s does not have the fields tls.crt and tls.key", secretName, rabbitmqCluster.Namespace))
 		r.Recorder.Event(rabbitmqCluster, corev1.EventTypeWarning, "TLSError", err.Error())
 		r.Log.Error(err, "Error setting up TLS", "namespace", rabbitmqCluster.Namespace, "name", rabbitmqCluster.Name)
-		return ctrl.Result{}, err
+		return err
 	}
 
 	// Mutual TLS: check if CA certificate is stored in a separate secret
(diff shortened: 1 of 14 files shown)


internal/resource/configmap.go
```diff
@@ -99,6 +99,11 @@ func (builder *ServerConfigMapBuilder) Update(object runtime.Object) error {
 				return err
 			}
 		}
+		if builder.Instance.Spec.TLS.DisableNonTLSListeners {
+			if _, err := defaultSection.NewKey("listeners.tcp", "none"); err != nil {
+				return err
+			}
+		}
```
ansd, line 106: As described in https://github.com/rabbitmq/cluster-operator/issues/459#issuecomment-725378817, we also need further settings if Web MQTT or Web STMOP is enabled and `DisableNonTLSListeners` is set.

ansd (changes requested):

internal/resource/configmap.go
ChunyiLyu: silly me indeed. Mark the PR as draft for now

ChunyiLyu: added settings for stomp, mqtt, web_stomp and web_mqtt. For management, the general `listeners.tcp = none` had the effect. Thanks for bring it up David.

internal/resource/service.go
```diff
@@ -109,19 +109,77 @@ func applySvcOverride(svc *corev1.Service, override *rabbitmqv1beta1.Service) er
...
+			Protocol:   corev1.ProtocolTCP,
+			Port:       8883,
+			Name:       "mqtts",
+			TargetPort: intstr.FromInt(8883),
+		}
+	}
+
+	if builder.Instance.MutualTLSEnabled() {
+		if builder.Instance.AdditionalPluginEnabled("rabbitmq_web_stomp") {
+			servicePortsMap["stomps"] = corev1.ServicePort{
```
ansd, line 147: Doesn't this override the STOMP TLS port defined a few lines above?

internal/resource/service.go
```diff
@@ -109,19 +109,77 @@ func applySvcOverride(svc *corev1.Service, override *rabbitmqv1beta1.Service) er
...
+				Protocol:   corev1.ProtocolTCP,
+				Port:       15673,
+				Name:       "web-stomp-tls",
+				TargetPort: intstr.FromInt(15673),
+			}
+		}
+		if builder.Instance.AdditionalPluginEnabled("rabbitmq_web_mqtt") {
+			servicePortsMap["mqtts"] = corev1.ServicePort{
```
ansd, line 155: Doesn't this override the MQTT TLS port defined a few lines above?

ansd (changes requested):

mkuratczyk:
I don't remember what we currently use `.status.reason=NoWarnings` for but it feels like the situation when `disableNonTLSListeners=true` but no TLS secret was provide should be one of them. The event is nice but seeing `NoWarnings=True` suggests that there is no reason to look at events. :)

internal/resource/configmap.go
```diff
@@ -139,6 +159,11 @@ func (builder *ServerConfigMapBuilder) Update(object runtime.Object) error {
 			if _, err := defaultSection.NewKey("web_stomp.ssl.keyfile", tlsKeyPath); err != nil {
 				return err
 			}
+			if builder.Instance.DisableNonTLSListeners() {
+				if _, err := defaultSection.NewKey("rabbitmq_web_stomp.tcp_config", "none"); err != nil {
```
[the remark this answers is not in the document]
ChunyiLyu: I took your word and didn't check the doc for more details :( Updated!

internal/resource/service.go
ChunyiLyu: Updated, thank you for catching it!

controllers/reconcile_tls.go
```diff
@@ -11,6 +11,22 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 )
 
+func (r *RabbitmqClusterReconciler) reconcileTLS(ctx context.Context, rabbitmqCluster *rabbitmqv1beta1.RabbitmqCluster) (ctrl.Result, error) {
```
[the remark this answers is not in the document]
ChunyiLyu: Updated, good suggestion!

mkuratczyk (approved):
As discussed, `NoWarnings` can be implemented separately as there are more conditions in which we should change it.

system_tests/system_tests.go
```diff
@@ -415,6 +407,22 @@ CONSOLE_LOG=new`
 					Expect(err).NotTo(HaveOccurred())
 					publishAndConsumeSTOMPMsg(hostname, rabbitmqNodePort(ctx, clientSet, cluster, "stomps"), username, password, cfg)
 				})
+
+				By("disabling non TLS listeners", func() {
+					// verify that rabbitmq.conf contains listeners.tcp = none
+					cfgMap := getConfigFileFromPod(namespace, cluster, "/etc/rabbitmq/rabbitmq.conf")
+					Expect(cfgMap).To(HaveKeyWithValue("listeners.tcp", "none"))
```
ansd, line 414: Let's add two more expectations here because that system test already enables MQTT, STOMP, TLS, and the flag `DisableNonTLSListeners`?
```
Expect(cfgMap).To(HaveKeyWithValue("mqtt.listeners.tcp", "none"))
Expect(cfgMap).To(HaveKeyWithValue("stomp.listeners.tcp", "none"))
```

ansd (approved):
Thanks @ChunyiLyu, looks very good!

I left one comment, up to you whether you want to include the two expectations.

system_tests/system_tests.go
ChunyiLyu: Good point David. Updated and will merge now 👍

PR #478, opened 2020-11-19
Reuse cluster generator in configmap tests

### LINKED ISSUE #464: Refactor configmap tests with common RabbitmqCluster and then modify properties as we need.
Similar to the stateful set tests were we have a common RabbitmqCluster generator. We could do something like this in the configmap tests too.

Neat pick: perhaps we can use `generateRabbitmqCluster()` and modify only the bits an pieces we want to test?

https://github.com/rabbitmq/cluster-operator/blob/911ef0e91ce9028599ee6539529af09397613ae9/internal/resource/statefulset_test.go#L1572-L1579

Something in the lines of:

```golang
instance = generateRabbitmqCluster()
instance.spec.ObjectMeta.Name = "rabbit-tls"
instance.spec.TLS.SecretName = "a-secret"
instance.spec.Rabbitmq.AdditionalPlugins = []rabbitmqv1beta1.Plugin{ ... }
```

_Originally posted by @Zerpet in https://github.com/rabbitmq/cluster-operator/pull/451#discussion_r521287325_

ablease:
This closes #464

## Summary Of Changes

As suggested by @Zerpet in #464 this commit reuses the cluster generator, to clean up the config map tests. 

## Local Testing

### DIFF
--- a/internal/resource/configmap_test.go
+++ b/internal/resource/configmap_test.go
@@ -21,7 +21,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	k8sresource "k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	defaultscheme "k8s.io/client-go/kubernetes/scheme"
 )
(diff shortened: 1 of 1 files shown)


internal/resource/configmap_test.go
```diff
@@ -235,7 +222,7 @@ CONSOLE_LOG=new`
 				Expect(configMap.Data).ToNot(HaveKey("rabbitmq-env.conf"))
 			})
 
-			Context("rabbitmq-env.conf is set", func() {
+			Context("rabbitmq-env.conf i set", func() {
```
MirahImage, line 225: You probably didn't mean to change this line

MirahImage (changes requested):
Other than the accidental change on one line, looks good.

internal/resource/configmap_test.go
ablease: Oh yes, thanks for spotting that.

MirahImage (approved):

Zerpet (approved):

PR #481, opened 2020-11-20
Make rabbitmq image configurable

ansd:
Make rabbitmq image configurable by setting environment variables so that we can run system tests using the commercial rabbitmq image as well as using the OSS rabbitmq image.

Next step is to add a Concourse job to run system tests with the commercial image.


MirahImage (approved):

ChunyiLyu (approved):
Awesome!

PR #486, opened 2020-11-23
Update service name in the binding

mkuratczyk:
We've changed the name of the service in https://github.com/rabbitmq/cluster-operator/issues/415


ChunyiLyu (approved):
Good catch Michal!

PR #487, opened 2020-11-23
Stop disabling management tcp listener when tls enabled

### LINKED ISSUE #482: Management TCP listener should still work when TLS is enabled
## Describe the bug

Enabling TLS disables management TCP listener. Management TCP listener should remain working unless `disableNonTLSListeners` is set to true.

## To Reproduce
1. Enable TLS
2. See that you cannot connect management through port `15672` anymore

Include any YAML or manifest necessary to reproduce the problem.

**Expected behavior**

I expect both TLS port `15671` and non TLS port `15672` to work, since I didn't set the `disableNonTLSListeners` to be true

## Version information
Latest main.

ChunyiLyu:
This closes #482 

## Summary Of Changes

- fix a bug where management tcp listener no longer works when tls is enabled and disableNonTLSListeners is set to false
- management tcp listener can be disabled by setting management.ssl.port without setting management.tcp.port
- management.tcp.port is set to 15672 when tls is enabled but disableNonTLSListeners is false
- do not set management.tcp.port when tls is enabled and disableNonTLSListeners is true
- increate timeout in one integration test case( it was flaking due to the short timeout)

I have thought about including `management.tcp.port = 15672` as the default configuration and deleting the key if `disableNonTLSListeners` is set to true. However, that will unfortunately change all clusters's rabbitmq.conf (no matter its tls configurations), that would be a breaking behavior. Here, `management.tcp.port = 15672` is only set if people enable tls and do not set `disableNonTLSListeners` to true.

## Local Testing

Added unit and system tests for the case.

### DIFF
--- a/internal/resource/configmap.go
+++ b/internal/resource/configmap.go
@@ -93,6 +93,13 @@ func (builder *ServerConfigMapBuilder) Update(object runtime.Object) error {
 			if _, err := defaultSection.NewKey("listeners.tcp", "none"); err != nil {
 				return err
 			}
+		} else {
+			// management plugin does not have a *.listeners.tcp settings like other plugins
+			// management tcp listener can be disabled by setting management.ssl.port without setting management.tcp.port
+			// we set management tcp listener only if tls is enabled and disableNonTLSListeners is false
+			if _, err := defaultSection.NewKey("management.tcp.port", "15672"); err != nil {
+				return err
+			}
 		}
 		if builder.Instance.AdditionalPluginEnabled("rabbitmq_mqtt") {
 			if _, err := defaultSection.NewKey("mqtt.listeners.ssl.default", "8883"); err != nil {
--- a/system_tests/system_tests.go
+++ b/system_tests/system_tests.go
@@ -409,9 +409,12 @@ CONSOLE_LOG=new`
 				By("disabling non TLS listeners", func() {
 					// verify that rabbitmq.conf contains listeners.tcp = none
 					cfgMap := getConfigFileFromPod(namespace, cluster, "/etc/rabbitmq/rabbitmq.conf")
-					Expect(cfgMap).To(HaveKeyWithValue("listeners.tcp", "none"))
-					Expect(cfgMap).To(HaveKeyWithValue("stomp.listeners.tcp", "none"))
-					Expect(cfgMap).To(HaveKeyWithValue("mqtt.listeners.tcp", "none"))
+					Expect(cfgMap).To(SatisfyAll(
+						HaveKeyWithValue("listeners.tcp", "none"),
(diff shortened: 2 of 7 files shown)


system_tests/system_tests.go
```diff
@@ -409,9 +409,12 @@ CONSOLE_LOG=new`
...
 					cfgMap := getConfigFileFromPod(namespace, cluster, "/etc/rabbitmq/rabbitmq.conf")
-					Expect(cfgMap).To(HaveKeyWithValue("listeners.tcp", "none"))
-					Expect(cfgMap).To(HaveKeyWithValue("stomp.listeners.tcp", "none"))
-					Expect(cfgMap).To(HaveKeyWithValue("mqtt.listeners.tcp", "none"))
+					Expect(cfgMap).To(SatisfyAll(
+						HaveKeyWithValue("listeners.tcp", "none"),
+						HaveKeyWithValue("stomp.listeners.tcp", "none"),
+						HaveKeyWithValue("mqtt.listeners.tcp", "none"),
+						HaveKeyWithValue("management.ssl.port", "15671"),
+						Not(HaveKey("management.tcp.port"))))
```
ansd, line 417: nice!

internal/resource/configmap.go
```diff
@@ -93,6 +93,13 @@ func (builder *ServerConfigMapBuilder) Update(object runtime.Object) error {
 			if _, err := defaultSection.NewKey("listeners.tcp", "none"); err != nil {
 				return err
 			}
+		} else {
+			// management plugin does not have a *.listeners.tcp settings like other plugins
+			// management tcp listener can be disable by setting management.ssl.port without setting management.tcp.port
```
ansd, line 98: spelling: "can be disabled"

ansd (approved):
Not directly related to your PR, but I'm just noticing we should also update https://github.com/rabbitmq/cluster-operator/blob/b2061cce36ab4b820fed4d2d96169c5a955a368a/bin/kubectl-rabbitmq#L130-L132 and use port 15671 if TLS is enabled.
Do you want to fix it or create an issue with label `bug`?

internal/resource/configmap.go
ChunyiLyu: Thanks!

ChunyiLyu:
@ansd maybe a separate PR? When tls is enabled, port `15672` would still work unless people configure the `disableNonTLSListeners` property to true, so I don't think this is urgent. Similarly, we should also look into the [perf-test command](https://github.com/rabbitmq/cluster-operator/blob/b2061cce36ab4b820fed4d2d96169c5a955a368a/bin/kubectl-rabbitmq#L111). Would the perf test fail to connect if `disableNonTLSListeners` is true?

mkuratczyk (approved):

PR #497, opened 2020-11-24
Test all examples in PR github action

### LINKED ISSUE #496: Test all docs/examples examples in PR github action flow

As discussed in standup and sync-up, we should test (kubectl apply with dry run flag) all example manifests in `docs/examples` in PRs to avoid having out-of-date examples documented.

 n/a

ChunyiLyu:
This closes #496 

## Summary Of Changes

As discussed in standup and sync-up, we should test (kubectl apply with dry run flag) all example manifests in `docs/examples` in PRs to avoid having out-of-date examples documented.

## Local Testing

Tested the github action flow locally using [act](https://github.com/nektos/act) and this tool is great and super convenient!

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -21,6 +21,23 @@ jobs:
       working-directory: charts/rabbitmq
       run: ./test.sh
 
+  test-all-examples:
+    name: test-all-examples
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+    - name: System tests
+      run: |
+        curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"  | bash
+        export GOPATH=$HOME/go
+        export PATH=$PATH:$GOPATH/bin
+        make install-tools
+        kind create cluster --image kindest/node:v1.19.1
+        kubectl apply --validate=false -f https://github.com/jetstack/cert-manager/releases/latest/download/cert-manager.crds.yaml
+        make install
+        kubectl apply --dry-run=server --recursive -f docs/examples/
+
   system_tests:
     name: system tests
     runs-on: ubuntu-latest


MirahImage (approved):

.github/workflows/pr.yml
```diff
@@ -21,6 +21,23 @@ jobs:
...
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+    - name: System tests
+      run: |
+        curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"  | bash
+        export GOPATH=$HOME/go
+        export PATH=$PATH:$GOPATH/bin
+        make install-tools
+        kind create cluster --image kindest/node:v1.19.1
+        kubectl apply --validate=false -f https://github.com/jetstack/cert-manager/releases/download/v1.0.4/cert-manager.crds.yaml
```
ansd, line 37: Does it make sense to use `https://github.com/jetstack/cert-manager/releases/latest/download/cert-manager.crds.yaml` such that we don't have to bump manually?

.github/workflows/pr.yml
```diff
@@ -21,6 +21,23 @@ jobs:
...
+    steps:
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+    - name: System tests
+      run: |
+        curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"  | bash
+        export GOPATH=$HOME/go
+        export PATH=$PATH:$GOPATH/bin
+        make install-tools
+        kind create cluster --image kindest/node:v1.19.1
```
coro, line 36: I could see an argument for doing this on all supported k8s versions, like the system tests do, in parallel. This might allow us to spot where a PR makes a specific use case of the operator no longer functional (though I could also see the argument that this would hopefully be caught in the system tests).

.github/workflows/pr.yml
ChunyiLyu: Sure I will do that.

.github/workflows/pr.yml
ChunyiLyu: The examples are tested as 'dry-runs' without testing specific functionality and only about api compatibility to the rabbitmqcluster CRD. The rabbitmqcluster CRD that's installed, whether to k8s `1.17` or `1.19`, is the same, so I don't see how testing against multiple k8s versions could increase test coverage.

ansd (approved):

PR #499, opened 2020-11-25
Lower staleness threshold to 60 days

coro:
This is an action from a team retro or sync up. The modal votes for the threshold were for 60 days to stale, 30 days to close.

PRs & issues can still be marked with the label `never-stale` to be exempt from this check.


MirahImage (approved):

ChunyiLyu (approved):
Nice!

PR #303, opened 2020-09-01
Support OpenShift

ansd:
Fixes #234 

### Changes:
1. Allow `rabbitmq-cluster-operator-role` to update `rabbitmqclusters/finalizers`
2. Do not `BlockOwnerDeletion` for PVCs (as done in https://github.com/elastic/cloud-on-k8s/pull/1891)
3. Change group owner of `/var/lib/rabbitmq/mnesia/` to `999`

### Docs:
https://github.com/ansd/rabbitmq-website/tree/k8s-openshift (need to create PR)

### Testing on OpenShift:
If you don't have an OpenShift cluster available, the easiest way to set one up within a few minutes is [CodeReady Containers](https://developers.redhat.com/products/codeready-containers/overview).

* Download CodeReady Containers
* `crc setup`
*  `crc config set memory 12000` (the default of 8192 MiB is too low to run the RabbitMQ operator and a RabbitMQ cluster instance since there isn't much memory left when starting crc)
* `crc start`
* `eval $(crc oc-env)`
* `oc login -u kubeadmin -p <password> https://api.crc.testing:6443`
* `make deploy-dev`
* Operator won’t deploy because
```
$ kubectl describe replicasets.apps
...
Events:
  Type     Reason        Age                From                   Message
  ----     ------        ----               ----                   -------
  Warning  FailedCreate  9s (x13 over 29s)  replicaset-controller  Error creating: pods "rabbitmq-cluster-rabbitmq-cluster-operator-6bb8fd7bf8-" is forbidden: unable to validate against any security context constraint: [fsGroup: Invalid value: []int64{1000}: 1000 is not an allowed group spec.containers[0].securityContext.securityContext.runAsUser: Invalid value: 1000: must be in the ranges: [1000610000, 1000619999]]
```
* As explained [here](https://access.redhat.com/solutions/2801791): `oc edit namespace rabbitmq-system` Change the `uid-range` and `supplemental-groups` for the operator:
```
apiVersion: v1
kind: Namespace
metadata:
  annotations:
...
    openshift.io/sa.scc.supplemental-groups: 1000/1
    openshift.io/sa.scc.uid-range: 1000/1
```
* Operator deploys successfully
* Change the `uid-range` and `supplemental-groups` for RabbitMQ (here we assume that the RabbitMQ cluster gets deployed into the `default` namespace):
`oc edit namespace default`
```
apiVersion: v1
kind: Namespace
metadata:
  annotations:
...
    openshift.io/sa.scc.supplemental-groups: 999/1
[...]

PR #507, opened 2020-11-30
Mounts and capabilities

ChunyiLyu:
This fixes issue: https://github.com/rabbitmq/tanzu-cluster-operator/issues/20

## Summary Of Changes

- always mount '/var/lib/rabbitmq/' before '/var/lib/rabbitmq/mnesia/' to avoid shadowing; popular open-sourced container runtimes like docker and containerD will sort mounts in alphabetical order to avoid this issue, but there's no guarantee that all container runtime would do so
- update initContainer capabilities to be able to deploy successfully in cluster with ESXi runtime
- Context on the capabilities: dropping all capabilities except CHOWN, DAC_OVERRIDE, and FOWNER. The first of these capabilities is necessary to change a file's owner, the second is necessary to permit chown to traverse directories to which root doesn't otherwise have access. FOWNER is required for chmod and chgrp

Strictly speaking DAC_OVERRIDE isn't necessary, only DAC_READ_SEARCH is, but as it's not included in the default set using it is slightly more complex. I don't think there's much practical benefit since by default the only process with any capabilities at all is the chown process (everything else runs as 472 and gets no capabilities at all).

## Additional Context

- to make sure that '/var/lib/rabbitmq/' always mounts before '/var/lib/rabbitmq/mnesia/', volumeMounts is handled as a special case if it's patched with override

## Local Testing

unit, integration, and system tests have all passed

### DIFF
--- a/internal/resource/statefulset.go
+++ b/internal/resource/statefulset.go
@@ -521,8 +521,20 @@ func (builder *StatefulSetBuilder) podTemplateSpec(previousPodAnnotations map[st
 					SecurityContext: &corev1.SecurityContext{
 						RunAsUser: pointer.Int64Ptr(0),
 						Capabilities: &corev1.Capabilities{
-							Drop: []corev1.Capability{"ALL"},
-							Add:  []corev1.Capability{"CHOWN", "FOWNER"},
+							// drop default set from Docker except for CHOWN, FOWNER, and DAC_OVERRIDE
+							Drop: []corev1.Capability{
+								"FSETID",
+								"KILL",
+								"SETGID",
+								"SETUID",
+								"SETPCAP",
+								"NET_BIND_SERVICE",
+								"NET_RAW",
+								"SYS_CHROOT",
+								"MKNOD",
+								"AUDIT_WRITE",
+								"SETFCAP",
+							},
 						},
 					},
 					Command: []string{
--- a/internal/resource/statefulset.go
+++ b/internal/resource/statefulset.go
@@ -229,10 +229,16 @@ func patchPodSpec(podSpec, podSpecOverride *corev1.PodSpec) (corev1.PodSpec, err
 		return corev1.PodSpec{}, fmt.Errorf("error unmarshalling patched Stateful Set: %v", err)
 	}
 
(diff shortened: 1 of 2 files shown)


mkuratczyk (approved):
Thanks!

internal/resource/statefulset.go
```diff
@@ -499,8 +521,19 @@ func (builder *StatefulSetBuilder) podTemplateSpec(previousPodAnnotations map[st
...
+								"SETGID",
+								"SETUID",
+								"SETPCAP",
+								"NET_BIND_SERVICE",
+								"NET_RAW",
+								"SYS_CHROOT",
+								"MKNOD",
+								"AUDIT_WRITE",
+								"SETFCAP",
+							},
```
ansd, line 536: @coro proposed in https://github.com/rabbitmq/cluster-operator/pull/303#discussion_r485026855 to drop all and only allow the needed capabilities.
Allow listing rather than deny listing is a better security standard.

IIRC the corresponding bug in the ESXi runtime will be fixed soon.
@Zerpet was discovering that issue and was in touch with that team.

ChunyiLyu: As I said in standup, I am aware of this fix: https://gitlab.eng.vmware.com/core-build/spherelet-agent/commit/31b0f133d2fca969b4e561d8f05a26b1f08dd498. However, It's unclear to me when this fix can be released and how often does user upgrade while using these environments with spherelet. As long as we are dropping all known unneeded capabilities, I don't see much difference of security protection between the two different approaches. In addition to that, this is only the security context for our initContainer, which is ephemeral and not long running. IMO we should merge this update to make sure that our operator work with spherelet as soon as possible. Thoughts? @mkuratczyk  @coro

ansd: Oh right, I thought the bug mentioned in standup was a different one. My bad.
In that case, let's merge this PR as is and revert this change in the future once the fix is rolled out?

ansd: Feel free to add a comment like `// Remove default capabilities allowed by Docker except for CHOWN and FOWNER`. Otherwise, it's hard to read what capabilities the container will run with.

ansd (approved):

PR #511, opened 2020-12-03
Ability to skip reconcilation by setting annotation

### LINKED ISSUE #509: Add an annotation that will prevent cluster restart
As a user I would like to decide which clusters will not go through restart (rolling update is a more accurate term) when the operator has a change that requires node restart

Adding an annotation that will tell the operator controller not to update this cluster now will help user control the timing of restarts

ChunyiLyu:
This closes #509

## Summary Of Changes

- when 'rabbitmq.com/skipReconcile' is set to true on a rabbitmqcluster CR, Reconcile() will exit at the beginning of method
- Reconcile() logs and publishes event to warn users that it has skipped updating a CR
- tested only in integration level; I don't think this feature needs to be tested in system level since the operator uses annotations to indicate whether statefulSet or configmap is updated or not, and integration level covers this

### DIFF
--- a/controllers/rabbitmqcluster_controller.go
+++ b/controllers/rabbitmqcluster_controller.go
@@ -48,8 +48,9 @@ var (
 )
 
 const (
-	ownerKey  = ".metadata.controller"
-	ownerKind = "RabbitmqCluster"
+	ownerKey                = ".metadata.controller"
+	ownerKind               = "RabbitmqCluster"
+	skipReconcileAnnotation = "rabbitmq.com/skipReconcile"
 )
 
 // RabbitmqClusterReconciler reconciles a RabbitmqCluster object
@@ -99,6 +100,16 @@ func (r *RabbitmqClusterReconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
 			"namespace", rabbitmqCluster.Namespace,
 			"name", rabbitmqCluster.Name)
 		return ctrl.Result{}, r.prepareForDeletion(ctx, rabbitmqCluster)
+	}
+
+	// exit if skip reconcile annotation is set to true
+	if v, ok := rabbitmqCluster.Annotations["rabbitmq.com/skipReconcile"]; ok && v == "true" {
+		logger.Info("Skip reconciling RabbitmqCluster",
+			"namespace", rabbitmqCluster.Namespace,
+			"name", rabbitmqCluster.Name)
+		r.Recorder.Event(rabbitmqCluster, corev1.EventTypeWarning,
+			"SkipReconcile", "annotation 'rabbitmq.com/skipReconcile' is set to true")
+		return ctrl.Result{}, err
 	}
 
 	// Ensure the resource have a deletion marker
--- a/controllers/rabbitmqcluster_controller_test.go
+++ b/controllers/rabbitmqcluster_controller_test.go
@@ -34,7 +34,7 @@ import (
 )
 
 const (
-	ClusterCreationTimeout = 5 * time.Second
+	ClusterCreationTimeout = 10 * time.Second
 	ClusterDeletionTimeout = 5 * time.Second
 )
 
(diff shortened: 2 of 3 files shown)


controllers/rabbitmqcluster_controller.go
```diff
@@ -101,6 +102,16 @@ func (r *RabbitmqClusterReconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
 		return ctrl.Result{}, r.prepareForDeletion(ctx, rabbitmqCluster)
 	}
 
+	// exit if skip reconcile annotation is set to true
+	if v, ok := rabbitmqCluster.Annotations["rabbitmq.com/skipReconcile"]; ok && v == "true" {
```
ansd, line 106: Let's use the constant defined above.

controllers/rabbitmqcluster_controller.go
```diff
@@ -101,6 +102,16 @@ func (r *RabbitmqClusterReconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
...
+		logger.Info("Skip reconciling RabbitmqCluster",
+			"namespace", rabbitmqCluster.Namespace,
+			"name", rabbitmqCluster.Name)
+		r.Recorder.Event(rabbitmqCluster, corev1.EventTypeWarning,
+			"SkipReconcile", "annotation 'rabbitmq.com/skipReconcile' is set to true")
```
ansd, line 111: Let's use the constant defined above.

ansd (commented):
Should we add user documentation now or will we add docs once the feature is released?

controllers/rabbitmqcluster_controller.go
```diff
@@ -101,6 +102,16 @@ func (r *RabbitmqClusterReconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
...
+		return ctrl.Result{}, err
```
ansd, line 112: Shouldn't we update the status before skipping reconciliation?
Skipping reconciliation means we don't update the actual cluster to match the desired spec.
However, reporting the current status should still be done although reconciliation is skipped.
Imagine a pod dies. The `AllReplicasReadyCondition` condition should still be updated to be `false`.

ansd (changes requested):

controllers/rabbitmqcluster_controller.go
ansd: Also, in the current implementation deletion is allowed even when reconciliation is skipped.
When reconciliation is skipped and the desired state is `Delete`, the operator shouldn't reconcile, i.e. it shouldn't call `prepareForDeletion`.

ansd:
Based on the conversation with @mkuratczyk and @yaronp68:
* We think it makes sense to use a label instead of an annotation. Using a label has the advantage that a user can set it via `kubectl label` and easily query all instances where reconciliation is paused.
* We will add `kubectl rabbitmq pause-reconciliation` and `kubectl rabbitmq resume-reconciliation` commands. We should rename the label `rabbitmq.com/skipReconcile` to `rabbitmq.com/pauseReconciliation` to use the same wording.
* If reconciliation is paused, we should update the Status `NoWarnings` to be `false` with Reason `reconciliation paused`. This has the advantage that we also see the `lastTransitionTime`, i.e. since when reconciliation has been paused. We should leave the status `ReconcileSuccess` to be `true` (interpreted as "successfully skipped reconciliation").

Zerpet:
> Based on the conversation with @mkuratczyk and @yaronp68:
> 
>     * We think it makes sense to use a label instead of an annotation. Using a label has the advantage that a user can set it via `kubectl label` and easily query all instances where reconciliation is paused.
> 
>     * We will add `kubectl rabbitmq pause-reconciliation` and `kubectl rabbitmq resume-reconciliation` commands. We should rename the label `rabbitmq.com/skipReconcile` to `rabbitmq.com/pauseReconciliation` to use the same wording.
> 
>     * If reconciliation is paused, we should update the Status `NoWarnings` to be `false` with Reason `reconciliation paused`. This has the advantage that we also see the `lastTransitionTime`, i.e. since when reconciliation has been paused. We should leave the status `ReconcileSuccess` to be `true` (interpreted as "successfully skipped reconciliation").

I think we should use an annotation because it has functional/behavioral implications. We can use `kubectl annotate` to easily annotate resources.

From the Kubernetes [docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/), annotations can be used for "directives from the end-user to the implementations to modify behavior or engage non-standard features".

Agreed that, by using annotations, we don't have the ability to query all the custom resources that are skipping reconciliation. If we want to include this feature, our plugin could also add a label, or the same label-value pair as the annotation-value pair, to the custom resource.

I believe that the use case to select all custom resources, from a given or all namespaces, that are skipping reconciliation should be uncommon because, as a user, you want the Operator to manage your RabbitMQ clusters.

controllers/rabbitmqcluster_controller.go
huangjiasingle: rabbitmqCluster.Annotations need to avoid go panic.

PR #521, opened 2020-12-09
Include k8s 1.20 in PR system tests

ChunyiLyu:

## Summary Of Changes

- Add k8s 1.20 in PR system tests
- Bump to k8s 1.19.1 to the latest 1.19.4

## Additional Context

Not removing 1.17 from system tests matrixes just yet since many people are still using older version of k8s and it's not adding any feedback times to our PR workflow since it's parallel anyways

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -33,7 +33,7 @@ jobs:
         export GOPATH=$HOME/go
         export PATH=$PATH:$GOPATH/bin
         make install-tools
-        kind create cluster --image kindest/node:v1.19.1
+        kind create cluster --image kindest/node:v1.20.0
         kubectl apply --validate=false -f https://github.com/jetstack/cert-manager/releases/latest/download/cert-manager.crds.yaml
         # Create CRD PodMonitor without running Prometheus operator
         curl https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/master/bundle.yaml | sed "s/replicas: 1$/replicas: 0/" | kubectl apply -f -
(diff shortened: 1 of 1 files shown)


.github/workflows/pr.yml
```diff
@@ -46,7 +46,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        k8s: [v1.17.11, v1.18.8, v1.19.1]
```
MirahImage, line 49: Did we decide to keep 1.17? Has it reached end of support?

.github/workflows/pr.yml
```diff
@@ -33,7 +33,7 @@ jobs:
         export GOPATH=$HOME/go
         export PATH=$PATH:$GOPATH/bin
         make install-tools
-        kind create cluster --image kindest/node:v1.19.1
+        kind create cluster --image kindest/node:v1.19.4
```
coro, line 36: 1.20 is GA, right? So I feel we should probably be testing on 1.20 by default wherever we have a default (like here)

coro (approved):
All good to me, just one question

.github/workflows/pr.yml
MirahImage: Sorry, meant to comment on the new version of this line, not the old one, but you get the point

ChunyiLyu: Dropping support for 1.17 is not a focus of this PR. Many people are still using older version of k8s and it's not adding any feedback times to our PR workflow for having one more version to test since checks are parallel anyways. 

I believe minimal version of k8s support is a product issue and we should discuss with @yaronp68  before making a decision in a PR.

.github/workflows/pr.yml
ChunyiLyu: Good point and updated!

MirahImage (approved):

PR #523, opened 2020-12-10
Pause reconciliation label

### LINKED ISSUE #509: Add an annotation that will prevent cluster restart

ChunyiLyu:
This closes #509

This PR is an updated version of PR #511. Updated based on feedback https://github.com/rabbitmq/cluster-operator/pull/511#issuecomment-740023563

## Summary Of Changes

- users can set label "rabbitmq.com/pauseReconciliation" to true to signal that they would like to pause reconciliation for this CR
- added helper command: `pause-reconciliation` which adds the label to a cr, `resume-reconciliation` which deletes the label from a cr, and `list-pause-reconciliation-instances` which lists all rabbitmqclusters with the label set to true in the current targeted namespace
- When `Reconcile()` detects the label, it logs, publishes events and sets NoWarnings status condition to false with reason `paused reconciliation`
- Deleting the label, or setting the label to any value other than 'true' will resume reconciliation, and NoWarnings status condition will be set to its original value (it can remain false if the resource requirement is configured wrong for example)
- tested at integration level, system test not needed imo

## Additional Context

**Label vs Annotation**

The purpose of this issue is to provide a way for users to skip reconciliation, in case that a rolling restart of the sts is needed, after an operator upgrade. In this particular use case, it's likely that users would have multiple rabbitmq clusters that they have paused reconciliation at the same time. Labels have the advantage of filtering and therefore better suited for fleet management.

## Local Testing

### DIFF
--- a/controllers/rabbitmqcluster_controller_test.go
+++ b/controllers/rabbitmqcluster_controller_test.go
@@ -34,7 +34,7 @@ import (
 )
 
 const (
-	ClusterCreationTimeout = 5 * time.Second
+	ClusterCreationTimeout = 10 * time.Second
 	ClusterDeletionTimeout = 5 * time.Second
 )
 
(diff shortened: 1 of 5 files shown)


controllers/rabbitmqcluster_controller_test.go
```diff
@@ -1096,6 +1065,104 @@ var _ = Describe("RabbitmqClusterController", func() {
...
+
+				// configMap and statefulSet do not have their update and restart annotations set
+				Consistently(func() map[string]string {
+					cfm := configMap(ctx, cluster, "server-conf")
+					return cfm.Annotations
+				}, 10*time.Second).ShouldNot(HaveKey("rabbitmq.com/serverConfUpdatedAt"))
+				Consistently(func() map[string]string {
+					sts := statefulSet(ctx, cluster)
+					return sts.Annotations
+				}, 10*time.Second).ShouldNot(HaveKey("rabbitmq.com/lastRestartAt"))
```
Zerpet, line 1106: Perhaps an optimisation here could be to expect consistently only in the first expectation. The reasoning here being that, if the first expectation fails after 10 seconds, it's unlikely other expectations are met (in fact test exits right away); on the hand, if the first expectation succeeds, that likely means that the custom resource has been reconciled and subsequent expectations will be met, given that all the fields we are asserting on are set in a single reconcile loop.

Unless you have evidence that this approach causes flakes, I would be in favour of this optimisation, which could save ~20 seconds of test run time.

Zerpet:
Forgot to mention in my review:

The StatefulSet controller also uses labels to identify what Pods it owns; I think it's ok to use labels instead of annotations for the added benefit of easier filtering.

controllers/rabbitmqcluster_controller_test.go
ChunyiLyu: Good point. Updated!

Zerpet:
We should document this feature in RabbitMQ docs, ideally before we release a new version.

mkuratczyk (approved):
Works great.

PR #524, opened 2020-12-11
Mount tls secrets using projected volume not subpath

### LINKED ISSUE #506: Add support for TLS rotation without node/cluster restart
RabbitMQ supports TLS credential rotation without restart - if the cert/key files change on disk, they will be immediately used for new connections (currently open connections stay open). Unfortunately, [we currently use `subPath`](https://github.com/rabbitmq/cluster-operator/blob/e4c3759d2029bce677ded63f5c37bc56319da19c/internal/resource/statefulset.go#L417-L439) to mount TLS credentials and containers with `Secrets` (and `ConfigMaps`) mounted through `subPath` do not receive updates. That means that TLS creds rotation requires a restart to be picked up.

I'd like to be able to rotate TLS without restart, therefore, if possible, `subPath` should not be used.

Perhaps the limitation around `subPath` could be removed from Kubernetes but even if, then it would take a long time before our users would be able to take advantage of that.

ChunyiLyu:
This closes #506 

## Summary Of Changes

- secrets and configmaps mounted with subpath do not not get updated in pods when changed, as reported in
issue: https://github.com/kubernetes/kubernetes/issues/50345
- RabbitMQ itself supports TLS credential rotation without restart. By mounting tls secrets without using subpath, rabbitmq pods will pick up cert changes and support tls rotation without server restart.

## Local Testing

```
```

```
$ kubectx dev-bunny
$ make destroy deploy-dev
$ make system-tests
```

### DIFF
--- a/controllers/reconcile_tls_test.go
+++ b/controllers/reconcile_tls_test.go
@@ -3,6 +3,7 @@ package controllers_test
 import (
 	"context"
 	"fmt"
+	"k8s.io/utils/pointer"
 
 	rabbitmqv1beta1 "github.com/rabbitmq/cluster-operator/api/v1beta1"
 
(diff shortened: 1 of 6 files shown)


mkuratczyk (approved):
Thanks!

MirahImage (approved):

PR #534, opened 2020-12-24
Use namespaced name for RabbitMQ cluster_name

ansd:
Before this PR, the Grafana RabbitMQ Overview dashboard was mixing metrics of multiple RabbitMQ clusters if they have the same (CRD) name but are in different namespaces.

For example having RabbitMQ cluster A in namespace 1, and a separate RabbitMQ cluster with the same name A in namespace 2, the Grafana RabbitMQ Overview dashboard only allowed filtering for RabbitMQ cluster A which then mixed / combined all metrics of the two separate clusters (e.g. showing 6 nodes in total although each had 3 nodes).

Two solutions come to my mind:
1. solution as implemented in this PR is to prepend the namespace to the `cluster_name`.

2. solution is to build new Grafana dashboards specifically for RabbitMQ on K8s.  That's possible because the `rabbitmq_identity_info` metric contains a `namespace` label: e.g. `rabbitmq_identity_info{rabbitmq_cluster="myrabbit", namespace="mynamesspace"}`.

I think it's simpler and cheaper to go with the 1st solution instead of maintaining a separate set of dashboards for K8s (2nd solution).

The (only) drawback of the 1st solution is that renaming existing clusters is a breaking change.
The [federation plugin docs](https://www.rabbitmq.com/federation-reference.html#cluster-name) state:
> The federation plugin uses the cluster name defined within the server to identify itself to other nodes in the federation graph. The default is constructed from the RabbitMQ node name and the fully-qualified domain name of the first node to form the cluster.

Therefore, existing clusters using federation would need to overwrite the `cluster_name` via `spec.rabbbitmq.additionalConfig` to avoid having the federated queues being doubled.

What are you thoughts @gerhard @mkuratczyk?

### DIFF
--- a/internal/resource/configmap.go
+++ b/internal/resource/configmap.go
@@ -81,7 +81,7 @@ func (builder *ServerConfigMapBuilder) Update(object runtime.Object) error {
 	}
 	defaultSection := cfg.Section("")
 
-	if _, err := defaultSection.NewKey("cluster_name", builder.Instance.Name); err != nil {
+	if _, err := defaultSection.NewKey("cluster_name", fmt.Sprintf("%s/%s", builder.Instance.Namespace, builder.Instance.Name)); err != nil {
 		return err
 	}
 
--- a/internal/resource/configmap_test.go
+++ b/internal/resource/configmap_test.go
@@ -22,10 +22,11 @@ import (
 	k8sresource "k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	defaultscheme "k8s.io/client-go/kubernetes/scheme"
 )
 
-func defaultRabbitmqConf(instanceName string) string {
+func defaultRabbitmqConf(instance types.NamespacedName) string {
 	return iniString(`
 cluster_formation.peer_discovery_backend = rabbit_peer_discovery_k8s
 cluster_formation.k8s.host               = kubernetes.default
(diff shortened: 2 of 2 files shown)


mkuratczyk:
At first glance, I'd prefer the other option to be honest. I don't think we need to maintain a separate set of dashboards - it should be possible to achieve this with a Grafana variable or something. I guess even with this change, the same problem would occur if we had two Kubernetes clusters and a same-name RabbitMQ and namespace in both (I'd expect this to be quite common in multi-AZ or DR environments). I'd like to at least investigate the alternative before going this way.

ansd:
I just tried out option 2:

Filtering by namespace can be achieved by adding another Grafana variable do the dashboard with Name `namespace` and Query `label_values(rabbitmq_identity_info, namespace)`. All panels in the RabbitMQ-Overview dashboard which filter by RabbitMQ cluster (currently 41) would then need to add the namespace to the query like `rabbitmq_identity_info{rabbitmq_cluster="$rabbitmq_cluster", namespace="$namespace"}`.

The same variable and adapted queries need to be applied to the other dashboards as well (e.g. Erlang-Distribution dashboard, RabbitMQ-Quorum-Queues-Raft dashboard).

If we go with option 2, either we will need to support a second set of RabbitMQ Grafana dashboards specifically for K8s (containing the new variable and queries) or users need to add the namespace variable to the dashboards and adapt the queries manually (as I just did).

Therefore, I think the change in this PR is simpler because it's less maintenance for us and does not require users to re-configure the dashboards.

Monitoring RabbitMQ clusters with the same name and namespace across different K8s clusters is yet another use case not solved by this PR. There are some blog posts out there on how to do this by adding more labels. It's not a RabbitMQ specific issue.

gerhard:
I thought about this, and option 2 is preferable, but without needing to create new set of K8S-specific dashboards.

If the metrics don't have a `namespace` label, then I would expect there to be no option in the Namespace variable drop-down, and we get all metrics, without any `namespace` filtering.

Is this something that you want to try out via https://github.com/rabbitmq/rabbitmq-server/blob/master/deps/rabbitmq_prometheus/docker/docker-compose-overview.yml?

ansd:
Thanks @gerhard, I'll give it a try :)

gerhard:
Fantastic! Give me a shout if you get stuck or something doesn't work as expected.

PR #533, opened 2020-12-23
Add support for  prometheus TLS

Gsantomaggio:
This closes https://github.com/rabbitmq/cluster-operator/issues/479

**Note to reviewers:** remember to look at the commits in this PR and consider if they can be squashed

## Summary Of Changes
Add TLS support for prometheus

## Local Testing

Please ensure you run the unit, integration and system tests before approving the PR.

To run the unit and integration tests:

```
$ make unit-tests integration-tests
```

You will need to target a k8s cluster and have the operator deployed for running the system tests.

For example, for a Kubernetes context named `dev-bunny`:
```
$ kubectx dev-bunny
$ make destroy deploy-dev
# wait for operator to be deployed
$ make system-tests
```

PR #539, opened 2021-01-04
Add prometheus  tls example

Gsantomaggio:
## Summary Of Changes

With this PR https://github.com/rabbitmq/cluster-operator/pull/533 it is possible to enable the Prometheus tls endpoint. 

Update the example to show how to use them

### DIFF
--- a/docs/examples/prometheus/kube-prometheus-stack-values-tls.yaml.example
+++ b/docs/examples/prometheus/kube-prometheus-stack-values-tls.yaml.example
@@ -1,4 +1,3 @@
-## Version > 1.3.0
 ---
 prometheus:
   additionalPodMonitors:
--- a/docs/examples/prometheus/kube-prometheus-stack-values-tls.yaml.example
+++ b/docs/examples/prometheus/kube-prometheus-stack-values-tls.yaml.example
@@ -1,14 +1,15 @@
 ## Version > 1.3.0
 ---
-additionalPodMonitors:
-- name: rabbitmq
- podMetricsEndpoints:
- - port: prometheus-tls
-   scheme: https
-   tlsConfig:
-     insecureSkipVerify: true
- selector:
-   matchLabels:
-     app.kubernetes.io/component: rabbitmq
- namespaceSelector:
-   any: true
+prometheus:
+  additionalPodMonitors:
+   - name: rabbitmq
+     podMetricsEndpoints:
+     - port: prometheus-tls
+       scheme: https
+       tlsConfig:
+           insecureSkipVerify: true
+     selector:
+       matchLabels:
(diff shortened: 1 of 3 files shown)


mkuratczyk (changes requested):
TLS values file seems to be incorrect:
```
helm upgrade --install prometheus prometheus-community/kube-prometheus-stack --values=kube-prometheus-stack-values-tls.yaml.example
Error: failed to parse kube-prometheus-stack-values-tls.yaml.example: error converting YAML to JSON: yaml: line 4: did not find expected key
```

Gsantomaggio:
@mkuratczyk fixed!

docs/examples/prometheus/kube-prometheus-stack-values-tls.yaml.example
```diff
@@ -0,0 +1,15 @@
+## Version > 1.3.0
```
ansd, line 1: Should we add to the comment that this is the rabbitmq/cluster-operator version?
Otherwise, this might be confusing to users?

docs/examples/prometheus/kube-prometheus-stack-values-tls.yaml.example
```diff
@@ -0,0 +1,15 @@
...
+---
+prometheus:
+  additionalPodMonitors:
+   - name: rabbitmq
+     podMetricsEndpoints:
+     - port: prometheus-tls
+       scheme: https
+       tlsConfig:
+           insecureSkipVerify: true
```
ansd, line 10: Prometheus scrapes the Prometheus endpoint of RabbitMQ.
Can we configure Prometheus to trust the TLS certificate of RabbitMQ such that we don't need to skip SSL validation?

ansd (changes requested):

docs/examples/prometheus/kube-prometheus-stack-values-tls.yaml.example
Gsantomaggio: that's a good point, I am not a Prometheus expert maybe @gerhard can help.

Btw, in general, it depends how you create the certificate with self-signed cert you could have:
```
Get "https://10.244.0.29:15691/metrics": x509: cannot validate certificate for 10.244.0.29 because it doesn't contain any IP SANs
```

We had the same problem while we were preparing the [TGIR S01E08: Secure public RabbitMQ clusters
](https://www.youtube.com/watch?v=mPuMBB9_MRI&t=348s)  with the chain browser --> treafik --> rabbitmq tls endpoints and we ended up to add `insecureSkipVerify`.

It is worth to investigate.

mkuratczyk (approved):
LGTM

Issue #537, opened 2021-01-03
cluster operator fails to deploy rabbitmq instance


thorion3006:
## Describe the bug
Rabbitmq cluster operator v1.3.0 fails to deploy rabbitmq instance in EKS v1.18 cluster

## To Reproduce
1. ```kubectl apply -f https://github.com/rabbitmq/cluster-operator/releases/download/v1.3.0/cluster-operator.yml```
2. ```kubectl apply -f rabbitmq.yaml -n resources```
```yaml
apiVersion: rabbitmq.com/v1beta1
kind: RabbitmqCluster
metadata:
  name: rabbitmq
spec:
  affinity:
    podAntiAffinity:
      preferredDuringSchedulingIgnoredDuringExecution:
        - weight: 50
          podAffinityTerm:
            labelSelector:
              matchLabels:
                app.kubernetes.io/name: rabbitmq
            topologyKey: topology.kubernetes.io/zone
      requiredDuringSchedulingIgnoredDuringExecution:
        - labelSelector:
            matchLabels:
              app.kubernetes.io/name: rabbitmq
          topologyKey: kubernetes.io/hostname
  override:
    service:
      spec:
        ports:
          - name: management-http
            protocol: TCP
            port: 15672
            targetPort: 15672
          - name: amqp-tcp
            protocol: TCP
            port: 5672
            targetPort: 5672
    statefulSet:
      spec:
        template:
          spec:
            containers:
              - name: rabbitmq
                env:
                  - name: RABBITMQ_QUORUM_DIR
                    value: /var/lib/rabbitmq/quorum-segments
                volumeMounts:
                  - mountPath: /etc/rabbitmq/rabbitmq_definitions.json
                    name: definitions
                  - mountPath: /var/lib/rabbitmq/quorum-segments
                    name: quorum-segments
                  - mountPath: /var/lib/rabbitmq/quorum-wal
                    name: quorum-wal
            nodeSelector:
              ebs-optimized: 'true'
            volumes:
              - name: definitions
                configMap:
                  name: rabbitmq-definitions # Name of the ConfigMap which contains definitions you wish to import
        volumeClaimTemplates:
          - apiVersion: v1
            kind: PersistentVolumeClaim
            metadata:
              name: quorum-wal
            spec:
              accessModes:
                - ReadWriteOnce
              resources:
                requests:
                  storage: 10Gi
          - apiVersion: v1
            kind: PersistentVolumeClaim
            metadata:
              name: quorum-segments
[...]

PR #545, opened 2021-01-06
Always set PVC override namespace to sts namespace

ChunyiLyu:

## Summary Of Changes

- PR related to bug reported in issue: #537
-  k8s allows pvc template in sts to not have namespace specified and assumes it's the same namespace as the sts
- operator needs to set namespace because controller reference can only be set when both object name and namespace are specified

## Local Testing

Have run unit and integration tests


MirahImage (approved):

mkuratczyk (approved):
Thanks!

Issue #503, opened 2020-11-26
Better way of handling image, operator and user configuration


mkuratczyk:
We discussed the interactions between the image, Operator and user configuration and decided to make the following improvements:

1. Tanzu image will ship its default configuration in `rabbitmq.conf` rather than `conf.d/` as it does right now. This will allow the Operator and its users, to easily override these values if needed. This change has already been implemented.

2. The Operator will not override/mount over `rabbitmq.conf` and instead:
a. mount generated configuration (k8s peer discovery, etc) to `conf.d/10-operatorDefaults.conf` or something similar
b. mount `additionalConfig` as `conf.d/90-additionalConfig.conf` or similar

Justification:
- we don't completely remove image's configuration (by mounting over it) and therefore don't have to repeat it
- all configuration - from the image, from the Operator and from the user are all visible in the running container
- the precedence is as expect - the user has the final say, otherwise they rely on the Operator defaults and then on image defaults (for properties not set by neither the user nor the Operator)

Caveats:
- `rabbitmq` image still ships a file in [conf.d](https://github.com/docker-library/rabbitmq/blob/master/3.8/ubuntu/Dockerfile#L254) that we wouldn't be able to override. This particular setting is not too important - users can just use `rabbitmq:management` which [deletes that files](https://github.com/docker-library/rabbitmq/blob/master/3.8/ubuntu/management/Dockerfile#L6)

Zerpet:
## Context

Started working on this and the changes I'm having in mind to complete this issue are:

- Change the config map resource module to generate a key in the Server conf with just the defaults.
- Use the config map resource module to generate a key in the Server conf with the contents of `spec.rabbitmq.additionalConfig`.
- Update Pod template spec in StatefulSet, so that RabbitMQ container mounts default configuration, default user and spec configuration as files in `conf.d`.

### Next steps

- Manually test how setting multiple sources in a `ProjectedVolume` works, in terms of mounting all the files in one place, without overlapping with the mounted directory i.e. mount files in `conf.d` and preserve the contents of `conf.d` before the mount.
- We are changing the Pod template spec in the StatefulSet, is this a breaking change? If so, how do wemove forward implementing this feature?

Zerpet:
### Context

We can implement this by adding the ConfigMap to the projected volume, it should look like this:

```
[...]
volumes:
      - name: rabbitmq-confd
        projected:
          defaultMode: 420
          sources:
          - secret:
              items:
              - key: default_user.conf
[...]

PR #548, opened 2021-01-07
Place RabbitMQ configuration in confd

Zerpet:
Related to #503

## Summary Of Changes
 - RabbitMQ configurations are placed in conf.d
 - RabbitMQ env and advanced config are still placed in the default location.
 - The Server conf ConfigMap is specified as its own volume only when advanced config or env configurations are set in the RabbitmqCluster Spec.
 - The server configuration is always added to a projected volume.

## Additional Context
N/A

## Local Testing

- There are changes to all test suites: unit, integration and system

### DIFF
--- a/internal/resource/configmap.go
+++ b/internal/resource/configmap.go
@@ -90,42 +90,45 @@ func (builder *ServerConfigMapBuilder) Update(object client.Object) error {
 		return err
 	}
 
+	cfgUserConfiguration := ini.Empty(ini.LoadOptions{})
+	userConfigurationSection := cfgUserConfiguration.Section("")
+
 	if builder.Instance.TLSEnabled() {
-		if err := cfg.Append([]byte(defaultTLSConf)); err != nil {
+		if err := cfgUserConfiguration.Append([]byte(defaultTLSConf)); err != nil {
 			return err
 		}
 		if builder.Instance.DisableNonTLSListeners() {
-			if _, err := defaultSection.NewKey("listeners.tcp", "none"); err != nil {
+			if _, err := userConfigurationSection.NewKey("listeners.tcp", "none"); err != nil {
 				return err
 			}
 		} else {
 			// management plugin does not have a *.listeners.tcp settings like other plugins
 			// management tcp listener can be disabled by setting management.ssl.port without setting management.tcp.port
 			// we set management tcp listener only if tls is enabled and disableNonTLSListeners is false
-			if _, err := defaultSection.NewKey("management.tcp.port", "15672"); err != nil {
+			if _, err := userConfigurationSection.NewKey("management.tcp.port", "15672"); err != nil {
 				return err
 			}
 
-			if _, err := defaultSection.NewKey("prometheus.tcp.port", "15692"); err != nil {
+			if _, err := userConfigurationSection.NewKey("prometheus.tcp.port", "15692"); err != nil {
 				return err
 			}
 		}
 		if builder.Instance.AdditionalPluginEnabled("rabbitmq_mqtt") {
-			if _, err := defaultSection.NewKey("mqtt.listeners.ssl.default", "8883"); err != nil {
+			if _, err := userConfigurationSection.NewKey("mqtt.listeners.ssl.default", "8883"); err != nil {
 				return err
 			}
 			if builder.Instance.DisableNonTLSListeners() {
-				if _, err := defaultSection.NewKey("mqtt.listeners.tcp", "none"); err != nil {
+				if _, err := userConfigurationSection.NewKey("mqtt.listeners.tcp", "none"); err != nil {
 					return err
 				}
 			}
 		}
 		if builder.Instance.AdditionalPluginEnabled("rabbitmq_stomp") {
-			if _, err := defaultSection.NewKey("stomp.listeners.ssl.1", "61614"); err != nil {
+			if _, err := userConfigurationSection.NewKey("stomp.listeners.ssl.1", "61614"); err != nil {
 				return err
 			}
 			if builder.Instance.DisableNonTLSListeners() {
-				if _, err := defaultSection.NewKey("stomp.listeners.tcp", "none"); err != nil {
+				if _, err := userConfigurationSection.NewKey("stomp.listeners.tcp", "none"); err != nil {
 					return err
 				}
 			}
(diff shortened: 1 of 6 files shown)


MirahImage (approved):

internal/resource/configmap.go
```diff
@@ -90,116 +91,115 @@ func (builder *ServerConfigMapBuilder) Update(object client.Object) error {
...
 		}
 	}
 
-	rmqProperties := builder.Instance.Spec.Rabbitmq
-	if err := cfg.Append([]byte(rmqProperties.AdditionalConfig)); err != nil {
-		return fmt.Errorf("failed to append spec.rabbitmq.additionalConfig: %w", err)
-	}
-
-	var rmqConfBuffer bytes.Buffer
+	// TODO refactor: use string builder
```
ChunyiLyu, line 201: outdated comment

internal/resource/configmap.go
```diff
@@ -90,116 +91,115 @@ func (builder *ServerConfigMapBuilder) Update(object client.Object) error {
 		return err
 	}
 
+	cfgUserConfiguration := ini.Empty(ini.LoadOptions{})
+	userConfigurationSection := cfgUserConfiguration.Section("")
+
 	if builder.Instance.TLSEnabled() {
-		if err := cfg.Append([]byte(defaultTLSConf)); err != nil {
```
ChunyiLyu, line 94: Naming suggestion:
1)  `cfgUserConfiguration ` to `userConfiguration `
2) `cfg` to `operatorConfiguration`?

internal/resource/configmap.go
```diff
@@ -90,116 +91,115 @@ func (builder *ServerConfigMapBuilder) Update(object client.Object) error {
...
 	}
 
-	rmqProperties := builder.Instance.Spec.Rabbitmq
-	if err := cfg.Append([]byte(rmqProperties.AdditionalConfig)); err != nil {
-		return fmt.Errorf("failed to append spec.rabbitmq.additionalConfig: %w", err)
-	}
-
-	var rmqConfBuffer bytes.Buffer
+	// TODO refactor: use string builder
+	var rmqConfBuffer strings.Builder
```
ChunyiLyu, line 202: nitpick: not a buffer anymore, maybe rename it to a `rmqConfBuilder`?

internal/resource/configmap.go
```diff
@@ -208,7 +208,20 @@ func (builder *ServerConfigMapBuilder) Update(object client.Object) error {
...
+		return fmt.Errorf("failed to append spec.rabbitmq.additionalConfig: %w", err)
+	}
+
+	if _, err := cfgUserConfiguration.WriteTo(&rmqConfBuffer); err != nil {
+		return err
+	}
+
+	configMap.Data["additionalConfig.conf"] = rmqConfBuffer.String()
 
 	updateProperty(configMap.Data, "advanced.config", rmqProperties.AdvancedConfig)
```
ChunyiLyu, line 226: The `additionalConfig` can also be empty right? If users do not configure `spec.rabbitmq.additionalConfig` and do not use TLS or additional plugins. Should we use `updateProperty()` to delete key `additionalConfig.conf` if it's empty?

internal/resource/configmap.go
Zerpet: The string builder is sort of a buffer, specialised for strings, and we use the word "Builder" for a different purpose in the `resource` package.

https://go.googlesource.com/go/+/go1.15.6/src/strings/builder.go#15

internal/resource/configmap_test.go
```diff
@@ -243,26 +233,20 @@ CONSOLE_LOG=new`
...
-
-prometheus.ssl.certfile = /etc/rabbitmq-tls/tls.crt
-prometheus.ssl.keyfile  = /etc/rabbitmq-tls/tls.key
-prometheus.ssl.port     = 15691
-
-management.tcp.port     = 15672
-prometheus.tcp.port     = 15692
-
-`)
+				expectedRabbitmqConf := iniString("ssl_options.certfile  = /etc/rabbitmq-tls/tls.crt\n" +
```
ChunyiLyu, line 236: What's the benefit of switching from comparing the config as a string to using iniString to construct the config on the fly? Personally, I found the `initString()` section harder to understand since it include new line characters `\n` manually and I need to parse it in my brain to understand what we are comparing 🤔  Thoughts?

internal/resource/configmap_test.go
```diff
@@ -343,42 +312,38 @@ prometheus.ssl.cacertfile = /etc/rabbitmq-tls/ca.crt
 					instance.Spec.TLS.CaSecretName = "tls-mutual-secret"
 					instance.Spec.Rabbitmq.AdditionalPlugins = additionalPlugins
 
-					expectedRabbitmqConf := iniString(defaultRabbitmqConf(builder.Instance.Name) + `
-			ssl_options.certfile   = /etc/rabbitmq-tls/tls.crt
-			ssl_options.keyfile    = /etc/rabbitmq-tls/tls.key
-			listeners.ssl.default  = 5671
+					expectedRabbitmqConf := iniString("ssl_options.certfile   = /etc/rabbitmq-tls/tls.crt\n" +
```
ChunyiLyu, line 315: Similar to the comment above, what's the benefit of using iniString here? I found it easier to read in the previous format 🤔

internal/resource/configmap.go
ChunyiLyu: Good!

ChunyiLyu (changes requested):
Looks good, but there is one thing @ansd and I would like to suggest and have a discussion with you before merging.

Regarding operator configurations and users provided configurations, the current PR have the default operator configurations in `10-operatorDefaults.conf`. TLS, plugins related plus users provided  `spec.rabbitmq.additionalConfig` all in `90-additionalConfig.conf`. We are wondering whether we can move TLS and plugins configurations that the operator sets out of  `90-additionalConfig.conf` to the same place as other operator defaults. (we can rename `10-operatorDefaults.conf` to something like `10-operatorConfig.conf` ). 

The benefit of doing it would be we won't rely on the order of loading the configurations to allow users to overwrite anything that the operator sets regarding TLS and plugins. For example, if users want to specify a different STOMP tls port and the operator sets STOMP TLS
in the same configuration object, the precedence depends on the order of code, which is implicit. If we move everything that the operator sets in one place, the precedence is clearer.

In addition to that, for users that are not familiar with how the operator generates `conf.d` files, when they jump on the pod and see that there are plenty of configurations that they didn't provide in `90-additionalConfig.conf`,  it could be confusing for them.

@Zerpet Thoughts?

internal/resource/configmap_test.go
Zerpet: We were using `initString()` before this PR. I'm not saying your comment is not valid, but it should be addressed in a different PR.

I'm switing to concatenating strings using `+` over a multi-line string because the multi-line string breaks folding in NeoVim :v:

internal/resource/configmap.go
Zerpet: Good catch, the additional config can potentially be empty. Adding conditional logic here will have implications in the StatefulSet builder. Given that RabbitMQ can handle having an empty file and in the spirit of #KeepItSimple, I'd say we leave it as it is.

internal/resource/configmap_test.go
ChunyiLyu: I know that we are using iniString() before. What I meant was we were not using iniString() to construct the expected configurations, but rather concatenating most of the string. I think having the expected configurations to just be the plain string as it was before this change is easier to read. I disagree that this is a separate issue. 

I'm not sure whether I would like to merge this change for an editor issue. Maybe @ansd could add his perspective here since he also uses vim?

ansd: Just had a call with @Zerpet:
`foldmethod=indent` doesn't work with these multi line strings (because it only checks indentation).
`foldmethod=syntax` does work (since it folds by Golang syntax, e.g. collapsing an entire test case).

internal/resource/configmap.go
ChunyiLyu: Sounds good!

internal/resource/configmap_test.go
ChunyiLyu: Awesome! I think we can revert some of these changes now considering vim folding problem is solved 👍

Zerpet:
Implemented the latest changes requested and based on the discussion we had over Zoom. The additional test is [here](https://github.com/rabbitmq/cluster-operator/pull/548/files#diff-09e10b9bd4fca528525752ae36c433dab31307afdaf8268dd186b372d63c3ff7R279-R297).

ChunyiLyu (approved):
Looks good!

PR #554, opened 2021-01-08
Make logger configurable

### LINKED ISSUE #551: Set log_level through env variable or a flag
Right now the log level for the cluster operator is set at DEBUG, and it's generating a lot of noise. I tried to look for a flag, but the documentation doesn't mention anything about the configuring the cluster operator.

To be able to set log verbosity through either a flag or env variable. Also detailed documentation on available flags for the operator.

ansd:
Before this PR, the cluster operator always used a hard-coded logger dev mode, even in production.

After this PR, the logger is prod mode by default.
This means from now on, by default, the logger will log at `Info` level and output JSON.

When developing locally deploying the operator using `make deploy-dev` and `make deploy-kind`, the logger will use dev mode.

Furthermore, users can now configure more zap flags (see https://github.com/kubernetes-sigs/controller-runtime/blob/66537ca5b7439b06f2f3b08901640f934834c9a1/pkg/log/zap/zap.go#L227-L233)

For example, they can set the log level as follows:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
 ...
spec:
 ...
    spec:
      containers:
      - name: operator
        command:
        - /manager
        args:
        - zap-log-level=error
```

This resolves #551.

For more context, read https://sdk.operatorframework.io/docs/building-operators/golang/references/logging/.

### DIFF
--- a/config/default/overlays/dev/manager_image_patch.yaml
+++ b/config/default/overlays/dev/manager_image_patch.yaml
@@ -17,3 +17,4 @@ spec:
       - image: ((operator_docker_image))
         name: operator
         imagePullPolicy: Always
+        args: ["--zap-devel"]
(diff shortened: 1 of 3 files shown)


coro (approved):
LGTM!

ChunyiLyu (approved):
We should document this improvement as well.

PR #559, opened 2021-01-14
Change default operator metrics port

ansd:
The new port matches the one listed in https://github.com/prometheus/prometheus/wiki/Default-port-allocations.

### DIFF
--- a/main.go
+++ b/main.go
@@ -44,7 +44,7 @@ func init() {
 
 func main() {
 	var metricsAddr string
-	flag.StringVar(&metricsAddr, "metrics-addr", ":12345", "The address the metric endpoint binds to.")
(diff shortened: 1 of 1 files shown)


coro (approved):
I approve of the change, but I think this results in a breaking change for anyone who a) consumes the operator metrics on our previously-default port 12345 and b) did not set this flag in the operator deployment spec. I'm fine with that for now as long as the next release notes mention the breaking change.

mkuratczyk:
We haven't been published any Operator metrics so there is no-one using the old port.

mkuratczyk (approved):

ansd:
@coro good point, let's mention this change in the next release notes.

PR #560, opened 2021-01-14
Run system tests against multiple RabbitMQ versions

ansd:


.github/workflows/pr.yml
```diff
@@ -46,12 +46,17 @@ jobs:
     strategy:
       matrix:
         k8s: [v1.17.11, v1.18.8, v1.19.4, v1.20.0]
+        rabbitmq-image:
+        - rabbitmq:3.8.8-management
+        - rabbitmq:3.8.9-management
+        - rabbitmq:3.8.10-rc.5-management
```
ChunyiLyu, line 52: `3.8-rc-management` could be a better tag because there might be `rc.6` in the future.

Zerpet (approved):

ChunyiLyu (changes requested):
Small thing, but please see comment 👍

ChunyiLyu:
@ansd  I don't know what happened with my comment. I was suggesting to use the tag `3.8-rc-management` to test rcs instead of a specific rc version.

ChunyiLyu (approved):
yayy 😄

PR #573, opened 2021-01-21
Update copyright statement to 2021

ChunyiLyu:
## Summary Of Changes

It's 2021 :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,4 +59,4 @@ Please read [contribution guidelines](CONTRIBUTING.md) if you are interested in 
 
 ## Copyright
 
-Copyright 2020 VMware, Inc. All Rights Reserved.
+Copyright 2021 VMware, Inc. All Rights Reserved.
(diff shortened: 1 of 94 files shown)


MirahImage:
It might be worth checking with legal if that's the right language. It might be something like 2020-2021.

ChunyiLyu:
@MirahImage Make sense, I will mark this as draft PR for now.

ansd:
The date as updated in this PR is not correct.

See https://stackoverflow.com/a/2391555
> the date should be the date of first publication

gerhard:
We use a range in rabbitmq-server & all related code. @dumbbell did the last update, he may have an automated way of doing it. Having said that, what @ansd linked to makes sense to me. I'm seeing that K8S is using the same approach, even though some of its dependencies don't.

The best outcome would be for all of RabbitMQ authors to use a consistent approach.

What do you think @michaelklishin?

MirahImage:
We've reached out to our liaison with legal, so hopefully we'll know what the "correct" approach is soon.

kmacoskey:
> liaison with legal

Is that us ? :smile: 

There's a number of changes we would want to make to this PR. We will get synced up with the team on the topic.

ChunyiLyu:
Closing it since there are more work needed for this change.

PR #586, opened 2021-02-03
Add support for -n/-A to select namespace(s)

### LINKED ISSUE #579: Kubectl plugin can't manage resources in other namespaces
The RabbitMQ kubectl plugin only allows managing clusters in the context's default namespace.

Ideally, `kubectl rabbitmq` subcommands would take the same `--namespace` argument as the rest of `kubectl`.

We already create and maintain our cluster resources ourselves (rather than using `kubectl rabbitmq create`). It would've been nice to be able to use the ad-hoc management commands (like `kubectl rabbitmq manage` or `kubectl rabbitmq secrets`), but for now we just manage with the native `kubectl` commands instead.

mkuratczyk:
You can now use `-n NAMESPACE` in most commands or `-A` in some to list
objects across namespaces.

Note: these flags need to be specified immediately after `kubectl
rabbitmq`. One day we'll probably rewrite the whole plugin in Go and it
will be easier to handle more flexible flag parsing.

This closes #579

### DIFF
--- a/bin/kubectl-rabbitmq
+++ b/bin/kubectl-rabbitmq
@@ -271,12 +271,14 @@ install_cluster_operator() {
 main() {
     if [[ "$1" == "--help" ]]; then
         usage
+        exit 0
     fi
 
     while getopts ':n:Ah' opt; do
         case "${opt}" in
         h)
             usage
+            exit 0
             ;;
         A)
             NAMESPACE="--all-namespaces"
--- a/bin/kubectl-rabbitmq
+++ b/bin/kubectl-rabbitmq
@@ -10,59 +10,62 @@
 
 set -euo pipefail
 
+NAMESPACE=""
+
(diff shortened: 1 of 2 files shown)


ChunyiLyu (approved):
LGTM

bin/kubectl-rabbitmq
```diff
@@ -382,18 +408,11 @@ main() {
...
-        ;;
-    "-h")
-        usage
-        ;;
     "version")
         version
         ;;
     *)
         usage
-        exit 1
```
ansd, line 396: Why are we removing exit code 1 on error?

ansd (approved):

bin/kubectl-rabbitmq
mkuratczyk: It's back

PR #590, opened 2021-02-03
Test for documented examples

### LINKED ISSUE #467: Implement regression checking of documented examples
In the docs directory of this repo, we provide a list of example manifests for various features, e.g. mTLS. The form of these manifests are validated in the upgrade pipeline, but there is no automated checking that the manifests deploy a RabbitMQ that does what the example intends. There is no guarantee that the examples displayed will work for all future Kubernetes server versions, or operator versions.

We should create a framework that all of the examples can use to verify they function across supported K8s versions in the operator CI. This could just be a YAML containing instructions to deploy the example in kind and checks to run.

Zerpet:
This closes #467

## Summary Of Changes
The examples now have test files to assert that RabbitMQ has been
configured as intended by the example. Some examples do not have tests
because they are either very basic or its use case is sufficiently
tested in code.

## Additional Context

Original proposal https://github.com/rabbitmq/cluster-operator/issues/467#issuecomment-761077982

## Local Testing

1. Provision a GKE cluster
1. Install cert manager
1. For each test, run `setup.sh` then `test.sh`


docs/examples/import-definitions/rabbitmq.yaml
```diff
@@ -12,7 +12,7 @@ spec:
             containers:
             - name: rabbitmq
               volumeMounts:
-              - mountPath: /path/to/exported/definitions.json
+              - mountPath: /path/to/exported/ # filename left out intentionally
```
ansd, line 15: Should we use a more sensible path here?
I know that this is just an example. However, as user I wouldn't know what to choose as path here.
Does `/etc/rabbitmq/` make sense here?

michaelklishin: `/etc/rabbitmq/definitions.json` sounds reasonable, this is a piece of configuration. We can also update the docs on rabbitmq.com but FWIW so far I haven't seen any complaints about this example.

Zerpet: What I like about `/path/to/exported/` is that it's clearly an example/dummy value and it also JustWorks™. Given the path is a dummy value, I think a user would be inclined to change it to something more sensible. What worries me about `/etc/rabbitmq/definitions.json` is that it might induce a user to think that the definitions must be in `/etc/rabbitmq`, which is not true.

What do you think?

ansd: Okay, let's leave it as is :)

coro (approved):
This is a great set of changes. Thanks for doing this!

docs/examples/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+Every folder with an example must have a file `test.sh`. This executable Bash file has to assert on
+the state of RabbitMQ to ensure that it has been configured according to the expectations of the example.
+
+If the test requires some preparation or setup, a file `setup.sh` can be provided to be executed
+**before** the RabbitMQ cluster is created. This file must be a Bash executable. This file can assume
+that `kubectl` is configured to execute commands against a working Kubernetes cluster.
+
+The script `test.sh` will be executed **after** `AllReplicasReady` condition is `True` in `RabbitmqCluster`
+object. The script `test.sh` should exit with code 0 is all assertions were successful; the script `test.sh` should
```
MirahImage, line 26: ~is~ if

docs/examples/README.md
```diff
@@ -0,0 +1,39 @@
...
+exit with non-zero code if any test or assertion failed. The same is expected for `setup.sh`.
+
+If the example should not run any tests because of the reasons mentioned above, the folder should contain
+a file `.ci-skip`, so that the example is not considered in our tests.
+
+Once the `test.sh` script has completed, the namespace where the example was applied will be deleted. This allows
+to have a clean slate for the next test to execute. This also means that `test.sh` does not need to
```
MirahImage, line 33: ~to have~ for

MirahImage (changes requested):
A minor bit of grammar nitpicking, but the code itself LGTM

MirahImage (approved):

PR #592, opened 2021-02-04
Test examples action should only apply YAML

Zerpet:
## Summary Of Changes

There are some JSON files in the examples related to RabbitMQ. Kubectl should not apply those JSON files because they don't contain Kubernetes manifests and would cause a failure in the action.

## Additional Context
Related to check errors in #590


ChunyiLyu (approved):
👍

PR #598, opened 2021-02-10
Fail to set plugins (or any other necessary cli commands) should result in ReconcileSuccess set to false

ChunyiLyu:
## Summary Of Changes

When `Reconcilie()` fails to set plugins, feature flags, or rebalance when necessary: 
- publish `failedReconcile` warning events
- set `ReconcileSuccess` to false 

## Additional Context

Removed an integration test because pods are never successfully created in envTest and plugins cannot be successfully set, so `ReconcileSuccess` can never be set to true in integration tests. 

## Local Testing

All tests passing locally

### DIFF
--- a/controllers/reconcile_cli.go
+++ b/controllers/reconcile_cli.go
@@ -25,7 +25,6 @@ func (r *RabbitmqClusterReconciler) runRabbitmqCLICommandsIfAnnotated(ctx contex
 		logger.Info("not all replicas ready yet; requeuing request to run RabbitMQ CLI commands")
 		return 15 * time.Second, nil
 	}
-
 	// Retrieve the plugins config map, if it exists.
 	pluginsConfig, err := r.configMap(ctx, rmq, rmq.ChildResourceName(resource.PluginsConfigName))
 	if client.IgnoreNotFound(err) != nil {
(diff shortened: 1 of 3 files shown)


controllers/reconcile_cli.go
```diff
@@ -95,29 +92,26 @@ func (r *RabbitmqClusterReconciler) runSetPluginsCommand(ctx context.Context, rm
...
-			logger.Error(err, "failed to set plugins",
-				"pod", podName,
-				"command", cmd,
-				"stdout", stdout,
-				"stderr", stderr)
-			return err
+			msg := "failed to set plugins on pod"
+			logger.Error(err, msg, "pod", podName, "command", cmd, "stdout", stdout, "stderr", stderr)
+			r.Recorder.Event(rmq, corev1.EventTypeWarning, "FailedReconcile", fmt.Sprintf("%s %s", msg, podName))
+			return fmt.Errorf(fmt.Sprintf("%s %s: %v", msg, podName, err))
```
Zerpet, line 98: Nitty picky: should we just do `fmt.Errorf("string %s %s %v", a, b, c)` ?
I think we can avoid the call with `fmt.Sprintf()`

controllers/reconcile_cli.go
```diff
@@ -95,29 +92,26 @@ func (r *RabbitmqClusterReconciler) runSetPluginsCommand(ctx context.Context, rm
...
-		ctrl.LoggerFrom(ctx).Error(err, "failed to run queue rebalance",
-			"pod", podName,
-			"command", cmd,
-			"stdout", stdout,
-			"stderr", stderr)
-		return err
+		msg := "failed to run queue rebalance on pod"
+		logger.Error(err, msg, "pod", podName, "command", cmd, "stdout", stdout, "stderr", stderr)
+		r.Recorder.Event(rmq, corev1.EventTypeWarning, "FailedReconcile", fmt.Sprintf("%s %s", msg, podName))
+		return fmt.Errorf(fmt.Sprintf("%s %s: %v", msg, podName, err))
```
Zerpet, line 114: Same story:
> Nitty picky: should we just do `fmt.Errorf("string %s %s %v", a, b, c)` ?
> I think we can avoid the call with `fmt.Sprintf()`

controllers/reconcile_cli.go
```diff
@@ -72,12 +71,10 @@ func (r *RabbitmqClusterReconciler) runEnableFeatureFlagsCommand(ctx context.Con
...
-		logger.Error(err, "failed to enable all feature flags",
-			"pod", podName,
-			"command", cmd,
-			"stdout", stdout,
-			"stderr", stderr)
-		return err
+		msg := "failed to enable all feature flags on pod"
+		logger.Error(err, msg, "pod", podName, "command", cmd, "stdout", stdout, "stderr", stderr)
+		r.Recorder.Event(rmq, corev1.EventTypeWarning, "FailedReconcile", fmt.Sprintf("%s %s", msg, podName))
+		return fmt.Errorf(fmt.Sprintf("%s %s: %v", msg, podName, err))
```
Zerpet, line 77: Same as below :D 
> should we just do `fmt.Errorf("string %s %s %v", a, b, c)` ?
> I think we can avoid the call with `fmt.Sprintf()`

controllers/reconcile_cli.go
ChunyiLyu: you are right! `fmt.Sprintf()` is totally redundant here. updated!

MirahImage (approved):

PR #607, opened 2021-02-16
Deploy RMQ 3.8.12 by default

### LINKED ISSUE #595: Bump default RabbitMQ version
We should keep up to date with RabbitMQ releases. Version 3.8.11 addressed a good number of bugs related to quorum queues and it added [individual health checks via management API](https://github.com/rabbitmq/rabbitmq-management/issues/844). 

RabbitMQ 3.8.12 is producing release candidates. We should bump RabbitMQ to 3.8.12 shortly after it gets released.

**Additional context**
Our system tests use a not so great aliveness check via the management API. As part of bumping to 3.8.12, we should change the system tests to use an individual health check e.g. `/api/health/checks/protocol-listener/amqp091`

ChunyiLyu:
This closes #595 

## Summary Of Changes

Deploy RMQ 3.8.12 by default.

## Additional Context

Should we add a warning in our README to warn people about the maintenance mode data loss issue for mirror queues?

## Local Testing

All passing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ under `site/kubernetes`.
 
 ## Supported Versions
 
-The operator deploys RabbitMQ `3.8.10` by default, and supports versions from `3.8.8` upwards. The operator requires Kubernetes `1.17` or newer.
+The operator deploys RabbitMQ `3.8.12` by default, and supports versions from `3.8.8` upwards. The operator requires Kubernetes `1.17` or newer.
 
 ## Versioning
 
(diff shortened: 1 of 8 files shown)


mkuratczyk (approved):
Failing kubectl plugin tests seem to be a flake. We may want to improve them but they seem to be a false negative as far as this change is concerned.

docs/api/rabbitmq.com.ref.asciidoc
```diff
@@ -84,7 +84,7 @@ PersistentVolumeClaim is an embedded version of k8s.io/api/core/v1.PersistentVol
 [id="{anchor_prefix}-github-com-rabbitmq-cluster-operator-api-v1beta1-plugin"]
 ==== Plugin (string) 
 
-
+kubebuilder validating tags 'Pattern' and 'MaxLength' must be specified on string type. Alias type 'string' as 'Plugin' to specify schema validation on items of the list 'AdditionalPlugins'
```
Zerpet, line 88: This comment doesn't make any sense. It looks like an error message from kubebuilder or `code-gen`. We should remove this (not include it for commit) or investigate and fix the root cause.

Zerpet (changes requested):
We might need @coro's input to address the comment below.

docs/api/rabbitmq.com.ref.asciidoc
ChunyiLyu: removed

Zerpet (approved):

PR #610, opened 2021-02-17
Update production ready example

nicksc423:
## Summary Of Changes

Added rabbitmq_peer_discovery_k8s to the production-ready example

## Additional Context

Without the rabbitmq_peer_discovery_k8s plugin the replicas wont join into a cluster.  Instead of a production-ready deployment you end up with 3 separate instances of RabbitMQ

### DIFF
--- a/docs/examples/production-ready/rabbitmq.yaml
+++ b/docs/examples/production-ready/rabbitmq.yaml
@@ -12,6 +12,8 @@ spec:
       cpu: 4
       memory: 10Gi
   rabbitmq:
+    additionalPlugins:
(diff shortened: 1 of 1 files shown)


ChunyiLyu:
`rabbitmq_peer_discovery_k8s` plugin, together with `rabbitmq_prometheus` and `rabbitmq_management` are always enabled by Cluster Operator. You don't need to enable it again through `spec.rabbitmq.additionalPlugins`

See: https://github.com/rabbitmq/cluster-operator/blob/main/internal/resource/rabbitmq_plugins.go#L17

PR #620, opened 2021-02-25
Add Slack notifications on failed GitHub actions

ansd:
We don't add the notifications for failed PRs since it would create too much noise in our Slack channel whenever someone opens a PR where tests are failing.

However, let's add notifications when other GitHub actions fail, e.g. failing to update the krew index. This one was unnoticed for a week.


ChunyiLyu (approved):
Great idea!

PR #621, opened 2021-03-01
Set Pod Management to 'Parallel' and disallow cluster scale down entirely

### LINKED ISSUE #298: Look into setting podManagementPolicy to Parallel

ChunyiLyu:
This closes #298 

## Summary Of Changes

- set `podManagementPolicy` to `Parallel`
- Use cluster_formation.randomized_startup_delay_range to avoid race condition between nodes during initial cluster formation
- This change solves the problem of failing to restart the cluster when all pods are deleted and recreated at once; with podManagementPolicy set to 'OrderedReady', pod 0 is also the first one getting recreated, but it may not be the last node to shut down. This problem was first reported in [community slack](https://rabbitmq.slack.com/archives/CTMSV81HA/p1609882247391800?thread_ts=1609802102.387100&cid=CTMSV81HA)

- **Disallow cluster scale down entirely** After talking to @mkuratczyk @MirahImage @yaronp68, we agreed that the cluster operator should prevent people from scaling down because it's not a properly supported and tested operation. Once `Reconcile()` has detect a scale down request, it will error, publish events, and set ReconcileSuccess to false.

- **PodManagementPolicy is immutable** For existing clusters, operators won't be able to update the policy successfully. Users would need to manually delete the statefulset with `cascading=false` first, and then the operator can recreate the statefulSet with the correct settings. This needs to be mentioned in release notes

## Additional Context

- It would have been a cleaner and more elegant solution if we can mark CRD requirements on `spec.replicas` people from updating it to a less number. That would involve a webhook which adds more component for us to maintain. The current solution is easier to achieve and considering we want to support scale down in the future, it's an OK temporary fix.

- controller-gen was updated in a previous PR, but the crd tag was not updated. I've included the change in this PR since it's a one liner.

### DIFF
--- a/config/crd/bases/rabbitmq.com_rabbitmqclusters.yaml
+++ b/config/crd/bases/rabbitmq.com_rabbitmqclusters.yaml
@@ -10,7 +10,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.4.1
+    controller-gen.kubebuilder.io/version: v0.5.0
   creationTimestamp: null
   name: rabbitmqclusters.rabbitmq.com
 spec:
--- a/controllers/rabbitmqcluster_controller.go
+++ b/controllers/rabbitmqcluster_controller.go
@@ -22,7 +22,7 @@ import (
 
 	"github.com/rabbitmq/cluster-operator/internal/resource"
 	"github.com/rabbitmq/cluster-operator/internal/status"
-	"k8s.io/apimachinery/pkg/api/errors"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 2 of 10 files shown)


mkuratczyk (approved):
Thanks!

MirahImage (approved):

PR #628, opened 2021-03-08
Add support for TLS-only instances to `kubectl rabbitmq manage` command

mkuratczyk:
Use 15672 or 15671 to access Management, based on TLS settings. Previously `kubectl rabbitmq manage TLS-ONLY-INSTANCE` would fail as it would try to port-forward and open 15672 which is not available.


ChunyiLyu (approved):
🌷

PR #642, opened 2021-03-19
Don't treat RabbitMQ CLI stderr output as reconcile error

ansd:
Don't consider RabbitMQ CLI stderr output to be an error.

Between RabbitMQ v3.8.14 and v3.8.15, the command `rabbitmq-queues rebalance quorum` started to output warn messages like the following to stderr:
```
[warn]  Node :"rabbit@r1-server-0.r1-nodes.default" only contains 1 queues, do nothing
```

Although this CLI command exits with `0`, the operator considered this as a reconcile error, retrying to execute that command.
The operator logs the following:
```
2021-03-19T08:39:56.305Z	ERROR
controller-runtime.manager.controller.rabbitmqcluster	Reconciler error
{"reconciler group": "rabbitmq.com", "reconciler kind": "RabbitmqCluster", "name": "r1", "namespace": "default", "error": "failed to run queue rebalance on pod r1-server-0:
```

In general when CLI commands output to stderr, it doesn't necessarily mean that there was an error.
So, in this PR we only consider non-zero exit codes as error.


controllers/reconcile_persistence.go
```diff
@@ -130,10 +131,7 @@ func (r *RabbitmqClusterReconciler) deleteSts(ctx context.Context, rmq *rabbitmq
 
 	if err := retryWithInterval(logger, "delete statefulSet", 10, 3*time.Second, func() bool {
 		_, getErr := r.statefulSet(ctx, rmq)
-		if k8serrors.IsNotFound(getErr) {
-			return true
-		}
-		return false
+		return k8serrors.IsNotFound(getErr)
```
ChunyiLyu, line 134: Nice catch!

ChunyiLyu (approved):
Great! Thank you.

MirahImage (approved):

PR #651, opened 2021-04-03
fix #648: delete the statefulset with precondition set to the correct…

### LINKED ISSUE #648: [BUG] Reading stale RabbitmqCluster information can lead to undesired statefulset deletion
## Describe the bug
We find that rabbitmq-cluster-operator could accidentally delete the statefulset when the controller experiences a restart and talks to one stale apiserver in an HA k8s cluster. After some inspection, we find that the root cause is the staleness populated from the apiserver that makes the controller think the RabbitmqCluster is going to be deleted (with a non-zero `DeletionTimestamp`), while it is actually not. One potential approach to handle this issue is to label each statefulset the UID of the RabbitmqCluster, and check the label before deleting the statefulset.

## To Reproduce
1. Create a RabbitmqCluster named `rabbitmq-cluster1` in a HA k8s cluster. The controller is talking to apiserver1 and the reconcile() will create a statefulset for `rabbitmq-cluster1`.
2. Delete `rabbitmq-cluster1`. Apiserver1 will send the update events with a non-zero `DeletionTimestamp` to the controller and the controller will delete the statefulset of `rabbitmq-cluster1` in `prepareForDeletion`. Meanwhile, apiserver2 is partitioned so its watch cache stops at the moment that `rabbitmq-cluster1` is tagged with a non-zero `DeletionTimestamp`.
3. Create the RabbitmqCluster with the same name again. Now, `rabbitmq-cluster1` and its statefulset get back. However, apiserver2 still holds the stale view that `rabbitmq-cluster1` has a non-zero `DeletionTimestamp` and is about to be deleted.
4. The controller crashes due to some node failure and restarts. This time the controller talks to the stale apiserver2. The restarted controller reads the stale information about `rabbitmq-cluster1` from apiserver2 that `rabbitmq-cluster1` has a non-zero `DeletionTimestamp`. Since the controller identifies the statefulset only with the name of the RabbitmqCluster, the statefulset belonging to the newly created RabbitmqCluster will be deleted in `prepareForDeletion`.

**Expected behavior**
The controller should be able to differentiate the statefulsets belonging to different RabbitmqCluster instances with the same name (in history). When reading the stale information about (previously deleted) RabbitmqCluster, the controller should be able to check whether the existing statefulset really belongs to that RabbitmqCluster (e.g., by using the UID) before perform any deletion.

- RabbitMQ: 3.8.12
- Kubernetes: 1.18.9
 
## Additional context
We are willing to issue a PR to help fix this issue.
As mentioned above, we can label each statefulset with the UID of the RabbitmqCluster, and check the label before deleting the statefulset to ensure we are deleting the correct statefulset.

srteam2020:
… UID

This closes #648

## Summary Of Changes
Ensure that the sts to be deleted has the correct uid (besides the name and namespace). A sts uid is a correct one if the sts's controller ref matches the current rabbitmqCluster's uid.

### DIFF
--- a/controllers/utils.go
+++ b/controllers/utils.go
@@ -7,9 +7,9 @@ import (
 	rabbitmqv1beta1 "github.com/rabbitmq/cluster-operator/api/v1beta1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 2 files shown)


controllers/utils.go
```diff
@@ -64,6 +67,18 @@ func (r *RabbitmqClusterReconciler) statefulSet(ctx context.Context, rmq *rabbit
...
+	if sts, err := r.statefulSet(ctx, rmq); err == nil {
+		if ref := metav1.GetControllerOf(sts); ref != nil {
+			if string(rmq.GetUID()) == string(ref.UID) {
+				return sts.UID, nil
+			}
+		}
+	}
+	return uid, fmt.Errorf("failed to get the uid of the statefulset owned by the current rabbitmqCluster")
+}
+
```
ChunyiLyu, line 81: Two suggestions: 1) I would add a line of comment to explain that  the statefulSet uid is returned successfully only when the controller reference uid matches the rabbitmqCluster uid. 2) I think errors of failing to get the statefulSet, failing to get controller reference,  and reference uid does not match rmq uid are all different failures. The function only returns one type of error at the moment, and it's unclear which step has failed. It's more verbose but the error msg is clearer:

```suggestion
func (r *RabbitmqClusterReconciler) statefulSetUID(ctx context.Context, rmq *rabbitmqv1beta1.RabbitmqCluster) (types.UID, error) {
	var err error
	var sts *appsv1.StatefulSet
	var ref *metav1.OwnerReference

	if sts, err = r.statefulSet(ctx, rmq); err != nil {
		return "", fmt.Errorf("failed to get statefulSet: %s", err.Error())
	}

	if ref = metav1.GetControllerOf(sts); ref == nil {
		return "", errors.New("failed to get controller reference for statefulSet")
	}

	if string(rmq.GetUID()) != string(ref.UID) {
		return "", errors.New("statefulSet not owner by current RabbitmqCluster")
	}

	return sts.UID, nil
}
```

ChunyiLyu (commented):
Thanks for taking the time! Looks good overall. Please see my comments.

srteam2020:
@ChunyiLyu Thank you for the comments! I will try to address them today or tmr.

ChunyiLyu (approved):
Thanks for taking my suggestions onboard 👍

Zerpet (approved):
Looks good to me. Thanks for the contribution!

PR #658, opened 2021-04-12
Errors when received PVC shrink requests

### LINKED ISSUE #653: [BUG] `deleteSts` and `updatePVC` get invoked unexpectedly when reading stale information from apiserver
## Describe the bug
We find that after a restart the controller could accidentally invoke `deleteSts` and `updatePVC`, which leads to unexpected statefulset deletion and a try to resize PVC. We find that the root cause is the staleness populated from the apiserver that makes the controller think the "desired" capacity (from the stale RabbitmqCluster spec) does not equal the current capacity in `needsPVCResize` and invokes `deleteSts` (in `expandPVC`) to delete the existing statefulset. This unexpected statefulset deletion can be fully avoided by checking whether the desired capacity is larger than the current capacity in `needsPVCResize`, given the size of PVC can only grow in one direction. If not, `expandPVC` should not be invoked because it indicates an invalid `desired` capacity  potentially caused by staleness in apiserver.

## To Reproduce
1. Create a RabbitmqCluster `rabbitmq-cluster1` in a HA k8s cluster with PVC sized 10Gi. The controller is talking to apiserver1.
2. Resize the PVC of `rabbitmq-cluster1` to 15 Gi. The controller will find the desired capacity is larger than the current one so it will delete the sts, enlarge the PVC and recreate the sts. Meanwhile, apiserver2 gets stuck at the view before resizing `rabbitmq-cluster1`'s PVC to 15Gi due to some temporary network issues.
3. The controller restarts after a node failure and talks to the stale apiserver2. Reading the stale information about `rabbitmq-cluster1` from apiserver2, the controller thinks the "desired" capacity is still 10Gi that is different from the current capacity (15Gi). The `needsPVCResize` will return true and the controller will delete the statefulset by calling `deleteSts`.

**Expected behavior**
The controller should be able to tell that the "desired" capacity shown in the RabbitmqCluster spec is invalid (stale) by checking if it is larger than the current capacity. Since the PVC size can only be enlarged (rather than reduced), any "desired" capacity value smaller than the current one is invalid (potentially caused by staleness in apiserver), and the controller should directly return with the error indicating staleness, instead of calling `deleteSts` in such case.

- RabbitMQ: 3.8.12
- Kubernetes: 1.18.9
 
## Additional context
We are willing to issue a PR to help fix this issue. We can change the check
```
	if currentCapacity.Cmp(desiredCapacity) != 0 {
		return true, nil
	}
```
to
```
	if currentCapacity.Cmp(desiredCapacity) == -1 {
		return true, nil
	}
```
in `needsPVCResize` (`-1` indicates `desiredCapacity` is larger than `currentCapacity`). So that stale "desired" capacity will not trigger statefulset deletion any more.

ChunyiLyu:
This closes #653 

## Summary Of Changes

- errors, logs, publishes warning events, and set ReconcileSuccess to false when the controller received request to shrink persistence volume sizes
- previously, when controller receives a shrink request, it does not distinguish from an expansion, and would go ahead deleting the statefulSet and then fails to shrink the PVC (PVC shrink gets rejected from k8s server)
- it's tested in intergration tests

## Additional Context

Reported by user in issue: #653

## Local Testing

Tested in integration level; all tests have passed.


controllers/reconcile_persistence.go
```diff
@@ -18,14 +18,14 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func (r *RabbitmqClusterReconciler) reconcilePVC(ctx context.Context, cluster *rabbitmqv1beta1.RabbitmqCluster, current, sts *appsv1.StatefulSet) error {
-	resize, err := r.needsPVCResize(current, sts)
+func (r *RabbitmqClusterReconciler) reconcilePVC(ctx context.Context, rmq *rabbitmqv1beta1.RabbitmqCluster, current, sts *appsv1.StatefulSet) error {
+	resize, err := r.needsPVCExpansion(ctx, rmq, current, sts)
```
MirahImage, line 22: Should we rename `resize` to `expand`?

MirahImage (approved):

controllers/reconcile_persistence.go
ChunyiLyu: Updated

ansd (approved):

PR #664, opened 2021-04-19
Use ServiceMonitor instead of PodMonitor

ansd:
Originally, we used PodMonitors to make Prometheus discover RabbitMQ
nodes to scrape.

Before this PR, when a user deploys both files rabbitmq-podmonitor.yaml
and rabbitmq-podmonitor-tls.yaml, Prometheus ends up scraping two
different instances (`<host>:<port>`) for the same RabbitMQ node for RabbitMQ
clusters where TLS is enabled but where `disableNonTLSListeners` is set to `false`.
In Grafana, this doubles the number of RabbitMQ nodes showing up.

As a solution, we could create a separate PodMonitor per RabbitMQ
cluster either including 15691 or 15692. This comes with the disadvantage
of having many PodMonitors and the user would need to change the
endpoint in the PodMonitor when TLS gets enabled or disabled (since the
RabbitMQ cluster operator doesn't deploy the PodMonitors itself).

Here, we choose a different solution: We use a ServiceMontitor instead
of PodMonitors. A RabbitMQ service will either expose 15691 or 15692,
but never both ports. 15691 is exposed if `disableNonTLSListeners` is set
to `true`.

In this PR, we also delete the example for the kube-prometheus-stack
Helm chart since it was a duplicate.
We delete the separate TLS PodMonitor as well since we can now combine
both 15691 and 15692 ports in a single ServiceMonitor.

The ability to use ServiceMonitors have also been asked for by the
community, see:
* https://rabbitmq.slack.com/archives/CTMSV81HA/p1617990998094000
* https://rabbitmq.slack.com/archives/CTMSV81HA/p1611187267023900


ChunyiLyu (approved):
LGTM. Just need to update docs on rabbitmq.com: https://www.rabbitmq.com/kubernetes/operator/operator-monitoring.html#prom-operator

mkuratczyk (approved):

PR #665, opened 2021-04-22
Add port for stream plugin

Gsantomaggio:
per conversation with @gerhard 

## Summary Of Changes
Add the stream port: `5551` when the stream plugin is enabled.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -44,11 +44,14 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        k8s: [v1.18.15, v1.19.7, v1.20.2]
+        k8s:
+        - v1.18.15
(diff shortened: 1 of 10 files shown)


ansd (commented):
LGTM.

Two questions:
1. Is there a TLS version of streams?
2. Can we use https://github.com/rabbitmq/rabbitmq-stream-go-client to add a [system test](https://github.com/rabbitmq/cluster-operator/tree/main/system_tests)?

Gsantomaggio:
> Is there a TLS version of streams?

Not yet!

> Can we use https://github.com/rabbitmq/rabbitmq-stream-go-client to add a system test?

Why not I will look in it

ansd (approved):

mkuratczyk (approved):

PR #672, opened 2021-04-29
Use Prometheus TLS port if TLS is enabled

ansd:
When TLS is enabled use the prometheus-tls port in the annotation.

This matches the behaviour of the Service:
When TLS is enabled, port 15691 is exposed in the Service.


mkuratczyk (approved):

ansd:
As @mkuratczyk pointed out, this change will cause a rolling restart for TLS enabled instances. Since the changes in this PR are not important and urgent enough to cause a rolling restart, we'll merge this PR once we cut a new release where the StatefulSet gets updated anyways (by some other change).

ChunyiLyu (approved):
Thanks for spotting this! LGTM

andrei-a-g:
Is it possible to have a flag to turn on/off the `prometheus-tls` ? In my case, I'm exposing `amqps` and `management-tls` via NodePort for external access through a Load Balancer, but I want to scrape Prometheus metrics locally .

Issue #662, opened 2021-04-14
Nodes that form a new cluster may not cluster correctly


gerhard:
## Describe the bug

Since https://github.com/rabbitmq/cluster-operator/pull/621 was introduced, nodes that form a new cluster may not cluster correctly.

In our (+@ansd) case, pod quick-rabbit-server-0 formed its own cluster, while quick-rabbit-server-1 & quick-rabbit-server-2 formed a second cluster with the same name as the first cluster. Everything looks healthy from K8S (3 ready pods) & Erlang perspective (6 healthy distribution links), but we have 2 RabbitMQ clusters, one with 1 node & one with 2 nodes, and this is clearly wrong.

## To Reproduce

This is a difficult one to reproduce as it's timing-specific. We are including all the logs and this can be reproduced with https://github.com/rabbitmq/observability-2021/tree/bf77efebc6760e16d8176bc0c7b750204d8b2a7e/talks/emea-tech-talk (private repo available to all maintainers of this repo) using the following steps:

```
make 1.k8s 2.k8s-rabbitmq 4.resolve-first-problem 5.second-problem
```

While `5.second-problem` is not technically needed, it makes it very obvious as to what the problem is:

![image](https://user-images.githubusercontent.com/3342/114733797-9007e580-9d3b-11eb-8620-5f3d27af83a3.png)
![image](https://user-images.githubusercontent.com/3342/114735216-ddd11d80-9d3c-11eb-9931-b7e40983523f.png)

## Version and environment information
- RabbitMQ: 3.8.14
- RabbitMQ Cluster Operator: 1.6.0
- Kubernetes: 1.19.9
- Cloud provider or hardware configuration: k3s (see https://github.com/rabbitmq/observability-2021/blob/bf77efebc6760e16d8176bc0c7b750204d8b2a7e/talks/emea-tech-talk/Makefile#L35)

## Quick fix

Our quick fix was to reset `quick-rabbit-server-0`:

```
rabbitmqctl stop_app
rabbitmqctl reset
rabbitmqctl start_app
```

## Additional context

<details>
<summary>kubectl logs --selector app.kubernetes.io/name=quick-rabbit --prefix=true --tail=-1</summary>
<pre>
...
[...]

PR #675, opened 2021-05-04
Wider randomized_startup_delay_range

mkuratczyk:
We've observed cluster formation failures at a rate of about 7%.
With these values, it seems to be below 1% (not a single failure in 100
tests). The ultimate solution remains to implement locking in the k8s
peer discovery plugin but this change should alleviate the issue until
then.

More info: https://github.com/rabbitmq/cluster-operator/issues/662


ansd (approved):
I also can't see a reason to set min above 0.

MirahImage (approved):

PR #676, opened 2021-05-04
Add support for RabbitMQ alerting without depending on Prometheus Operator

ansd:
Prior to this PR, we documented how to set up RabbitMQ alerting rules assuming Prometheus is installed via the Prometheus Operator. This assumed that we can use custom resources like PodMonitor, ServiceMonitor, PrometheusRule. However, some environments use Prometheus without Prometheus Operator.

Therefore, we should document how to integrate RabbitMQ Alerting with and without Prometheus Operator.

If Prometheus Operator is not used, users can now use the scrape configs in `observability/prometheus/config-file.yml` and the Prometheus rules in `observability/prometheus/rule-file.yml` to get started. The latter is a single auto-generated file which always contains the same rules as the ones defined in `observability/prometheus/rules/`.

I'll soon update https://www.rabbitmq.com/kubernetes/operator/operator-monitoring.html accordingly.


mkuratczyk (approved):
Great stuff!

PR #680, opened 2021-05-05
Explicitly use rabbitmq container in kubectl exec

mkuratczyk:
New versions of `kubectl` print a warning otherwise.
While at it, simplify enable-all-feature-flags by using the "all" feature introduced in 3.8.10.


ansd (approved):
Thank you :)

PR #687, opened 2021-05-11
Explore: classic peer discovery with randomised startup delay

ansd:
Relates #662.

Dynamic peer discovery is not needed for RabbitMQ Clusters deployed by the Cluster Operator. All nodes are known at deploy time. The Cluster Operator knows the number of replicas and their host names.

This PR uses classic peer discovery with a static list of nodes instead of the dynamic `rabbitmq_peer_discovery_k8s` plugin.
In the case of a scale out (i.e. more RabbitMQ nodes added to the RabbitMQ cluster), existing nodes do not get restarted.

**Pros:**
* Nodes can join other nodes that are `running` but not yet `ready` increasing the likelihood of discovering peers
* no sophisticated locking mechanism

**Cons:**
* There might still be cases where clusters do not get formed correctly since this approach still relies on randomised startup delays. @mkuratczyk and I are doing some more testing with different parameters.

**Alternatives:**
* Enforce dedicated node to form the cluster: #689 
* Use a lock in `rabbitmq_peer_discovery_k8s` (e.g. as done in controller-runtime leader election)

### DIFF
--- a/api/v1beta1/rabbitmqcluster_types.go
+++ b/api/v1beta1/rabbitmqcluster_types.go
@@ -280,7 +280,7 @@ type Plugin string
 
 // RabbitMQ-related configuration.
 type RabbitmqClusterConfigurationSpec struct {
-	// List of plugins to enable in addition to essential plugins: rabbitmq_management, rabbitmq_prometheus, and rabbitmq_peer_discovery_k8s.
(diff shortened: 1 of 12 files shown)


ansd:
Closing this PR in favor of https://github.com/rabbitmq/rabbitmq-server/pull/3075.

PR #703, opened 2021-05-26
Add support for Stream over TLS

mkuratczyk:
Automatically configure TLS for the Stream plugin when TLS is enabled.

To test, I'd recommend `openssl s_client -connect INSTANCE-nodes.default.svc.cluster.local:5551`.

Integration tests will come later as the client's TLS support is still being worked on.


ChunyiLyu (approved):
Looks good although I have not manually tested it.

ansd (changes requested):

ansd (approved):

PR #718, opened 2021-06-04
Status refactors

ChunyiLyu:

## Summary Of Changes

- extract a helper function to set `ReconcileSuccess` status condition. We do this for 4 times in our `Reconcile()` func.
- set  `ReconcileSuccess` to false if there is any error from `reconcile_tls()`. I realized that TLS failures in controller was only logging and publishing events, but not setting the status condition. It should! status condition is tested in integration tests.


Zerpet:
The errors in the system tests seem related to an bad cluster formation. I'll trigger a re-run.

Zerpet (approved):

MirahImage (approved):

PR #728, opened 2021-06-10
Generate CRD with new k8s api

ChunyiLyu:
This closes #

## Summary Of Changes

As a followup PR to PRs that has bumped controller-runtime, controller-tools, and k8s api versions:
- https://github.com/rabbitmq/cluster-operator/pull/721
- https://github.com/rabbitmq/cluster-operator/pull/723

There are several new fields from k8s `1.21`. However these are all optional fields with no default value, so this change won't cause any rolling restart or behavior change for our existing users (I have tested an upgrade locally and no restart).

## Local Testing

Have run all local tests.


docs/api/rabbitmq.com.ref.asciidoc
```diff
@@ -298,7 +298,7 @@ Spec is the desired state of the RabbitmqCluster Custom Resource.
 | Field | Description
 | *`replicas`* __integer__ | Replicas is the number of nodes in the RabbitMQ cluster. Each node is deployed as a Replica in a StatefulSet. Only 1, 3, 5 replicas clusters are tested. This value should be an odd number to ensure the resultant cluster can establish exactly one quorum of nodes in the event of a fragmenting network partition.
 | *`image`* __string__ | Image is the name of the RabbitMQ docker image to use for RabbitMQ nodes in the RabbitmqCluster. Must be provided together with ImagePullSecrets in order to use an image in a private registry.
-| *`imagePullSecrets`* __link:https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#localobjectreference-v1-core[$$LocalObjectReference$$] array__ | List of Secret resource containing access credentials to the registry for the RabbitMQ image. Required if the docker registry is private.
+| *`imagePullSecrets`* __link:https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#localobjectreference-v1-core[$$LocalObjectReference$$]__ | List of Secret resource containing access credentials to the registry for the RabbitMQ image. Required if the docker registry is private.
```
coro, line 301: This is an odd change. Do you know why this has become a single object rather than a list? (In the API it seems to have not changed, so wondering why it's being rendered as such)

coro (approved):
One question, otherwise LGTM

docs/api/rabbitmq.com.ref.asciidoc
ChunyiLyu: Looks to me a problem with the api generation tool we are using. I will uncommit this change

PR #729, opened 2021-06-11
Add additional printing columns

### LINKED ISSUE #647: Add printing columns to display relevant status information
The output of `kubectl get rabbitmqcluster` doesn't provide much information, other than the existence of the object itself and its age. We should expose relevant status information here, as this output is more concise and convenient than the verbose `kubectl describe`.

Add [printing columns](https://book.kubebuilder.io/reference/generating-crd.html#additional-printer-columns) to display the status conditions `AllReplicasReady` and `ReconcileSuccess`. These two conditions provide information as "are all RabbitMQ nodes ready?" and "does this cluster have up to date configuration?".

Exposing all status conditions as columns. I'm not sure it's interesting to expose `ClusterAvailable` because most of the time, as a user, one would want to have all nodes in a good state. The `ClusterAvailable` condition is set to true as long as there is one node ready.

Exposing `NoWarnings` would surface the presence of a not-recomended configuration e.g. memory requests mismatching memory limits. This might be interesting to surface, but at the moment it only checks for the mentioned memory mismatch, which doesn't happen often I believe. In the future, if we add more checks, it could be interesting to expose.

**Additional context**
Follow up from #646

coro:
This closes #647 

## Summary Of Changes

This adds two columns to the output of `kubectl get` for `RabbitmqClusters`: 

- AllReplicasReady
- ReconcileSuccess

These correspond to the equivalent fields in the Status of the object in question.

## Additional Context

For example, when creating a 3-node cluster:
image

Once the deployment is complete:
image

After applying an update to the cluster:
image

After applying an update that fails to reconcile (in this case, some override config that was incorrect):
image

I considered adding system tests for this feature, however decided against it as I believe these tests would be too brittle to changes to kubectl itself, and feels outside the remit of testing for this repo.

## Local Testing

```
```

```
$ kubectx dev-bunny
$ make destroy deploy-dev
$ make system-tests
```

### DIFF
--- a/api/v1beta1/rabbitmqcluster_types.go
+++ b/api/v1beta1/rabbitmqcluster_types.go
@@ -21,6 +21,8 @@ import (
 
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="AllReplicasReady",type="string",JSONPath=".status.conditions[?(@.type == 'AllReplicasReady')].status"
+// +kubebuilder:printcolumn:name="ReconcileSuccess",type="string",JSONPath=".status.conditions[?(@.type == 'ReconcileSuccess')].status"
 // +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
 // +kubebuilder:resource:shortName={"rmq"},categories=all
 // RabbitmqCluster is the Schema for the RabbitmqCluster API. Each instance of this object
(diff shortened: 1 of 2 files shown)


ChunyiLyu (approved):
WOW! nice feature 🚢

ChunyiLyu:
I think @Zerpet would be very happy about this PR 😄  I remember him talking about adding print columns before.

ansd (approved):

ablease:
This is great, thanks connor! :)

ablease (approved):

PR #731, opened 2021-06-14
Use RabbitMQ user for init container instead of root

### LINKED ISSUE #357: Init container runs as root
Creating a `RabbitmqCluster` in a cluster with Pod Security Policies enabled that disable to run as root result in the Pods
not being created because the init container tries to run as root.

Snippet of the events in the StatefulSet:

```
  Warning  FailedCreate  3s (x12 over 13s)  statefulset-controller  create Pod bunny-rabbitmq-server-0 in StatefulSet bunny-rabbitmq-server failed error: pods "bunny-rabbitmq-server-0" is forbidden: unable to validate against any pod security policy: [spec.initContainers[0].securityContext.runAsUser: Invalid value: 0: running with the root UID is forbidden spec.initContainers[0].securityContext.capabilities.add: Invalid value: "CHOWN": capability may not be added spec.initContainers[0].securityContext.capabilities.add: Invalid value: "FOWNER": capability may not be added]
```

We should verify if there is any requirements to run as root in the init container and adapt the security context
accordingly. If it is strictly necessary to run as root in the init container, we should document this.

This was tested with Cluster Operator 0.46.0.

coro:
This closes #357 

## Summary Of Changes
This changes the init container from running as root to running as the same specified RabbitMQ user - 999. This allows RabbitmqClusters to be deployed on environments with restrictive security policies that do not allow for root containers, while still retaining support for Openshift.

## Additional Context
RabbitMQ requires group ownership of the Mnesia database, which is mounted as a persistent volume.

The root init container was added in #303, as a means of changing the GID of this volume on Openshift - in Openshift, we observed that the GID was not being modified by `StatefulSet.Spec.SecurityContext.FSGroup`, and so the volume had the wrong owner. We added the init container to set the GID of this volume correctly on all systems.

The default set of capabilities on Openshift's container runtime, CRI-O, as well as on containerd & Docker, includes `SETGID`, so one shouldn't need to run as root to perform these actions. In testing, the initContainer seems to handle running as the RabbitMQ user `999`, and correctly sets the permissions & owner of the Mnesia volume on all three container runtimes (CRI-O tested on Openshift, containerd & Docker tested on GKE).

I'm not sure if something has changed to make this work, if we falsely assumed we needed root for this in the past, or if CRC (where we previously tested Openshift) uses a different default set of capabilities.

## Local Testing

```
```

```
$ kubectx dev-bunny
$ make destroy deploy-dev
$ make system-tests
```

### DIFF
--- a/internal/resource/statefulset.go
+++ b/internal/resource/statefulset.go
@@ -540,23 +540,8 @@ func (builder *StatefulSetBuilder) podTemplateSpec(previousPodAnnotations map[st
 					Name:  "setup-container",
 					Image: builder.Instance.Spec.Image,
 					SecurityContext: &corev1.SecurityContext{
-						RunAsUser: pointer.Int64Ptr(0),
-						Capabilities: &corev1.Capabilities{
-							// drop default set from Docker except for CHOWN, FOWNER, and DAC_OVERRIDE
-							Drop: []corev1.Capability{
-								"FSETID",
-								"KILL",
-								"SETGID",
-								"SETUID",
-								"SETPCAP",
-								"NET_BIND_SERVICE",
-								"NET_RAW",
-								"SYS_CHROOT",
-								"MKNOD",
-								"AUDIT_WRITE",
-								"SETFCAP",
-							},
-						},
+						RunAsGroup: &rabbitmqGID,
+						RunAsUser:  &rabbitmqUID,
 					},
 					Command: []string{
 						"sh", "-c", "cp /tmp/erlang-cookie-secret/.erlang.cookie /var/lib/rabbitmq/.erlang.cookie " +
--- a/internal/resource/statefulset_test.go
+++ b/internal/resource/statefulset_test.go
@@ -1217,26 +1217,13 @@ var _ = Describe("StatefulSet", func() {
 			initContainers := statefulSet.Spec.Template.Spec.InitContainers
 			Expect(initContainers).To(HaveLen(1))
 
+			rmqGID, rmqUID := int64(999), int64(999)
 			initContainer := extractContainer(initContainers, "setup-container")
 			Expect(initContainer).To(MatchFields(IgnoreExtras, Fields{
 				"Image": Equal("rabbitmq-image-from-cr"),
 				"SecurityContext": PointTo(MatchFields(IgnoreExtras, Fields{
-					"Capabilities": PointTo(MatchAllFields(Fields{
-						"Drop": ConsistOf([]corev1.Capability{
-							"FSETID",
-							"KILL",
-							"SETGID",
-							"SETUID",
-							"SETPCAP",
-							"NET_BIND_SERVICE",
-							"NET_RAW",
-							"SYS_CHROOT",
-							"MKNOD",
-							"AUDIT_WRITE",
-							"SETFCAP",
-						}),
-						"Add": BeEmpty(),
-					})),
+					"RunAsUser":  Equal(&rmqUID),
+					"RunAsGroup": Equal(&rmqGID),
 				})),
 				"Command": ConsistOf(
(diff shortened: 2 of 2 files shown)


Zerpet (approved):

ansd:
Thanks @coro.

With this change, we can omit section 3 (Create a Security Context Constraint that allows the RabbitMQ pod to have the capabilities FOWNER and CHOWN) and section 5 (Assign Security Context Constraint to Service Account) described in https://www.rabbitmq.com/kubernetes/operator/install-operator.html#openshift, correct?

>I'm not sure if something has changed to make this work, if we falsely assumed we needed root for this in the past, or if CRC (where we previously tested Openshift) uses a different default set of capabilities.

I just tested this change with the latest CRC and the init container can't change ownership of `/var/lib/rabbitmq/mnesia/` anymore:
```zsh
> crc version
CodeReady Containers version: 1.28.0+08de64bd
OpenShift version: 4.7.13 (not embedded in executable)

> k get pods
NAME          READY   STATUS             RESTARTS   AGE
r1-server-0   0/1     CrashLoopBackOff   7          13m

> k logs r1-server-0 -c setup-container
chown: changing ownership of '/var/lib/rabbitmq/mnesia/': Operation not permitted

> k logs r1-server-0 -c rabbitmq
Configuring logger redirection
10:49:58.466 [warning] Failed to write PID file "/var/lib/rabbitmq/mnesia/rabbit@r1-server-0.r1-nodes.david.pid": permission denied
10:49:58.613 [error] Failed to create Ra data directory at '/var/lib/rabbitmq/mnesia/rabbit@r1-server-0.r1-nodes.david/quorum/rabbit@r1-server-0.r1-nodes.david', file system operation error: eacces
10:49:58.614 [error] Supervisor ra_sup had child ra_system_sup started with ra_system_sup:start_link() at undefined exit with reason {error,"Ra could not create its data directory. See the log for details."} in context start_error
10:49:58.614 [error] CRASH REPORT Process <0.248.0> with 0 neighbours exited with reason: {error,"Ra could not create its data directory. See the log for details."} in ra_system_sup:init/1 line 43
10:49:58.614 [error] CRASH REPORT Process <0.242.0> with 0 neighbours exited with reason: {{shutdown,{failed_to_start_child,ra_system_sup,{error,"Ra could not create its data directory. See the log for details."}}},{ra_app,start,[normal,[]]}} in application_master:init/4 line 142
{"Kernel pid terminated",application_controller,"{application_start_failure,ra,{{shutdown,{failed_to_start_child,ra_system_sup,{error,\"Ra could not create its data directory. See the log for details.\"}}},{ra_app,start,[normal,[]]}}}"}
Kernel pid terminated (application_controller) ({application_start_failure,ra,{{shutdown,{failed_to_start_child,ra_system_sup,{error,"Ra could not create its data directory. See the log for details."}

Crash dump is being written to: erl_crash.dump...%
```

The K8s node of CRC uses cri-o container runtime:
```
  Container Runtime Version:                  cri-o://1.20.2-12.rhaos4.7.git9f7be76.el8
```
Are you getting the same behaviour on CRC?
If this issue applies only to CRC, and not to any other K8s runtime, I'm happy with this change.

coro:
I just tested this with @ansd . We're observing that `fsGroup` is successfully applied on a real Openshift cluster hosted in GCP, meaning that we don't need the root user in the init container. This is likely because the CRC volume driver does not support the `fsGroup` flag, whereas the one used in the real cluster does.

We also noticed that the chown actions done by the initContainer were redundant on systems where the storage driver supports `fsGroup`.

Another thing we noticed is that the rabbitmq user `999` is not actually strictly necessary for the container to run. Replacing this number with an arbitrary value in `runAsUser` and `fsGroup` in the security contexts of the containers still led to the cluster being successfully created.

ansd (approved):

PR #739, opened 2021-06-24
Remove chmod operations from init Container

coro:
This is no longer needed as fsGroup modifies this for us.

## Summary Of Changes

`fsGroup`, set in the securityContext of mounted volumes, already sets the ownership of files within the volumes, so explicitly setting them in the initContainer is unnecessary. We also no longer run as GID 999, so part of the chown operation is actively doing the wrong thing now.

## Additional Context
This brings the operator closer to supporting Openshift's arbitrary user flow. 

Following this PR will be updates to the docs around installing on Openshift.

## Local Testing

```
```

```
$ kubectx dev-bunny
$ make destroy deploy-dev
$ make system-tests
```


ChunyiLyu (approved):
Love it！

ansd (approved):

PR #749, opened 2021-06-29
Allow SecurityContext and PodSecurityContext to be reset to default

### LINKED ISSUE #741: Override does not patch with null values
## Describe the bug

Create a RMQ with manifest like so:

```yaml
kind: RabbitmqCluster
metadata:
  name: sample
  namespace: rabbitmq-system
spec:
  image: rabbitmq:3.8.17-management
  override:
    statefulSet:
      spec:
        template:
          spec:
            securityContext: null
            containers: []
            initContainers:
            - name: setup-container
               securityContext: null
```

**Expected behavior**

Deployed statefulSet pod spec should have securityContext set to `null` or not set (empty). However, both `securityContext: null` is  omitted after the manifest is saved (if you get the RMQ manifest after creation, you don't see `securityContext: null` in the manifest).

Setting fields to `null` is a documented way of deleting a field for [strategic merge patch ](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-api-machinery/strategic-merge-patch.md#maps). When applying the same patch but using `kubectl patch` directly, securityContext can be set to null succesfully. 

Example patch:
```yaml
spec:
 template:
   spec:
     securityContext: null
     initContainers:
     - name: setup-container
       securityContext: null
```
applied by running `k patch sts sample-server --patch "$(cat patch.yaml)"`

- RabbitMQ: 3.8.17
- RabbitMQ Cluster Operator: latest main
- Kubernetes: [e.g. 1.18.0]
- Cloud provider or hardware configuration:

coro:
This closes #741 and closes #746 

## Summary Of Changes
If an override is provided with the SecurityContext of either the rabbitmq container or init container set to `{}`, this ensures the resultant STS uses the default securityContext of the container runtime.

## Additional Context
Unlike #746, this preserves the validation of override types, rather than failing at reconcile time if a user provides an unknown field.

This only works for securityContext right now; the only other fields I think this affects are:
`topologySpreadConstraints`
`resources`

I don't think they're a high priority to make this work for, but I could add them in if needed.

## Local Testing

```
```

```
$ kubectx dev-bunny
$ make destroy deploy-dev
$ make system-tests
```

### DIFF
--- a/controllers/rabbitmqcluster_controller_test.go
+++ b/controllers/rabbitmqcluster_controller_test.go
@@ -959,7 +959,7 @@ var _ = Describe("RabbitmqClusterController", func() {
 			}, 3).Should(Equal("my-great-image-2"))
 		})
 
-		It("can reset a controller-set field to default", func() {
+		It("can reset the securityContext of the RabbitMQ container and initContainer to default", func() {
 			Expect(updateWithRetry(cluster, func(r *rabbitmqv1beta1.RabbitmqCluster) {
 				cluster.Spec.Override.StatefulSet.Spec.Template.Spec.SecurityContext = &corev1.PodSecurityContext{}
 				cluster.Spec.Override.StatefulSet.Spec.Template.Spec.InitContainers = []corev1.Container{
--- a/controllers/rabbitmqcluster_controller_test.go
+++ b/controllers/rabbitmqcluster_controller_test.go
@@ -22,6 +22,7 @@ import (
 	. "github.com/onsi/ginkgo"
 
 	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gstruct"
 	rabbitmqv1beta1 "github.com/rabbitmq/cluster-operator/api/v1beta1"
 	"github.com/rabbitmq/cluster-operator/internal/status"
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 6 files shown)


controllers/rabbitmqcluster_controller_test.go
```diff
@@ -957,6 +958,28 @@ var _ = Describe("RabbitmqClusterController", func() {
 				return c.Image
 			}, 3).Should(Equal("my-great-image-2"))
 		})
+
+		It("can reset a controller-set field to default", func() {
```
ChunyiLyu, line 962: Should we make the test description more specific here? Our controller is only treating pod and container SecurityContext as special cases, and it currently does not handle any other fields

internal/resource/statefulset.go
```diff
@@ -258,6 +259,16 @@ func patchPodSpec(podSpec, podSpecOverride *corev1.PodSpec) (corev1.PodSpec, err
 		sortVolumeMounts(patchedPodSpec.Containers[0].VolumeMounts)
 	}
 
+	if podSpecOverride.SecurityContext != nil && reflect.DeepEqual(*podSpecOverride.SecurityContext, corev1.PodSecurityContext{}) {
```
ChunyiLyu, line 262: I would suggest adding comments here with context about why we are handing securityContext as a special case. I can see myself getting confused in the future.

ChunyiLyu (approved):
LGTM. Just some minor suggestions on comments and tests descriptions.

ablease (approved):
LGTM. Nothing to add to Chunyi's suggestions. 👍

ChunyiLyu (approved):

PR #750, opened 2021-06-30
Add YTT overlay to automate removing default operator securityContext

coro:

## Summary Of Changes
Adds a YTT overlay in the `hack/` directory to remove the securityContext on the operator Deployments (both `cluster-operator` and `messaging-topology-operator`).

This allows you to simply use the overlay to remove this field, rather than having to manually download the manifest, edit it then kubectl apply it.

## Additional Context

Docs PR is here: https://github.com/rabbitmq/rabbitmq-website/pull/1220

## Local Testing

```
```

```
$ kubectx dev-bunny
$ make destroy deploy-dev
$ make system-tests
```

### DIFF
--- a/hack/remove-operator-securityContext.yml
+++ b/hack/remove-operator-securityContext.yml
@@ -1,0 +1,8 @@
+#@ load("@ytt:overlay", "overlay")
+#@overlay/match by=overlay.subset({"kind": "Deployment"}), expects="1+"
+---
+spec:
+  template:
+    spec:
+      #@overlay/remove
+      securityContext:


MirahImage:
Would it make sense to provide an Openshift example as well?

coro:
> Would it make sense to provide an Openshift example as well?

Are you thinking of something distinct from https://github.com/rabbitmq/cluster-operator/tree/main/docs/examples/default-security-context ? I'm also writing openshift-specific docs on the rabbitmq.com website alongside this.

Zerpet (approved):
Let Carvel take over!

coro:
Do I have good news: I don't think this is even necessary. The operator already runs as UID/GID 1000 thanks to the Dockerfile, by default. The operator doesn't mount any volumes, so fsGroup does nothing. We can actually just straight-up remove the securityContext altogether, I think. I will test this now.

coro:
Superseded by https://github.com/rabbitmq/cluster-operator/pull/751

PR #751, opened 2021-06-30
Remove securityContext from manager Deployment

coro:
## Summary Of Changes
Removing unnecessary securityContext from manager Pod:
- runAsUser & runAsGroup are already enabled by the `USER` line of the
Dockerfile
- fsGroup is unneeded since the manager mounts no volumes.

The securityContext field overrides some of Openshift's arbitrary user
handling, while providing no benefit to the operator. Removing it
provides the same security guarantees as before, while also making
Openshift installation experience the same as any other Kubernetes
distribution.

## Local Testing

```
```

```
$ kubectx dev-bunny
$ make destroy deploy-dev
$ make system-tests
```

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -30,10 +30,6 @@ spec:
         app.kubernetes.io/part-of: rabbitmq
     spec:
       serviceAccountName: rabbitmq-cluster-operator
-      securityContext:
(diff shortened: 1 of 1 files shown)


MirahImage (approved):

ChunyiLyu (approved):
👍

PR #755, opened 2021-07-02
Add host and port to default-user secret.

MirahImage:
This provides additional information in the context of the Provisioned
Service Binding spec https://k8s-service-bindings.github.io/spec/#provisioned-service

When TLS is configured, we provide the AMQPS port, otherwise we provide
the AMQP port.

The host is the dns name of the service.

When additional plugins supporting other protocols are enabled, ports for those protocols are also added to the secret.


docs/api/rabbitmq.com.ref.asciidoc
```diff
@@ -298,7 +298,7 @@ Spec is the desired state of the RabbitmqCluster Custom Resource.
 | Field | Description
 | *`replicas`* __integer__ | Replicas is the number of nodes in the RabbitMQ cluster. Each node is deployed as a Replica in a StatefulSet. Only 1, 3, 5 replicas clusters are tested. This value should be an odd number to ensure the resultant cluster can establish exactly one quorum of nodes in the event of a fragmenting network partition.
 | *`image`* __string__ | Image is the name of the RabbitMQ docker image to use for RabbitMQ nodes in the RabbitmqCluster. Must be provided together with ImagePullSecrets in order to use an image in a private registry.
-| *`imagePullSecrets`* __link:https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#localobjectreference-v1-core[$$LocalObjectReference$$] array__ | List of Secret resource containing access credentials to the registry for the RabbitMQ image. Required if the docker registry is private.
+| *`imagePullSecrets`* __link:https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#localobjectreference-v1-core[$$LocalObjectReference$$]__ | List of Secret resource containing access credentials to the registry for the RabbitMQ image. Required if the docker registry is private.
```
ChunyiLyu, line 301: I think `imagePullSecrets` is still an array here? Would you mind checking this change out?

internal/resource/default_user_secret.go
```diff
@@ -87,6 +94,61 @@ func (builder *DefaultUserSecretBuilder) Update(object client.Object) error {
...
+	const (
+		AMQPPort  = "5672"
+		AMQPSPort = "5671"
+	)
+	portNames := map[v1beta1.Plugin]string{
+		"rabbitmq_mqtt":      "mqtt-port",
+		"rabbitmq_stomp":     "stomp-port",
+		"rabbitmq_stream":    "stream-port",
+		"rabbitmq_web_mqtt":  "web-mqtt-port",
+		"rabbitmq_web_stomp": "web-stomp-port",
```
ChunyiLyu, line 107: From an application perspective, does it ever need the management port?

ChunyiLyu (changes requested):
Did some manual testing. Plugin ports are added successfully when the plugin is enabled at creation. However, when plugins are enabled after the cluster is already created, the service binding secret is not updated newly added plugin port.

internal/resource/default_user_secret.go
MirahImage: I don't think management should be needed

docs/api/rabbitmq.com.ref.asciidoc
MirahImage: Yeah, this was generated, happy to change it back.

ChunyiLyu (approved):
Update works now! Thank you :)

PR #757, opened 2021-07-02
Add new podAntiAffinity example and modify existing production-ready example

MrPsycho5:
This PR adds `podAntiAffinity` example and modifies the existing procudtion-ready example so that the pods will not be scheduled on the same node. This should ensure higher availability in any Kubernetes cluster.

The pod-anti-affinity example uses `preferredDuringSchedulingIgnoredDuringExecution` type, while production-ready example uses `requiredDuringSchedulingIgnoredDuringExecution` as no production cluster should have potential one point of failure (i.e. one node).

### DIFF
--- a/docs/examples/production-ready/rabbitmq.yaml
+++ b/docs/examples/production-ready/rabbitmq.yaml
@@ -20,6 +20,18 @@ spec:
   persistence:
     storageClassName: ssd
     storage: "500Gi"
+  affinity:
+    podAntiAffinity:
+      requiredDuringSchedulingIgnoredDuringExecution:
+      - weight: 100
+        podAffinityTerm:
+          labelSelector:
+            matchExpressions:
+              - key: app.kubernetes.io/name
+                operator: In
+                values:
+                - production-ready
+          topologyKey: kubernetes.io/hostname
   override:
     statefulSet:
       spec:
--- a/docs/examples/production-ready/rabbitmq.yaml
+++ b/docs/examples/production-ready/rabbitmq.yaml
@@ -23,15 +23,13 @@ spec:
   affinity:
     podAntiAffinity:
       requiredDuringSchedulingIgnoredDuringExecution:
-      - weight: 100
-        podAffinityTerm:
-          labelSelector:
-            matchExpressions:
-              - key: app.kubernetes.io/name
-                operator: In
-                values:
-                - production-ready
-          topologyKey: kubernetes.io/hostname
+      - labelSelector:
+          matchExpressions:
+            - key: app.kubernetes.io/name
+              operator: In
+              values:
+              - production-ready
+        topologyKey: kubernetes.io/hostname
   override:
     statefulSet:
       spec:
--- a/docs/examples/pod-anti-affinity/README.md
+++ b/docs/examples/pod-anti-affinity/README.md
@@ -1,0 +1,17 @@
+# Pod Anti Affinity Example
+
+This simple pod anti affinity rule will try to assign pods to different nodes according to [Assigning Pods to Nodes](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/). 
+
+The `preferredDuringSchedulingIgnoredDuringExecution` type can be replaced by `requiredDuringSchedulingIgnoredDuringExecution` to require pod separation as in production-ready example.
+
+You can deploy this example like this:
+
+```shell
+kubectl apply -f rabbitmq.yaml
+```
+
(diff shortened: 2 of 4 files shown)


MirahImage (approved):
Thanks for your contribution!

mkuratczyk:
@MrPsycho5 can you explain why the changes to the production example are necessary? It already includes `topologySpreadConstraint` to make sure RabbitMQ nodes run in different AZs, which by extension also requires them to run on different Kubernetes nodes. Unless I'm missing something we should use one or the other but not both because:
1. users with multiple AZs don't need this additional anti-affinity rule
2. users with a single AZ need to remove the `topologySpreadConstratint` anyway (otherwise the pods can't be scheduled)

Thanks,

MrPsycho5:
I have the following situations in my mind:

1. User has 3 availability zones, 5+ nodes and 5 rabbit replicas. We should not only make sure the replicas are spread across different availability zones, but also different nodes within single zone. 
Perhaps we should also change `DoNotSchedule` to `ScheduleAnyway` to allow for such situation.

2. If the user only has single availability zone (in which case he indeed needs to remove the `topologySpreadConstratint`) we should at least keep the good practice and require the spread of replicas on different nodes.

docs/examples/production-ready/rabbitmq.yaml
```diff
@@ -20,6 +20,18 @@ spec:
...
+      requiredDuringSchedulingIgnoredDuringExecution:
+      - weight: 100
+        podAffinityTerm:
+          labelSelector:
+            matchExpressions:
+              - key: app.kubernetes.io/name
+                operator: In
+                values:
+                - production-ready
+          topologyKey: kubernetes.io/hostname
```
MirahImage, line 34: The `topologyKey` is indented one level too far, causing the validation to fail if this example is deployed.

MirahImage (changes requested):

docs/examples/production-ready/rabbitmq.yaml
MrPsycho5: Can you please confirm the issue? According to my Kubernetes cluster (v1.20.8) both `labelSelector` and `topologyKey` should be on the same level (i.e. under `requiredDuringSchedulingIgnoredDuringExecution`). In the file I commited both `labelSelector` and `topologyKey` are indented by 10 spaces.

If you want to confirm that on your cluster run for example `kubectl explain deployment.spec.template.spec.affinity.podAntiAffinity.requiredDuringSchedulingIgnoredDuringExecution`

mkuratczyk: Something is definitely wrong:
```
$ kubectl apply -f https://raw.githubusercontent.com/rabbitmq/cluster-operator/15b3d20bf924b02a68d4a9bda7afa9d7f84b9271/docs/examples/production-ready/rabbitmq.yaml                    ⎈ kind-kind/examples
error: error validating "https://raw.githubusercontent.com/rabbitmq/cluster-operator/15b3d20bf924b02a68d4a9bda7afa9d7f84b9271/docs/examples/production-ready/rabbitmq.yaml": error validating data: [ValidationError(RabbitmqCluster.spec.affinity.podAntiAffinity.requiredDuringSchedulingIgnoredDuringExecution[0]): unknown field "podAffinityTerm" in com.rabbitmq.v1beta1.RabbitmqCluster.spec.affinity.podAntiAffinity.requiredDuringSchedulingIgnoredDuringExecution, ValidationError(RabbitmqCluster.spec.affinity.podAntiAffinity.requiredDuringSchedulingIgnoredDuringExecution[0]): unknown field "weight" in com.rabbitmq.v1beta1.RabbitmqCluster.spec.affinity.podAntiAffinity.requiredDuringSchedulingIgnoredDuringExecution, ValidationError(RabbitmqCluster.spec.affinity.podAntiAffinity.requiredDuringSchedulingIgnoredDuringExecution[0]): missing required field "topologyKey" in com.rabbitmq.v1beta1.RabbitmqCluster.spec.affinity.podAntiAffinity.requiredDuringSchedulingIgnoredDuringExecution]; if you choose to ignore these errors, turn validation off with --validate=false
```

MrPsycho5: Turns out changing from `preferred` to `required` isn't just some cosmetics. Obviously weight does not play a role when we require something. I have commited a fix and I can apply it on v1.20.8.

MirahImage (approved):

mkuratczyk (approved):

mkuratczyk:
Dzięki Jakub! ;)

Issue #75, opened 2020-04-06
Hard-coded cluster domain prevents pods from starting on K8s clusters using alternative domains


st3v:
**Describe the bug**
At the moment, our operator hard-codes the cluster domain in the peer-discovery-related env vars of the pod template to `cluster.local`. I.e.

```yaml
apiVersion: apps/v1
kind: StatefulSet
metadata:
  ...
spec:
  ...
  template:
    spec:
      ...
      containers:
      - env:
        ...
        - name: RABBITMQ_NODENAME
          value: rabbit@$(MY_POD_NAME).$(K8S_SERVICE_NAME).$(MY_POD_NAMESPACE).svc.cluster.local
        - name: K8S_HOSTNAME_SUFFIX
          value: .$(K8S_SERVICE_NAME).$(MY_POD_NAMESPACE).svc.cluster.local
        ...
```

Now while `cluster.local` is the most common domain used for K8s clusters, it is configurable by the cluster admin and there are clusters that use alternative domains. One such example are guest clusters running in Project Pacific. They use `managedcluster.local` as their cluster domain.

```yaml
$ k -n kube-system get configmap coredns -oyaml
kind: ConfigMap
apiVersion: v1
data:
  Corefile: |
    .:53 {
        errors
        health
        kubernetes managedcluster.local in-addr.arpa ip6.arpa {
           pods insecure
           upstream
           fallthrough in-addr.arpa ip6.arpa
           ttl 30
        }
...
```

If DNS for the K8s cluster is configured with such an alternative domain, the pods in our `StatefulSet` are failing to start up with the following log message ...

```
$ k get pod
NAME                                         READY   STATUS             RESTARTS   AGE
on-demand-claim-rabbitmq-server-0            0/1     CrashLoopBackOff   8          17m

$ k logs on-demand-claim-rabbitmq-server-0
ERROR: epmd error for host on-demand-claim-rabbitmq-server-0.on-demand-claim-rabbitmq-headless.default.svc.cluster.local: nxdomain (non-existing domain)
```

**To Reproduce**
1. Get ahold of a cluster with a cluster domain set to something else than `cluster.local`
2. Create a `RabbitmqCluster`
3. See the first pod failing to startup with the error message shown above

**Expected behavior**
Two alternatives. 
  1. Deploy `RabbitmqCluster` without changing anything and it just works despite the cluster domain not being `cluster.local`

  **OR**

[...]

PR #778, opened 2021-07-21
Remove `.cluster.local` from host in default-user secret.

MirahImage:
Cluster name is configurable and not fixed.

This is related to #75 and [Messaging Topology Operator issue 186](https://github.com/rabbitmq/messaging-topology-operator/issues/186)

## Summary Of Changes

Removes hardcoded `.cluster.local` suffix from the service address in the `host` field of the default-user secret.


ChunyiLyu (approved):
Great!

PR #784, opened 2021-07-29
Remove TSMgr-specific content

mkuratczyk:
Since TSMgr has reached the end of general support, we can remove this content.


ChunyiLyu:
@mkuratczyk should this comment be removed as well? https://github.com/rabbitmq/cluster-operator/blob/main/charts/rabbitmq/test.sh#L18 It mentions `plans` so I assume it's TSM related.

ChunyiLyu (approved):
@mkuratczyk should this comment be removed as well? https://github.com/rabbitmq/cluster-operator/blob/main/charts/rabbitmq/test.sh#L18 It mentions plans so I assume it's TSM related. Everything else looks good

PR #785, opened 2021-07-29
Namespace access

njbarber:
This closes https://github.com/rabbitmq/messaging-topology-operator/issues/147

## Summary Of Changes

Effectively just adds a new `messagingTopologyNamespaces` field.

## Additional Context

The field is used by the messaging topology operator when creating resources to determine if they have access to given cluster.

Related to https://github.com/rabbitmq/messaging-topology-operator/pull/201.

## Local Testing

```
```

```
$ kubectx dev-bunny
$ make destroy deploy-dev
$ make system-tests
```


config/crd/bases/rabbitmq.com_rabbitmqclusters.yaml
```diff
@@ -527,6 +527,11 @@ spec:
                         type: string
                     type: object
                   type: array
+                messagingTopologyNamespaces:
+                  description: The list of namespaces to allow messaging topology resources to be created from. If one is not defined messaging topology resources can only be created in the same namespace as the cluster.
```
michaelklishin, line 531: Is this meant to say "topology resources to be created in"?

michaelklishin, line 531: I find this to be fairly confusing to those who do not have any background on this. That said, I don't have a much better wording to offer. To me this suggests it may be a difficult feature to explain (or a highly unusual one?)

michaelklishin:
So while previously any namespace could contribute topology definitions, now all of them have to be explicitly whitelisted.
This sounds like a significant change in terms of usability.

Zerpet (changes requested):
Hello @njbarber,

Thanks for contributing to this project! I'm afraid we can't accept the PR with the current API bump, as it does not adhere to our compatibility commitments, documented [here](https://github.com/rabbitmq/cluster-operator/blob/main/docs/proposals/20200806-versioning-release-strategy.md#api-versions). Specifically, this change does not comply with rule 4a of [Kubernetes API deprecation policy](https://kubernetes.io/docs/reference/using-api/deprecation-policy/). This link can be found in our compatibility commitments, shared in the first link.


>Rule #4a: Other than the most recent API versions in each track, older API versions must be supported after their announced deprecation for a duration of no less than:
>
>   - GA: 12 months or 3 releases (whichever is longer)
>   - Beta: 9 months or 3 releases (whichever is longer)
>   - Alpha: 0 releases

Would you share the reasoning to bump the API version as part of this PR?

I'm happy to explore potential solutions to the problem described in https://github.com/rabbitmq/messaging-topology-operator/issues/147. Given that this issue involves multiple repos, let's move the conversation about how to tackle this problem to the GitHub issue.

Any conversation related to the code in this PR, I'm happy to keep it here.

njbarber:
@michaelklishin 

> So while previously any namespace could contribute topology definitions, now all of them have to be explicitly whitelisted.
This sounds like a significant change in terms of usability.

Not quite. Only if a cluster is planning on accepting messaging topology resources from another namespace does it need to define `messagingTopologyNamespaces`. Additionally if a messaging topology resource defines a `rabbitmqClusterReference` without namespace the operator assumes the cluster is in the same namespace as the resource (which is identical to the previous behavior).

njbarber:
@Zerpet my understanding is that spec changes require an API version bump. I'd love to know if that's not necessarily always the case (as I had to a lot of find and replace that wasn't particularly fun).

Zerpet:
> @Zerpet my understanding is that spec changes require an API version bump. I'd love to know if that's not necessarily always the case (as I had to do a lot of find and replace that wasn't particularly fun).

If you add an **optional** field to the spec, you don't need to bump the API version. Adding an optional field keeps the API compatible with previous versions, since older clients will still work, even if they don't send the new field in the request.

We refer to [this guide](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#on-compatibility) on API changes, from SIG Architecture.

njbarber:
@Zerpet

> If you add an optional field to the spec, you don't need to bump the API version.

I did not know that.

In that case I believe both sets of changes do not require a version bump.

https://github.com/njbarber/messaging-topology-operator/blob/namespace-access/internal/cluster_reference.go#L32-L43 will never be executed for existing messaging topology resources that have a cluster reference, for instance (meaning `messagingTopologyNamespaces` isn't required on the cluster spec).

njbarber:
This is no longer relevant due to recent changes on https://github.com/rabbitmq/messaging-topology-operator/pull/201.

PR #786, opened 2021-07-30
Don't create PVC if spec.persistence.storage==0

### LINKED ISSUE #776: Allow creating clusters without persistence
I would like to use the operator to create disposable RabbitMQ clusters without persistence enabled.

`RabbitmqClusterPersistenceSpec` be controlled by a boolean `persistence: bool` which will allow not creating any persistent volume.

mkuratczyk:
For test environments, especially in CI/CD, it may be beneficial to skip
creating the PVC altogether since the whole cluster is disposable
anyway. With this change, if spec.persistence.storage is set to 0, the
PVC is not created and an `emptyDir` volume is used instead. Changing from
0 to anything else is not supported (and doesn't make much sense
anyway).

This closes #776

### DIFF
--- a/controllers/reconcile_no_persistence_test.go
+++ b/controllers/reconcile_no_persistence_test.go
@@ -1,0 +1,95 @@
+package controllers_test
+
+import (
+	"context"
+	"fmt"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	rabbitmqv1beta1 "github.com/rabbitmq/cluster-operator/api/v1beta1"
+	"github.com/rabbitmq/cluster-operator/internal/status"
+	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	k8sresource "k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


ChunyiLyu (approved):
Everything works as expected. For an edge case, if someone set `persistence.storage` to `0` but defines PVCs through statefulSet override, they do not get any PVC created. Example manifest:
```yaml
kind: RabbitmqCluster
metadata:
  name: multiple-disks
spec:
  replicas: 1
  persistence:
    storage: 0
  override:
    statefulSet:
      spec:
        volumeClaimTemplates:
          - apiVersion: v1
            kind: PersistentVolumeClaim
            metadata:
              name: persistence
              namespace: default
            spec:
              accessModes:
                - ReadWriteOnce
              resources:
                requests:
                  storage: 10Gi
              volumeMode: Filesystem
```
Is this the desired behavior? If yes, may I suggest to add a unit level test for this behavior to capture in the code? We are getting this behavior because the code block of 
```
if instance.Spec.Persistence.Storage.Cmp(zero) == 0 {
		return []corev1.PersistentVolumeClaim{}, nil
	}
```
is before override was applied and would be great to add a test to ensure so we don't change the behavior by accident.

coro (approved):

ChunyiLyu (approved):

PR #828, opened 2021-09-02
Add extra logging and event detail when attempting to scale down cluster

### LINKED ISSUE #758: [BUG] Unexpected rabbitmqCluster replica number when operator misses particular scale events
## Describe the bug

The rabbitmq operator have the following checking to prevent users to scale down statefulset (and pvc) of the cluster,

```go
if r.scaleDown(ctx, rabbitmqCluster, current, sts) {
	// return when cluster scale down detected; unsupported operation
	return ctrl.Result{}, nil
}
```

If the user tries to change the replica from 3 to 2, the scaling down should not happen and the replicas should still remain 3.

We ran a workload that changes replica from 1 -> 3 -> 2. Ideally, the replicas should remain 3 since scaling down is not allowed. However, we observed there are 2 replicas at the end in some cases.

The reason is that the goroutine of performing reconciliation (g1) and the goroutine of updating the locally cached rabbitmqCluster object (g2) are concurrent, and certain interleaving of the two goroutines can lead to the unexpected scaling behavior.
Ideally, this interleaving will lead to the correct result:
1. g2: set replica number in the rabbitmq (from 1) to 3
2. g1: read the rabbitmq object and scales sts from 1 to 3
2. g2: set replica number in the rabbitmq (from 3) to 2
3. g1: read the rabbitmq object and refuse to scale from 3 to 2

**And we find that the following interleaving (when the operator's reconcile goroutine runs relatively slowly) can lead to the unexpected scaling behavior mentioned above:**
1. g2: set replica number in the rabbitmq (from 1) to 3
2. g2: set replica number in the rabbitmq (from 3) to 2
3. g1: read the rabbitmq object and scales sts from 1 to 2

## To Reproduce
As mentioned above, we change the replica number as 1 -> 3 -> 2. When the operator misses the view of "replica = 3" and directly sees "replica = 2", the rabbitmq statefulset will end up with 2 replicas, not 3.

**Expected behavior**
Since scale down is not allowed, we should have 3 replicas eventually.

- RabbitMQ: 3.8.12
- Kubernetes: 1.18.9
 
## Additional context
This issue is probably hard to fix from the operator side, as how the operator reads events from users (e.g., scaling up/down) and how the different goroutines interleave with each other are decided by the `controller-runtime` package, not the operator code. We didn't find a good way to solve this issue with the existing `controller-runtime` package.
Instead, we can at least make the potential debugging process easier when other users encounter similar problems later again by adding more logs. For example, we can log the `currentReplicas` and `desiredReplicas` in `scaleDown` so users can know the exact sequence of replica number seen by the operator.
We are willing to send a PR to add one more log here.

coro:
This closes #758

## Summary Of Changes

Adds extra detail to the warning generated when attempting to scale down the cluster. The log now states the number of replicas is marked as current, and the desired number of replicas that were used to calculate whether a scale down was occurring. This should help debug other race conditions as demonstrated in #758 .

Example status:
image

## Local Testing

```
```

```
$ kubectx dev-bunny
$ make destroy deploy-dev
$ make system-tests
```

### DIFF
--- a/controllers/reconcile_scale_down.go
+++ b/controllers/reconcile_scale_down.go
@@ -3,6 +3,8 @@ package controllers
 import (
 	"context"
 	"errors"
+	"fmt"
+
 	"github.com/go-logr/logr"
 	"github.com/rabbitmq/cluster-operator/api/v1beta1"
 	"github.com/rabbitmq/cluster-operator/internal/status"
@@ -18,7 +20,7 @@ func (r *RabbitmqClusterReconciler) scaleDown(ctx context.Context, cluster *v1be
 	currentReplicas := *current.Spec.Replicas
 	desiredReplicas := *sts.Spec.Replicas
 	if currentReplicas > desiredReplicas {
-		msg := "Cluster Scale down not supported"
+		msg := fmt.Sprintf("Cluster Scale down not supported; tried to scale cluster from %d nodes to %d nodes", currentReplicas, desiredReplicas)
 		reason := "UnsupportedOperation"
 		logger.Error(errors.New(reason), msg)
 		r.Recorder.Event(cluster, corev1.EventTypeWarning, reason, msg)
--- a/controllers/reconcile_scale_down_test.go
+++ b/controllers/reconcile_scale_down_test.go
@@ -3,6 +3,7 @@ package controllers_test
 import (
 	"context"
 	"fmt"
+
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	rabbitmqv1beta1 "github.com/rabbitmq/cluster-operator/api/v1beta1"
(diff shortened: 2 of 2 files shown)


ChunyiLyu (approved):
LGTM thanks!

ChunyiLyu:
@coro approved!

A test failure seems to be related to the change:

```
Cluster scale down
/__w/cluster-operator/cluster-operator/controllers/reconcile_scale_down_test.go:17
  does not allow cluster scale down [It]
  /__w/cluster-operator/cluster-operator/controllers/reconcile_scale_down_test.go:33

  Timed out after 5.001s.
  Expected
      <string>: "...orted; trie..."
  to equal               |
      <string>: "...orted"
```

ansd (approved):

PR #838, opened 2021-09-08
Make PVC scaling more idempotent & resilient to crashes

### LINKED ISSUE #782: [BUG] PVC expansion fails if the operator crashes in the middle of reconciliation
## Describe the bug
We find that if the operator crashes at some point in the middle of performing PVC expansion, the actual PVC size will never meet user's spec even after the operator restarts.

Specifically, The rabbitmq operator performs the following steps to expand the capacity of statefulset's PVC,
1. get cluster statefulset, and if cluster statefulset exist
   1. compare **capacity in CRD Spec** with **capacity in statefulset's `volumeClaimTemplates`** to determine whether to resize PVC
    ```
	currentCapacity, err := persistenceStorageCapacity(current.Spec.VolumeClaimTemplates)
	...
	desiredCapacity, err := persistenceStorageCapacity(desired.Spec.VolumeClaimTemplates)
	...
	cmp := currentCapacity.Cmp(desiredCapacity)
    ```
   2. delete cluster statefulset
   3. update the actual capacity of PVC
    ```
   if err := r.deleteSts(ctx, rmq); err != nil {
		return err
	}

	if err := r.updatePVC(ctx, rmq, *current.Spec.Replicas, desiredCapacity); err != nil {
		return err
	}
    ```
2. create/update cluster statefulset (update PVC capacity in statefulset's `volumeClaimTemplates`)

We ran a workload that changes PVC capacity from 10Gi to 15Gi, and observed that PVC was never resized to 15Gi if the operator crashed at some point between **deleting cluster statefulset (ii)** and **updating PVC capacity (iii)**.

In that case, the cluster statefulset was actually deleted ((ii) is executed before crash) while the actual PVC capacity is not successfully resized ((iii) is not executed before crash). After the operator restarts from the crash, it will try to get the cluster statefulset. Since the statefulset doesn't exist, the operator will **skip (i),(ii),(iii)**, and directly create the cluster statefulset with the **same PVC capacity in its `volumeClaimTemplates` as in the CRD spec**. Since the two values become the same, the PVC expansion logic will never be triggered until the user manually change the PVC capacity in the CRD spec again.

## To Reproduce
As mentioned above, resize the PVC from 10Gi to 15Gi and make the operator crash between (ii) and (iii). After restart, the PVC expansion will not succeed eventually.

**Expected behavior**
The operator should recover correctly from the crash and continue the PVC expansion.

- RabbitMQ: 3.8.12
- Kubernetes: 1.18.9
 
## Additional context
We are willing to send a PR to help fix this issue. A simple fix could be to compare **capacity in CRD Spec** the **actual PVC capacity** (instead of the capacity in statefulset's `volumeClaimTemplates`), so that even after the untimely crash the PVC expansion can still work.

coro:

## Summary Of Changes

This allows for this section of the code to have its own test suite,
without relying on the restrictions placed by envtest. This allows us to
test more scenarios, such as where a controller has previously crashed
and left the existing API objects in a bad state (e.g. statefulsets
missing, PVCs being the wrong size, etc.).

This also fixes #782, as the process of scaling PVCs should now be more
idempotent than before:
- StatefulSets don't need to be present on the cluster in order to scale
the PVCs (as long as the PVCs exist)
- PVC scaling is done on a PVC-by-PVC case, rather than all at once, so
that if one or more replicas have already scaled their PVCs it does not
block the others
- The current size of PVCs is determined by the live reading of the K8s
API, rather than what's reported in the RabbitmqCluster object.

## Local Testing

```
```

```
$ kubectx dev-bunny
$ make destroy deploy-dev
$ make system-tests
```

### DIFF
--- a/controllers/rabbitmqcluster_controller.go
+++ b/controllers/rabbitmqcluster_controller.go
@@ -170,20 +170,22 @@ func (r *RabbitmqClusterReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 				return ctrl.Result{}, err
 			}
 
-			// only checks for PVC expansion and scale down if statefulSet is created
+			// only checks for scale down if statefulSet is created
 			// else continue to CreateOrUpdate()
 			if !k8serrors.IsNotFound(err) {
 				if err := builder.Update(sts); err != nil {
-					return ctrl.Result{}, err
-				}
-				if err = r.reconcilePVC(ctx, rabbitmqCluster, current, sts); err != nil {
-					r.setReconcileSuccess(ctx, rabbitmqCluster, corev1.ConditionFalse, "FailedReconcilePVC", err.Error())
 					return ctrl.Result{}, err
 				}
 				if r.scaleDown(ctx, rabbitmqCluster, current, sts) {
 					// return when cluster scale down detected; unsupported operation
 					return ctrl.Result{}, nil
 				}
+			}
+
+			// The PVCs for the StatefulSet may require expanding
+			if err = r.reconcilePVC(ctx, rabbitmqCluster, sts); err != nil {
+				r.setReconcileSuccess(ctx, rabbitmqCluster, corev1.ConditionFalse, "FailedReconcilePVC", err.Error())
+				return ctrl.Result{}, err
 			}
 		}
 
--- a/controllers/reconcile_persistence.go
+++ b/controllers/reconcile_persistence.go
@@ -2,114 +2,26 @@ package controllers
 
 import (
 	"context"
-	"errors"
 	"fmt"
-	"time"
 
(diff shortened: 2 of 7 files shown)


MirahImage (approved):
I'll leave it up to you as to whether you want to leave the `BeforeEach` and `JustBeforeEach` in the `scaling_test_suite` or move them into the `scaling_test`.

ansd (approved):
Very good changes.

Especially the unit tests are excellent 😍 ... not only the Gomega matchers you wrote but also that you use the K8s unit test packages 🙌

@coro I pushed two commits.
Logs were polluted at Info level with large PVC structs.
I set it to Debug level. Feel free to remove that log line completely.

PR #839, opened 2021-09-09
add set-login-password-usernam sample

isdaniel:
Hi I had added a sample to descript how to set the default account and password in the rabbitMQ cluster in k8s


Zerpet (approved):
Thank you for the contribution!

PR #841, opened 2021-09-13
Prefix generated username with 'default_user_'

ChunyiLyu:
This closes #

## Summary Of Changes

- this change will make it easier to specify exclusion rules whenusing schema sync plugin
- After this change, an example username and password could look like:
```
username: default_user_mDj_Yp58qvSynGbUEC3
password: AbdFbGnu2HaVAIkJ1W1AUq42uVglXW1E
```

## Additional Context

Generated username and password logic is tested in unit test level.

### DIFF
--- a/internal/resource/default_user_secret.go
+++ b/internal/resource/default_user_secret.go
@@ -29,6 +29,7 @@ const (
 	DefaultUserSecretName = "default-user"
 	bindingProvider       = "rabbitmq"
 	bindingType           = "rabbitmq"
+	usernamePrefix        = "default_user_"
 )
 
 type DefaultUserSecretBuilder struct {
@@ -40,7 +41,7 @@ func (builder *RabbitmqResourceBuilder) DefaultUserSecret() *DefaultUserSecretBu
 }
 
 func (builder *DefaultUserSecretBuilder) Build() (client.Object, error) {
-	username, err := randomEncodedString(24)
+	username, err := generateUsername(24)
 	if err != nil {
 		return nil, err
 	}
@@ -144,6 +145,18 @@ func (builder *DefaultUserSecretBuilder) updatePorts(secret *corev1.Secret) {
 			}
 		}
 	}
+}
+
+// generateUsername returns a base64 string that has "default_user_" as prefix
+// returned string has length 'l' when base64 decoded
+func generateUsername(l int) (string, error) {
+	encodedStr, err := randomEncodedString(l)
+	if err != nil {
+		return "", err
+	}
+
+	encodedStrSlice := []byte(encodedStr)
+	return string(append([]byte(usernamePrefix), encodedStrSlice[0:len(encodedStrSlice)-len(usernamePrefix)]...)), nil
 }
 
 func (builder *DefaultUserSecretBuilder) pluginEnabled(plugin v1beta1.Plugin) bool {
--- a/internal/resource/default_user_secret.go
+++ b/internal/resource/default_user_secret.go
@@ -150,13 +150,13 @@ func (builder *DefaultUserSecretBuilder) updatePorts(secret *corev1.Secret) {
 // generateUsername returns a base64 string that has "default_user_" as prefix
 // returned string has length 'l' when base64 decoded
 func generateUsername(l int) (string, error) {
-	encodedStr, err := randomEncodedString(l)
+	encoded, err := randomEncodedString(l)
 	if err != nil {
 		return "", err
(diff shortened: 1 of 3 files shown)


internal/resource/default_user_secret.go
```diff
@@ -146,6 +147,18 @@ func (builder *DefaultUserSecretBuilder) updatePorts(secret *corev1.Secret) {
...
+func generateUsername(l int) (string, error) {
+	encodedStr, err := randomEncodedString(l)
+	if err != nil {
+		return "", err
+	}
+
+	encodedStrSlice := []byte(encodedStr)
+	return string(append([]byte(usernamePrefix), encodedStrSlice[0:len(encodedStrSlice)-len(usernamePrefix)]...)), nil
+}
+
```
ansd, line 161: I think it's a bit clearer to read when directly generating the random string with the correct length, i.e. `randomEncodedString(l - len(usernamePrefix))` instead of generating 24 bytes random string and thereafter only appending the first 24 - 13 = 11 bytes.

internal/resource/default_user_secret.go
```diff
@@ -146,6 +147,18 @@ func (builder *DefaultUserSecretBuilder) updatePorts(secret *corev1.Secret) {
...
 
+// generateUsername returns a base64 string that has "default_user_" as prefix
+// returned string has length 'l' when base64 decoded
+func generateUsername(l int) (string, error) {
+	encodedStr, err := randomEncodedString(l)
+	if err != nil {
+		return "", err
+	}
+
+	encodedStrSlice := []byte(encodedStr)
```
ansd, line 158: We could rename the variable `encodedStrSlice` since it's a byte slice, not a string slice.

internal/resource/default_user_secret_test.go
```diff
@@ -72,15 +70,17 @@ var _ = Describe("DefaultUserSecret", func() {
...
+			By("creating a rabbitmq username that has a prefix 'default_user_', base64 encoded, and 24 characters in length when decoded", func() {
 				username, ok = secret.Data["username"]
 				Expect(ok).NotTo(BeFalse(), "Failed to find a key \"username\" in the generated Secret")
-				decodedUsername, err := b64.URLEncoding.DecodeString(string(username))
+				Expect(string(username)).To(HavePrefix("default_user_"))
+				//strList := strings.SplitAfter(string(username), "default_user_")
+				decoded, err := b64.URLEncoding.DecodeString(string(username))
 				Expect(err).NotTo(HaveOccurred())
-				Expect(len(decodedUsername)).To(Equal(24))
+				Expect(len(decoded)).To(Equal(24))
```
ansd, line 80: We should use `HaveLen` matcher.

internal/resource/default_user_secret_test.go
```diff
@@ -72,15 +70,17 @@ var _ = Describe("DefaultUserSecret", func() {
 				Expect(secret.Type).To(Equal(corev1.SecretTypeOpaque))
 			})
 
-			By("creating a rabbitmq username that is base64 encoded and 24 characters in length", func() {
+			By("creating a rabbitmq username that has a prefix 'default_user_', base64 encoded, and 24 characters in length when decoded", func() {
 				username, ok = secret.Data["username"]
 				Expect(ok).NotTo(BeFalse(), "Failed to find a key \"username\" in the generated Secret")
```
ansd, line 75: We should assert `Expect(ok).To(BeTrue())` instead of using double negation.

internal/resource/default_user_secret_test.go
```diff
@@ -72,15 +70,17 @@ var _ = Describe("DefaultUserSecret", func() {
...
-				decodedUsername, err := b64.URLEncoding.DecodeString(string(username))
+				Expect(string(username)).To(HavePrefix("default_user_"))
+				//strList := strings.SplitAfter(string(username), "default_user_")
```
ansd, line 77: Should we remove the comment?

ansd (commented):
I left some in-line comments. Feel free to amend or merge.

internal/resource/default_user_secret.go
ChunyiLyu: It does not work as `prefix + randomEncodedString(l - len(usernamePrefix))`. base64 encoded string does not have the same length as the input string.

Zerpet (approved):
Not to be False or To be True 🤔 🙈

ansd (approved):

Issue #840, opened 2021-09-13
Add an option to do not create default-user secret


olfway:
I would like to use a predefined admin username and password and prevent rabbitmq-operator from creating it

It seems I can create default-user secret first and then create a cluster, so the operator will use that secret,
but unfortunately in gitops way you cannot guarantee the order of resource creating

Also, I'm planning to use external-secret controller to get the secret from aws secret manager and it will take some time to fetch data and create a secret.

So I need a way to tell rabbitmq-operator to not create a secret and just wait for it

ChunyiLyu:
Hi @olfway, we discussed in a team meeting about supporting different secret backend (vault support is coming soon, but we haven't considered AWS secret for your information). Our conclusion is it's a relatively complicated track of work that we might start looking at soon. However, at this time, we cannot provide an estimate/guarantee about timeline.

On the other hand, there is a workaround that you can try with the current operator. The Rabbitmq Cluster Operator creates the k8s secret for the default user secret with a specific format and object name. The secret name is `RMQNAME-default-user`. If you create the secret with this name then create the RabbitMQ cluster, operator will not generate the user credential and will take the username and password from the pre existing secret. Would this workaround work for use case?

The default user secret has to have key `default_user.conf` and with the object name of `RMQNAME-default-user`. For example, if your rabbitmqcluster is named "sample". You can create a k8s secret such as:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: sample-default-user
stringData:
  default_user.conf: |
    default_user = <USERNAME-HERE>
    default_pass = <PASSWORD-HERE>
  password: <PASSWORD-HERE> # this line is only necessary if you also use messaging topology operator
  username: <USERNAME-HERE> # this line is only necessary if you also use messaging topology operator
```

olfway:
@ChunyiLyu thank you for the information!
I'm using argo-cd (git ops tool), so all manifest applied to k8s at once
I cannot create secret first, then wait and after that create cluster

I found I can override rabbitmq-confd volume itself:
```
apiVersion: rabbitmq.com/v1beta1
kind: RabbitmqCluster
spec:
  override:
    statefulSet:
      spec:
        template:
          spec:
            containers: []
            volumes:
              - name: rabbitmq-confd
                projected:
                  defaultMode: 420
                  sources:
                    - secret:
                        name: rabbitmq-cluster-admin
                        items:
                          - key: default_user.conf
                            path: default_user.conf
                    - configMap:
[...]

Issue #824, opened 2021-08-26
To create a cluster with TLS enabled, the certificate and private key should come from Vault's secret


MarcialRosales:
Context
This feature is part of the Vault integration opportunity. The goal is not to use K8s secrets to store any RabbitMQ secret such as a default user's credential or TLS cert/keys. Instead, these secrets should be in Vault and retrieved with the appropriate Vault authentication mechanism.

Scope
When the operator creates a cluster, it should retrieve the TLS certificate and private key from Vault rather than from a secret. We use Vault Agent (a.k.a. sidecar) to retrieve both files from Vault and store them in the pod's in the exact same locations where RabbitMQ expects them. Rotation is given for granted because Erlang will automatically picks the new cert/key when their corresponding files change.

Given that we are only enabling TLS, and not mTLS, we do not seem to need to retrieve the ca certificate used to validate client's certificates.

twhite0:
@MarcialRosales:

We would be interested in this functionality.  Do you have thoughts on if this would be configured to use a [PKI backend](https://www.vaultproject.io/docs/secrets/pki) or generated _out of band_ and stored in a KV2 for retrieval?

Thanks in advance.

MarcialRosales:
Hi @twhite0 , indeed we are supporting PKI backend. You will see in the examples we provide that you need to specify the path to the `pki/issue/<role>` of your choice. 
There is no need to have the PKI artifacts in a kv store path.

ansd:
Closing since https://github.com/rabbitmq/cluster-operator/pull/846 is merged.

Issue #823, opened 2021-08-26
Update default user's password in RabbitMQ when it has changed in Vault


MarcialRosales:
# Context
This feature is part of the vault integration opportunity. It is about rotating the default user's password. Not the username !

# Scope
Vault Agent (a.k.a. sidecar) keeps a RabbitMQ config file with the default user's credentials in sync with Vault. In other words, when the credentials change in Vault, the agent updates the local file in the cluster's pod.
We need a sidecar that monitor the local config file in the pod and updates RabbitMQ. Either by means of the management rest api or via rabbitmqctl (doing exec onto the rabbitmq container)

twhite0:
@MarcialRosales:  
We've used the [Vault Agent](https://www.vaultproject.io/docs/agent) with some success at fetching and initiating a command when credentials have changed.  In particular the [command](https://www.vaultproject.io/docs/agent/template#command) portion of the template allows for the agent to initiate a _simple_ command.  
We have products using this along with a _refresh_ endpoint which updates the credentials (command = curl ....).  We have other products which prefer to simply initiate a new Deployment on a credential change (command = kubectl ...).  They are incrementing an environment variable on the sidcar container such that all running pods will update/noop to the same value.  This does require patch access to the ServiceAccount in use which may be an issue in some shops.

MarcialRosales:
@twhite0 Thanks! that is indeed very convenient. In our initial exploration we opted to monitor the file. But invoking a command is far neater.

MarcialRosales:
@twhite0 :
We have tried the [command](https://www.vaultproject.io/docs/agent/template#command) option however we are limited by the commands available in the image that vault agent runs (`hashicorp/vault:1.8.1`). This docker image has a rather old/limited `wget` version (`BusyBox v1.32.1 `) which only supports  http `GET` method. However, in order to rotate the password in RabbitMQ, we would need `PUT` or `POST`. 

We are no longer very keen on our initial approach of having sidecar watcher container. This leaves us with just one alternative  -at least for now- which is the following:  Use  "**stateful override**" feature to set `vault.hashicorp.com/agent-image` annotation with a **customer-provided-image** which has all the necessary runtime (e.g. wget + some scripts)  to rotate the password in RabbitMQ.  what do you think? Based on your comment [above](https://github.com/rabbitmq/cluster-operator/issues/823#issuecomment-910663881) you must be using a custom image with either `curl` or `kubectl`.

twhite0:
[...]

PR #846, opened 2021-09-17
Support Vault via vault-k8s

### LINKED ISSUE #822: To create a cluster the default user's credentials should come from Vault's secret
# Context
This feature is part of the Vault integration opportunity. The goal is not to use K8s secrets to store any RabbitMQ secret such as a default user's credential. Instead, these secrets should be in Vault and retrieved with the appropriate Vault authentication mechanism.

# Scope
When the operator creates a cluster, it should retrieve the default user's username and password from Vault rather than auto-generating it. 

# Concerns
- What happens if we deployed a cluster with Vault disable and later on we enable it? Should the auto-generated secret for the default user (a.k.a. adminuser) be deleted? or left around?
- To unit test this functionality, we need Vault running and configured with the appropriate role and secret path the unit tests can refer to.

ansd:
This closes #822, #823, #824.

Before this PR, the cluster-operator was always creating the credentials of the default (admin) user putting them into a K8s Secret object. TLS is supported by the user providing a K8s Secret object containing certificate and private key.

However, some users want to store credentials outside of K8s in external secret stores such as HashiCorp Vault, AWS Secrets Manager, Google Secrets Manager or Azure Key Vault. Benefits of doing so include:
* all credentials in a single store
* encrypt data at rest
* auditing
* third-party rotation features

As explained in [this KubeCon talk](https://youtu.be/w0k7MI6sCJg?t=177) there 4 different approaches in K8s to consume external secrets:
1.  Direct API
2. Controller to mirrors secrets in K8s
3. Sidecar + MutatingWebhookConfiguration
4. Secrets Store CSI Driver

In this PR, we take the 3rd approach (`Sidecar + MutatingWebhookConfiguration`)  integrating with Vault using [vault-k8s](https://github.com/hashicorp/vault-k8s). If `spec.secretBackend.vault` is configured in the RabbimqCluster CRD, no K8s Secret will be created for the default user credentials. Instead a Vault init container + sidecar fetch credentials from Vault.

In future, we might (no promises) support approaches `2. Controller to mirrors secrets in K8s` as for example requested in #840 and `4. Secrets Store CSI Driver`. The latter seems to be the future proof K8s native way of fetching external secrets.

This PR adds support for 3 features:
1. #822: When creating a RabbitmqCluster, the default user credentials are fetched from Vault. The assumption is that the credentials were already created in Vault (via some cluster-operator external mechanism). The cluster-operator will not create a K8s Secret for the default user. Instead, a Vault init container will fetch the credentials and mount them to be used in the RabbitMQ container.
2. #823: Password rotation without restarting Pods. When the default user password changes in Vault, the Vault sidecar will update the password in the `/etc/rabbitmq/conf.d/11-default_user.conf` file. However, RabbitMQ cannot pick up config file changes on the fly. Therefore, some component need to update the password on the RabbitMQ server. The Vault sidecar container does not seem to contain the tools to HTTP PUT to the RabbitMQ Management API or to `kubectl exec` into the RabbitMQ container. Since we do not want to change the default Vault image, a 2nd side car can optionally be deployed. It contains a single Go binary that watches `11-default_user.conf` for changes. If a new password is detected, it HTTP PUT to the RabbitMQ management API updating the password server side and copies the password into `/var/lib/rabbitmq/.rabbitmqadmin.conf` to be used for the `rabbitmqadmin` CLI. Username rotation is not supported.
3. #824: To enable TLS, RabbitMQ server certificates are provided on-demand by [Vault PKI Engine](https://www.vaultproject.io/docs/secrets/pki) instead of being pre-put into a K8s Secret. The RabbitMQ pod uses a short-lived leaf certificate (aligning with Vault's philosophy of short-lived secrets). After each RabbitMQ pod start-up, a new certificate is issued by Vault. The private key is never stored in Vault. Before the short-lived server certificate expires, vault sidecar container will request new certificate putting it into `/etc/rabbitmq-tls/` where it will be picked up on-the-fly by the Erlang VM without the need to restart the pod.

No changes are done to the Erlang cookie: it will still be created by the cluster-operator and stored in a K8s Secret.

**Tests:**
- Unit tests were added for the new `admin-password-updater` sidecar container.
- Unit tests in the cluster-operator
- End-to-end tests are provided in the `docs/examples/vault-default-user` and `docs/examples/vault-tls` directories covering the above 3 features and will be run in our Concourse pipeline. The `setup.sh` scripts give users a high level idea of what needs to be setup on Vault server side (although the examples use only Vault server dev mode).

Most of the work in this PR was done by @MarcialRosales 🙌

This PR is a draft because the following **TODOs** are left:
- [x] Get feedback: @twhite0 would you be interested in testing the changes?
- [x] Re-test everything.
- [x] Add docs: specifically READMEs in the `docs/examples/vault-*/` directories.
- [ ] Merge squashing all commits.
- [ ] Merge https://github.com/rabbitmq/rabbitmq-website/pull/1272


twhite0:
@ansd:  
Thanks for reaching out!  We would love to review and test it.  Timing is great as I was starting the internal logistics of pulling, building and deploying branch builds to our clusters.

twhite0:
@ansd:  Quick update:
We were able to build and deploy the operator internally.
* Leveraged current cluster-operator.yml release process.
     - Modified the Deployment spec with our internal image built off of the vault-integration branch.
     - Modified the crd specification; adding [the new pieces](https://github.com/rabbitmq/cluster-operator/blob/vault-integration/config/crd/bases/rabbitmq.com_rabbitmqclusters.yaml#L3798-L3829).

We ran into some vault injector issues, we'll be picking it back up Monday AM.

admin-password-updater/main.go
```diff
@@ -0,0 +1,153 @@
...
+			log.Error(err, "failed to read CA file", "file", caFile)
+			return nil, err
+		}
+		caCertPool := x509.NewCertPool()
+		caCertPool.AppendCertsFromPEM(caCert)
+		tlsConfig := &tls.Config{
+			RootCAs: caCertPool,
+		}
+		transport := &http.Transport{TLSClientConfig: tlsConfig}
+		rmqc, err := rabbithole.NewTLSClient(managementURI, "", "", transport)
```
coro, line 117: Interested what the effect is on management API peer verification - I think this would mean peer verification on the management API isn't supported with this flow, though it's not widely used at all so that may be OK. Could be worth calling out.

admin-password-updater/updater/event_handler.go
```diff
@@ -0,0 +1,172 @@
...
+
+func fileChanged(defaultUserFile string, event fsnotify.Event) bool {
+	return event.Name == defaultUserFile &&
+		(event.Op&fsnotify.Create == fsnotify.Create ||
+			event.Op&fsnotify.Write == fsnotify.Write)
+}
+
+// updateInRabbitMQ sets newPasswd for existingUser in the RabbitMQ server.
+// It returns true if password could be sucessfully updated or password was already up-to-date.
+func (u *PasswordUpdater) updateInRabbitMQ(existingUser, newPasswd string) (ok bool) {
```
coro, line 118: The function signature here I found rather confusing. The functions nested within this one all refer to failures, error handling, etc., which make me think this should return an error rather than a bool

docs/api/rabbitmq.com.ref.asciidoc
```diff
@@ -298,7 +298,7 @@ Spec is the desired state of the RabbitmqCluster Custom Resource.
 | Field | Description
 | *`replicas`* __integer__ | Replicas is the number of nodes in the RabbitMQ cluster. Each node is deployed as a Replica in a StatefulSet. Only 1, 3, 5 replicas clusters are tested. This value should be an odd number to ensure the resultant cluster can establish exactly one quorum of nodes in the event of a fragmenting network partition.
 | *`image`* __string__ | Image is the name of the RabbitMQ docker image to use for RabbitMQ nodes in the RabbitmqCluster. Must be provided together with ImagePullSecrets in order to use an image in a private registry.
-| *`imagePullSecrets`* __link:https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#localobjectreference-v1-core[$$LocalObjectReference$$] array__ | List of Secret resource containing access credentials to the registry for the RabbitMQ image. Required if the docker registry is private.
```
coro, line 301: This particular line shouldn't be modified, though the autogen keeps trying to for some reason... We should probably separately look into why this happens

admin-password-updater/main.go
ansd: That's a good call.

When a user sets
```
management.ssl.verify = verify_peer
management.ssl.fail_if_no_peer_cert = true
```
I'd expect that RabbitMQ HTTP API clients are supposed to be configured only with client certificates?
In that use case, why would the admin user have a password in the first place?
If the admin user doesn't have a password, there is no need to rotate the password, hence no need to use the admin-password-updater.

As of today, i.e. before this PR, the cluster-operator will always create username and password putting it into a K8s secret. The whole flow of an admin user set up with client certs instead of username and password isn't well supported and not documented at all in the cluster-operator.  It's something we discussed with @yaronp68 and @cdickmann on Friday. We may want to support it in future. Therefore, I wouldn't call it out specifically for the Vault integration.

coro (approved):

docs/api/rabbitmq.com.ref.asciidoc
ansd: Fixed, thank you.

api/v1beta1/rabbitmqcluster_types.go
```diff
@@ -84,6 +85,74 @@ type RabbitmqClusterSpec struct {
...
+	// A sidecar container that updates credentials in RabbitMQ server.
+	// As of now, only image "rabbitmqoperator/admin-password-updater" exists.
+	// If this image is set, a sidecar container will be deployed that watches file
+	// /etc/rabbitmq/conf.d/11-default_user.conf for changes.
+	// This file is changed when the Vault sidecar observes a new default user password in Vault server
+	// placing the new password into this file.
+	// Since RabbitMQ server cannot pick up this config file change on the fly, the admin-password-updater
+	// will update the password on the RabbitMQ server and also copy the new password to file
+	// /var/lib/rabbitmq/.rabbitmqadmin.conf (used by rabbitmqadmin CLI).
+	// In other words, image "rabbitmqoperator/admin-password-updater>" exists to allow
```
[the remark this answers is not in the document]
ansd: Good spot, fixed, thank you.

admin-password-updater/updater/event_handler.go
ansd: It now returns an error. Thank you.

admin-password-updater/updater/event_handler.go
```diff
@@ -0,0 +1,172 @@
...
+
+	u.Log.V(3).Info("HTTP response", "method", http.MethodPut, "path", pathUsers, "status", resp.Status)
+	u.Log.V(2).Info("updated password on RabbitMQ server", "user", existingUser)
+	return true
+}
+
+func (u *PasswordUpdater) handleHTTPError(err error, httpMethod, pathUsers, newPasswd string) (ok bool) {
+	// as returned in
+	// https://github.com/michaelklishin/rabbit-hole/blob/1de83b96b8ba1e29afd003143a9d8a8234d4e913/client.go#L153
+	if err.Error() == "Error: API responded with a 401 Unauthorized" {
```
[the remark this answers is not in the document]
ansd: Very good call. I thought about the same. We could use https://github.com/michaelklishin/rabbit-hole/blob/a692ae424d424bff3b5c9a8bbd97f4f362756a0f/error.go#L5-L14. However, I fear we might break the rabbithole API by doing so since other users might depend on that specific error message (as for example implemented in here)? WDYT, should we risk it? I'm happy either way.

twhite0:
@ansd:

Couple quick followups:
*  What was the driver for adding in the SecurityContext work in the Statefulset.
*  We found a couple more annotations which we needed to add into the Statefulset override section of our RabbitmqCluster definition to get things cooking.  Still unpacking the changes to see if this might be something valuable to add or if its specific to us.
     - `vault.hashicorp.com/namespace`: _name of Vault namespace_
     - `vault.hashicorp.com/auth-path`: _auth path_

ansd:
@twhite0 thank you for your followups, these are excellent questions.

> What was the driver for adding in the SecurityContext work in the Statefulset.

We set a PodSecurityContext to run as user 999 and FSGroup 999:
```
			SecurityContext: &corev1.PodSecurityContext{
				FSGroup:   &rabbitmqUID,
				RunAsUser: &rabbitmqUID,
			},
```

We just removed the other container SecurityContext via https://github.com/rabbitmq/cluster-operator/pull/846/commits/6b529d91c07d3fe17aeeb8c7b74d7a0256a5ca02 since they are indeed not needed because they were only setting `RunAsUser: &rabbitmqUID` which was already set in the PodSecurityContext.

The driver to set the PodSecurityContext is to run processes as RabbitMQ user (as opposed to running as root user).
FSGroup ensures that mounted volumes belong to the RabbitMQ group.

For OpenShift, this is not needed since the runtime will assign arbitrary user IDs to the containers as documented in https://www.rabbitmq.com/kubernetes/operator/using-on-openshift.html#arbitrary-user-ids.

> We found a couple more annotations which we needed to add into the Statefulset override section of our RabbitmqCluster definition to get things cooking.

You can now set arbitrary Vault annotations as documented in https://github.com/rabbitmq/cluster-operator/blob/2f0e49c5142ba842298d96ee193427c7827c9c78/docs/examples/vault-default-user/rabbitmq.yaml#L10-L13 User set Vault annotations will override cluster-operator-set Vault annotations.

ansd:
Thanks a lot @twhite0 for your feedback. That's perfect!

@MarcialRosales had some good refactoring suggestions yesterday, he will change Vault field names slightly. Thereafter we should be ready to merge.

api/v1beta1/rabbitmqcluster_types.go
```diff
@@ -153,10 +149,17 @@ type VaultTLSSpec struct {
...
-	return spec.PathDefaultUser != ""
+	return spec.DefaultUserPath != ""
+}
+func (spec *VaultSpec) ResolveDefaultUserUpdaterInage() string {
+	if spec.DefaultUserUpdaterImage == "" {
+		return "rabbitmqoperator/admin-password-updater:0.1.1"
+	}
+	return spec.DefaultUserUpdaterImage
+
 }
```
ansd, line 163: Why don't we use kubebuilder default?
E.g. as done in https://github.com/rabbitmq/cluster-operator/blob/588649d4cb6c0e8efbd0f1639cd5b89e55363c5f/api/v1beta1/rabbitmqcluster_types.go#L54

config/crd/bases/rabbitmq.com_rabbitmqclusters.yaml
```diff
@@ -3809,9 +3806,12 @@ spec:
                             type: string
                           description: Vault annotations that override the Vault annotations set by the cluster-operator. For a list of valid Vault annotations, see https://www.vaultproject.io/docs/platform/k8s/injector/annotations
                           type: object
-                        pathDefaultUser:
+                        defaultUserPath:
                           description: Path in Vault to access a KV secret with the fields username and password for the default user. For example "secret/data/rabbitmq/config".
                           type: string
+                        defaultUserUpdaterImage:
+                          description: 'sidecar container that updates the defaultUser''s password in RabbitMq when it changes in Vault. Additionally, it updates /var/lib/rabbitmq/.rabbitmqadmin.conf (used by rabbitmqadmin CLI). A default image name and version is provided (rabbitmqoperator/admin-password-updater:0.1.1). Only override when the default image version is not appropriate. When set to empty string, it disables the sidecar and passwords are not updated in RabbitMQ. e.g. defaultUserUpdaterImage: ""`'
```
ansd, line 3813: typo:
`defaultUser''s password in RabbitMq`

docs/api/rabbitmq.com.ref.asciidoc
ansd: @MarcialRosales we shouldn't change this line. See Connor's comment.

docs/examples/vault-default-user/rabbitmq.yaml
```diff
@@ -11,8 +11,9 @@ spec:
...
-      pathDefaultUser: secret/data/rabbitmq/config
-    # Optionally, deploy a sidecar container that rotates admin password in RabbitMQ server
-    # whenever Vault sidecar container puts a new password into
-    # /etc/rabbitmq/conf.d/11-default_user.conf
-    credentialUpdaterImage: "rabbitmqoperator/admin-password-updater:0.1.1"
+      defaultUserPath: secret/data/rabbitmq/config
+      # Optionally, override the sidecar container image that rotates admin password in RabbitMQ server
+      # whenever Vault sidecar container puts a new password into
+      # If not specified, it defaults to "rabbitmqoperator/admin-password-updater:0.1.1"
+      # /etc/rabbitmq/conf.d/11-default_user.conf
```
ansd, line 18: The ordering of the lines is not correct.

Also, let's not define the default version here because we don't want to update the default image version in the examples whenever the image version changes.

internal/resource/statefulset_test.go
```diff
@@ -975,7 +978,7 @@ default_pass = {{ .Data.data.password }}
 									"memory": k8sresource.MustParse("512Ki"),
 								},
 							},
-							Image: instance.Spec.SecretBackend.CredentialUpdaterImage,
+							Image: "rabbitmqoperator/admin-password-updater:0.1.1",
```
ansd, line 981: That's brittle because the unit test will fail when we bump the default version.

internal/resource/statefulset_test.go
```diff
@@ -1031,12 +1035,21 @@ default_pass = {{ .Data.data.password }}
...
 					})
+
+					When("DefaultUserUpdaterImage is set to non-empty string", func() {
+						BeforeEach(func() {
+							instance.Spec.SecretBackend.Vault.DefaultUserUpdaterImage = "another-image"
+						})
+						It("sidecar is configured with the configured image name", func() {
+							Expect(sidecar.Image).To(Equal("another-image"))
+						})
+					})
```
ansd, line 1046: The CRD docs say:
> When set to empty string, it disables the sidecar and passwords are not updated in RabbitMQ.

Can we add a unit test for that?

api/v1beta1/rabbitmqcluster_types.go
MarcialRosales: If we use kubebuilder, it provides that default value under secretBackend.vault which several tests cases did not expect hence they failed.  For this reason, I decided to provide the default via logic rather than thru this kubebuilder annotation.

internal/resource/statefulset_test.go
MarcialRosales: That is true. I was going to check against non-emtpy but given that we always check against full strings I thought it was more appropriate. But yes, it is more brittle because we have to update the unit tests too.

internal/resource/statefulset_test.go
MarcialRosales: No, that was unintentionally left behind.

api/v1beta1/rabbitmqcluster_types.go
ansd: Fixed in https://github.com/rabbitmq/cluster-operator/pull/846/commits/9b9a3d0d61f0ed838b279ac767dcdbbdc236a0b2

docs/examples/vault-default-user/rabbitmq.yaml
ansd: Fixed in https://github.com/rabbitmq/cluster-operator/pull/846/commits/9b9a3d0d61f0ed838b279ac767dcdbbdc236a0b2

internal/resource/statefulset_test.go
ansd: Fixed in https://github.com/rabbitmq/cluster-operator/pull/846/commits/9b9a3d0d61f0ed838b279ac767dcdbbdc236a0b2

config/crd/bases/rabbitmq.com_rabbitmqclusters.yaml
ansd: Fixed in https://github.com/rabbitmq/cluster-operator/pull/846/commits/9b9a3d0d61f0ed838b279ac767dcdbbdc236a0b2

docs/api/rabbitmq.com.ref.asciidoc
ansd: Fixed in https://github.com/rabbitmq/cluster-operator/pull/846/commits/a231127c53cca2479cdebcf822ee85fe9837d5f9

twhite0:
@ansd:

We noticed the `status` is defaulting to supplying `secretReference` information. 
* [Definition](https://github.com/rabbitmq/cluster-operator/blob/vault-integration/api/v1beta1/rabbitmqcluster_status.go#L10-L44)
* [Recon Invocation](https://github.com/rabbitmq/cluster-operator/blob/c813d56785460dd9ecef7a2423851c90afeeaf4e/controllers/rabbitmqcluster_controller.go#L219-L221)
* [Status setting](https://github.com/rabbitmq/cluster-operator/blob/c813d56785460dd9ecef7a2423851c90afeeaf4e/controllers/reconcile_status.go#L21-L29)

credit: @ssheth1

controllers/rabbitmqcluster_controller_test.go
```diff
@@ -191,6 +191,39 @@ var _ = Describe("RabbitmqClusterController", func() {
 
 	})
 
+	FContext("Vault is enabled for DefaultUser", func() {
```
ansd, line 194: This test case is focused.

controllers/rabbitmqcluster_controller_test.go
```diff
@@ -191,6 +191,39 @@ var _ = Describe("RabbitmqClusterController", func() {
...
+
+			Expect(client.Create(ctx, cluster)).To(Succeed())
+			waitForClusterCreation(ctx, cluster, client)
+		})
+
+		AfterEach(func() {
+			Expect(client.Delete(ctx, cluster)).To(Succeed())
+		})
+
+		FIt("does not expose DefaultUser or its Binding as status", func() {
```
ansd, line 219: This test case is focused too

ansd:
@ssheth1 @twhite0 very good spot, thank you :) This is fixed now.

MarcialRosales (approved):

PR #858, opened 2021-09-28
Move default image from CRD spec to command line flag

### LINKED ISSUE #856: CRD needs templating which does not enable us to deploy the operator instance in the bitnami/rabbitmq-cluster-operator chart
## Describe the bug

I want to use the [Bitnami RabbitMQ operator chart](https://github.com/bitnami/charts/tree/master/bitnami/rabbitmq-cluster-operator) to create the RabbitMQ operator and also want to instantiate the operator with an instance. 
Normally for Operator Helm charts, we place the CRD(s) under the crds/ folder. This is useful because this way Helm ensure that CRDs are installed BEFORE the rest of the manifests (avoiding what you faced). This is explained in the [Helm official docs](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/)
However, for this very specific chart we couldn’t use the crds/ folder since the CRD(s) used by this operator depend on some templates, and templates can only be applied to manifests under the templates/ folder. Some examples:
https://github.com/bitnami/charts/blob/master/bitnami/rabbitmq-cluster-operator/templates/crd-rabbitmq-cluster.yaml#L996
https://github.com/bitnami/charts/blob/master/bitnami/rabbitmq-cluster-operator/templates/crd-rabbitmq-cluster.yaml#L1004

## To Reproduce
1. Deploy the [Bitnami RabbitMQ Operator helm chart](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/) using `extraDeploy` parameter to instantiate the operator instance.
2. The installation fails saying the operator is not deployed.

**Expected behavior**
CRDs should not need any templating, in which case, CRDs get deployed before the rest of the templates in a chart. So we will not face the issue described above. 

## Screenshots
NA

- RabbitMQ: [e.g. 3.8.0]
- RabbitMQ Cluster Operator: 1.8.3
- Kubernetes: 1.20
- Cloud provider or hardware configuration: TKGs on Prem
 
## Additional context
NA

coro:
## Summary Of Changes

Instead of the default image for RabbitmqClusters being hard-coded into
the CRD, it is now available as a command-line flag to the operator's
`manager` binary. Changing the operator's Deployment manifest to call
`/manager --default-rabbitmq-image image:tag` will now cause all
RabbitmqClusters deployed by the operator to use that default image if
one is not specified in `spec.image`.

This closes #856

## Additional Context
This change will make it easier to deploy the operator using helm - the CRD itself will no longer change when bumping the default image, and so the raw CRD manifest can be placed in the `crd/` directory, while still allowing the default image to be configurable on a per-operator-Deployment scope.

## Local Testing

```
```

```
$ kubectx dev-bunny
$ make destroy deploy-dev
$ make system-tests
```

### DIFF
--- a/controllers/rabbitmqcluster_controller.go
+++ b/controllers/rabbitmqcluster_controller.go
@@ -57,12 +57,13 @@ const (
 // RabbitmqClusterReconciler reconciles a RabbitmqCluster object
 type RabbitmqClusterReconciler struct {
 	client.Client
-	Scheme        *runtime.Scheme
-	Namespace     string
-	Recorder      record.EventRecorder
-	ClusterConfig *rest.Config
-	Clientset     *kubernetes.Clientset
-	PodExecutor   PodExecutor
+	Scheme               *runtime.Scheme
+	Namespace            string
+	Recorder             record.EventRecorder
+	ClusterConfig        *rest.Config
+	Clientset            *kubernetes.Clientset
+	PodExecutor          PodExecutor
+	DefaultRabbitmqImage string
 }
 
 // the rbac rule requires an empty row at the end to render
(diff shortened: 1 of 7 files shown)


controllers/rabbitmqcluster_controller.go
```diff
@@ -113,6 +114,20 @@ func (r *RabbitmqClusterReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+		if err = r.Update(ctx, rabbitmqCluster); err != nil {
+			if k8serrors.IsConflict(err) {
+				logger.Info("failed to update image because of conflict; requeueing...",
+					"namespace", rabbitmqCluster.Namespace,
+					"name", rabbitmqCluster.Name)
+				return ctrl.Result{RequeueAfter: 2 * time.Second}, nil
+			}
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{RequeueAfter: 2 * time.Second}, err
```
Zerpet, line 128: If the image in the spec is empty, we set it to the default, update is successful, what's the reasoning to requeue here? 🤔

Zerpet (approved):
Approving after doing acceptance and adding my own commit to the Makefile 🙈

PR #859, opened 2021-09-29
Add documented examples for NetworkPolicies

### LINKED ISSUE #835: Network policy resource

In a production environnement, all network traffic is blocked between pods, we must declare network-policy to allow traffic between operator and rabbitmq server pods.

The operator should create this network policy.

Create mannually the network policy, hard to figure out traffic rules between operator and servers pods (which ports ?)

coro:
This closes #835 

## Summary Of Changes
[NetworkPolicies](https://kubernetes.io/docs/concepts/services-networking/network-policies/) allow users to restrict the source/destination of ingress/egress traffic of their Pods. This documented example demonstrates the kinds of traffic a user may wish to restrict/allow on their cluster.

## Additional Context
You can enable the network plugin on GKE for testing with:
```
gcloud container clusters create CLUSTER_NAME --enable-network-policy
```

## Local Testing

```
```

```
$ kubectx dev-bunny
$ make destroy deploy-dev
$ make system-tests
```

### DIFF
--- a/docs/examples/network-policies/README.md
+++ b/docs/examples/network-policies/README.md
@@ -8,8 +8,20 @@ By defining NetworkPolicies, you can restrict the network entities with which yo
 from reaching the cluster. It is important to note that once a RabbitmqCluster Pod, or any other Pod for that matter, is the target of any
 NetworkPolicy, it becomes isolated to all traffic except that permitted by a NetworkPolicy.
 
-The first policy in this example, [allow-inter-node-traffic.yaml](./allow-inter-node-traffic.yaml) ensures that the only nodes in the RabbitmqCluster
-send or receive traffic with each other on the ports used for inter-node communication.
+The following example policies all target (and therefore, affect the Pods of) the specific RabbitmqCluster deployed by [rabbitmq.yaml](./rabbitmq.yaml).
+This is done by targetting the RabbitmqCluster Pods using podSelector label matching:
+```yaml
+spec:
+  podSelector:
+    matchLabels:
+      app.kubernetes.io/component: rabbitmq
+      app.kubernetes.io/name: network-policies
+```
+To create policies that match any RabbitmqCluster, you can remove the `app.kubernetes.io/name` labelSelector. Bear in mind this might not always
+be appropriate for all NetworkPolicies; for example, inter-node traffic should be restricted on a per-RabbitmqCluster scope.
+
+The first policy in this example, [allow-inter-node-traffic.yaml](./allow-inter-node-traffic.yaml) ensures that only the Pods in the RabbitmqCluster
+can send or receive traffic with each other on the ports used for inter-node communication.
 
 The second policy, [allow-operator-traffic.yaml](./allow-operator-traffic.yaml), allows the cluster-operator and the messaging-topology-operator to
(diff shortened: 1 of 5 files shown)


docs/examples/network-policies/README.md
```diff
@@ -0,0 +1,23 @@
...
+information on NetworkPolicies, see the [Kubernetes documentation](https://kubernetes.io/docs/concepts/services-networking/network-policies/).
+
+By defining NetworkPolicies, you can restrict the network entities with which your RabbitmqCluster can communicate, and prevent unrecognised traffic
+from reaching the cluster. It is important to note that once a RabbitmqCluster Pod, or any other Pod for that matter, is the target of any
+NetworkPolicy, it becomes isolated to all traffic except that permitted by a NetworkPolicy.
+
+The first policy in this example, [allow-inter-node-traffic.yaml](./allow-inter-node-traffic.yaml) ensures that the only nodes in the RabbitmqCluster
+send or receive traffic with each other on the ports used for inter-node communication.
+
+The second policy, [allow-operator-traffic.yaml](./allow-operator-traffic.yaml), allows the cluster-operator and the messaging-topology-operator to
```
ChunyiLyu, line 14: I think only the admin password update logic from cluster operator connects to RMQ via HTTP. Do you think it's worth calling it out?

docs/examples/network-policies/allow-inter-node-traffic.yaml
```diff
@@ -0,0 +1,58 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: allow-inter-node-traffic
+spec:
+  podSelector:
+    matchLabels:
+      app.kubernetes.io/component: rabbitmq
```
ChunyiLyu, line 8: The same suggestion applies for the other network policies as well. Do you think we should mention that people could set labels to only matches a particular RabbitmqCluster. The label `app.kubernetes.io/name` is set to the name of the CR.

ChunyiLyu (approved):
Looks great! Please see my comments for minor suggestions. 

Plus, what do you think about adding a section in our `rabbitmq.com` doc to link to this example? I think `rabbitmq.com` could be a natural place for people to discover how to configure network policies.

docs/examples/network-policies/allow-rabbitmq-traffic.yaml
```diff
@@ -0,0 +1,28 @@
...
+  policyTypes:
+  - Ingress
+  ingress:
+  - ports:
+    - port: 5672  # AMQP
+    - port: 5671  # AMQP + TLS
+    - port: 5552  # Streams
+    - port: 5551  # Streams + TLS
+    - port: 15672 # HTTP API
+    - port: 15672 # HTTP API + TLS
```
Zerpet, line 19: Do you mean 15671? 🙈

docs/examples/network-policies/README.md
coro: I figured I'd leave it generic so that if we have cause to expand it later it's not going to break anything by being to restrictive. Similarly if we merge the two operators - they both have the label `app.kubernetes.io/component: rabbitmq-operator` so nothing needs to change there.

PR #886, opened 2021-11-03
Answer some production-ready questions that came up in private threads

gvmw:
I expect this to preempt some follow-up questions which I have seen come up a few times already. I encourage others to answer follow-up questions in the same way. It will save us from repeating the same answers, and sharing what we know to be true.

If you think that there is too much information in this README, we can split the questions into a separate file. I think that approach will fragment the information and make it less accessible, but I am open to what others feel would work better.


gvmw:
Follow-up to https://github.com/rabbitmq/cluster-operator/pull/401

ansd (approved):

Zerpet (approved):

mkuratczyk (approved):

ChunyiLyu (approved):

PR #887, opened 2021-11-03
Do not requeue error when disableNonTLSListen is misconfigured

### LINKED ISSUE #884: Improve UX when Disable non-TLS listeners is true and TLS is not configured
The following YAML can be applied successfully:

```yaml
kind: RabbitmqCluster
metadata:
  name: my-rabbitmqcluster
spec:
  replicas: 1
  tls:
    disableNonTLSListeners: true
```

The operator will return an error in the reconcile loop, and it will try infinitely. We know this can't succeed, therefore we shouldn't retry. We should log the error, as we currently do, set a status condition reflecting this problem, and stop requeueing.

ChunyiLyu:
This closes #884

## Summary Of Changes
- do not requeue on error when `disableNonTLSListen` is set to true but TLS is not enabled; this is a user configuration error and need to be fixed by user updating rabbitmqcluster manifest which will trigger reconcile anyways
- `reconcileTLS()` returns a special error `disableNonTLSConfigErr`. `Reconcile()` identifies the error and exit the reconcile so no k8s resource is created for the misconfigured rabbitmqcluster.

### DIFF
--- a/controllers/rabbitmqcluster_controller.go
+++ b/controllers/rabbitmqcluster_controller.go
@@ -13,6 +13,7 @@ package controllers
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"reflect"
 	"strings"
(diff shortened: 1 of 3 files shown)


Zerpet (approved):
🚀

PR #905, opened 2021-11-26
Generate `rabbitmq_cluster` label in observability examples

binarin:
## Summary Of Changes

Currently grafana dashboards are always joining on `rabbitmq_identity_info` to be able to filter by cluster. This makes all queries barely readable, and have a performance cost. But we can order prometheus to provide `rabbitmq_cluster` label directly, greatly simplifying everything.

### DIFF
--- a/observability/prometheus/monitors/rabbitmq-servicemonitor.yml
+++ b/observability/prometheus/monitors/rabbitmq-servicemonitor.yml
@@ -10,10 +10,18 @@ spec:
     scheme: http
     interval: 15s
     scrapeTimeout: 14s
+    relabelings:
+    - sourceLabels:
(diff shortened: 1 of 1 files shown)


binarin:
This pattern repeats over and over everywhere
```
sum(
  rate(
    rabbitmq_channel_messages_published_total[60s])
    * on(instance)  group_left(rabbitmq_cluster)
      rabbitmq_identity_info{rabbitmq_cluster="$rabbitmq_cluster", namespace="$namespace"})
```
But with better prometheus config it's possible to go for the needed metric directly
```
sum(rate(rabbitmq_channel_messages_published_total[60s])
       {rabbitmq_cluster="$rabbitmq_cluster", 
         namespace="$namespace"})
```

PR #936, opened 2022-01-24
fix: don't set empty image pull secrets

oliver-rew:
## Summary Of Changes
commit "5f98b43" added the ability to set image pull secrets from the
DEFAULT_IMAGE_PULL_SECRETS env var. 
https://github.com/rabbitmq/cluster-operator/commit/5f98b4304fbee6153cbf6b4f9ccb9255993f08de

In the case where the `DEFAULT_IMAGE_PULL_SECRETS` env var is unset, `DefaultImagePullSecrets` will be `""`, and `strings.Split(r.DefaultImagePullSecrets, ",")` will return an array of length one containing an empty string and an image pull secret will be created with an empty string. This commit adds check to the output of `strings.Split()` to ensure we don't set an empty strings.

When this empty string is saved and parsed by k8s, resulting operations get this error:
```
$ kubectl rollout restart statefulset/rabbitmq-cluster-server
error: statefulsets.apps "rabbitmq-cluster-server" map: map[] does not contain declared merge key: name
```

you can also see the problem here:
```
$ kubectl get statefulset/rabbitmq-cluster-server -o json | grep -i imagepullsecret -A2
                "imagePullSecrets": [
                    {}
                ],
```

## Local Testing

```
```

```
$ kubectx dev-bunny
$ make destroy deploy-dev
$ make system-tests
```

### DIFF
--- a/controllers/rabbitmqcluster_controller.go
+++ b/controllers/rabbitmqcluster_controller.go
@@ -131,8 +131,12 @@ func (r *RabbitmqClusterReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	}
 
 	if rabbitmqCluster.Spec.ImagePullSecrets == nil {
+		// split the comma separated list of default image pull secrets from
+		// the 'DEFAULT_IMAGE_PULL_SECRETS' env var, but ignore empty strings.
 		for _, reference := range strings.Split(r.DefaultImagePullSecrets, ",") {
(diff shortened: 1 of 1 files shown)


Zerpet (approved):
Thank you! The error in the system tests looks like a flake caused by GitHub Actions.

coro (approved):
Thanks for the contribution!

Issue #869, opened 2021-10-14
Distribute operator image on multiple registries


torblerone:
**Is your feature request related to a problem? Please describe.**
The new pricing model of Docker Inc. is making it hard to still use DockerHub as a public image registry for businesses without paying for it. There's plenty other public registry providers that can be used.

**Describe the solution you'd like**
Maybe `quay.io`?

Zerpet:
We are planning to publish our images to GitHub Container Registry, in addition to DockerHub. It's not clear whether we'll stop publishing to DockerHub.

torbendury:
Hello @Zerpet, are there any news on this, any ETA? :slightly_smiling_face:

Zerpet:
Hey folks, I brought this up to the team a few days ago, and we decided that it makes sense to publish to `quay.io`. I'm exploring that registry and noting down what needs to be done on our CI to enable this registry.

Zerpet:
Hey folks, I have manually pushed the latest version of the Cluster Operator to `quay.io`:

https://quay.io/repository/rabbitmqoperator/cluster-operator?tab=tags

Our CI is almost ready to start publishing to this registry following releases. We will publish only GA images. Dev, alphas, or other pre-releases will continue to live in DockerHub.

Hope this helps!

philipp-hb:
cool, thx

Zerpet:
Closing this issue, as we are publishing the images to quay.io and publishing a YAML manifest to our releases to deploy using the image from quay.io registry, [starting from 1.12.0](https://github.com/rabbitmq/cluster-operator/releases/tag/v1.12.0) and onwards.

P.s: I noticed the double extension in the file name. Future releases won't have that problem 🙂

PR #689, opened 2021-05-12
Explore: classic peer discovery without randomised startup delay

ansd:
Relates #662.

Enforce pod 0 to form the cluster.

**Pros:**
* no randomised startup delay => always a single node (pod 0) forms the cluster
* no sophisticated locking mechanism

**Cons:**
* If pod 0 never starts, the cluster wont't be created. I think that's okay. For new clusters, all nodes should be up & running.
* Sporadically, there are container restarts since nodes crash with below `schema_integrity_check_failed` error. Eventually the cluster gets created successfully and all pods become `ready`. Solution: have one node initialize itself at a time.
```
2021-05-12 15:34:02.786 [debug] <0.67.0> Supervisor kernel_safe_sup started pg_local:start_link() at pid <0.1065.0>
2021-05-12 15:34:02.876 [info] <0.44.0> Application mnesia exited with reason: stopped

2021-05-12 15:34:02.877 [error] <0.273.0>
2021-05-12 15:34:02.877 [info] <0.44.0> Application mnesia exited with reason: stopped
2021-05-12 15:34:02.877 [error] <0.273.0> BOOT FAILED
2021-05-12 15:34:02.877 [error] <0.273.0> ===========
2021-05-12 15:34:02.877 [error] <0.273.0> Error during startup: {error,
2021-05-12 15:34:02.877 [error] <0.273.0>                           {schema_integrity_check_failed,
BOOT FAILED
===========
Error during startup: {error,
                          {schema_integrity_check_failed,
2021-05-12 15:34:02.877 [error] <0.273.0>                               [{table_attributes_mismatch,rabbit_user,
2021-05-12 15:34:02.878 [error] <0.273.0>                                    [username,password_hash,tags,
                              [{table_attributes_mismatch,rabbit_user,
                                   [username,password_hash,tags,
                                    hashing_algorithm],
                                   [username,password_hash,tags,
                                    hashing_algorithm,limits]},
2021-05-12 15:34:02.878 [error] <0.273.0>                                     hashing_algorithm],
2021-05-12 15:34:02.878 [error] <0.273.0>                                    [username,password_hash,tags,
2021-05-12 15:34:02.878 [error] <0.273.0>                                     hashing_algorithm,limits]},
                               {table_attributes_mismatch,rabbit_vhost,
2021-05-12 15:34:02.878 [error] <0.273.0>                                {table_attributes_mismatch,rabbit_vhost,
2021-05-12 15:34:02.878 [error] <0.273.0>                                    [virtual_host,limits],
[...]

PR #942, opened 2022-01-28
Helper overlay to patch the Operator image

Zerpet:

## Summary Of Changes
This will be used in the pipeline to publish to additional registries.

## Additional Context
Related [#869]

## Local Testing

Make sure you have `ytt` installed, then run:

```bash
kustomize build config/installation/ | ytt -f- -f config/ytt/overlay-manager-image.yaml \
  --data-value operator_image=quay.io/bunny:latest
```

Observe the image of the Operator Deployment is set to `quay.io/bunny:latest`.

Edit: tag correct issue 🙈

### DIFF
--- a/config/ytt/overlay-manager-image.yaml
+++ b/config/ytt/overlay-manager-image.yaml
@@ -1,7 +1,7 @@
 #@ load("@ytt:overlay", "overlay")
 #@ load("@ytt:data", "data")
 
-#@overlay/match by=overlay.subset({"kind": "Deployment"}), expects="1+"
+#@overlay/match by=overlay.subset({"kind": "Deployment", "metadata":{"name": "rabbitmq-cluster-operator"}}), expects="1+"
 ---
 spec:
   template:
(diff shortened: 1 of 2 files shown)


coro:
Not sure why #689 is related!

config/ytt/overlay-manager-image.yaml
```diff
@@ -0,0 +1,13 @@
+#@ load("@ytt:overlay", "overlay")
+#@ load("@ytt:data", "data")
+
+#@overlay/match by=overlay.subset({"kind": "Deployment"}), expects="1+"
```
coro, line 4: Is it worth maybe being specific about which Deployment is in the subset? Just in case one used this with a large group of manifests containing other Deployments.

Zerpet:
> Not sure why #689 is related!

Oops 🙊 It is actually #869

config/ytt/overlay-manager-image.yaml
Zerpet: Good catch! I just hope `kustomize` doesn't mangle the name of the Deployment ever.

Zerpet:
Made the changes to match the Operator deployment more precisely. If we ever change the name of the Operator deployment, the overlay won't match anything, in which case we'll get a red pipeline somewhere.

coro (approved):

PR #1010, opened 2022-04-06
Adding an https-url import definition example to import a large impor…

DanielePalaia:
…t definition at boot time from the web

This closes #

## Local Testing

```
```

```
$ kubectx dev-bunny
$ make destroy deploy-dev
$ make system-tests
```

### DIFF
--- a/docs/examples/import-https-url-definition/rabbitmq.yaml
+++ b/docs/examples/import-https-url-definition/rabbitmq.yaml
@@ -1,0 +1,21 @@
+apiVersion: rabbitmq.com/v1beta1
+kind: RabbitmqCluster
+metadata:
+  name: import-https-url-definitions 
+spec:
+  service:
+    type: NodePort
+  replicas: 1
+  resources:
+    requests:
+      cpu: 2
+      memory: 8Gi
+    limits:
+      cpu: 2
+      memory: 8Gi
+  rabbitmq:
+    additionalConfig: |
+      definitions.import_backend = https
+      definitions.https.url = https://raw.githubusercontent.com/rabbitmq/sample-configs/main/lot-of-queues/5k-queues.json
+      definitions.tls.versions.1 = tlsv1.2
+      vm_memory_high_watermark_paging_ratio = 0.85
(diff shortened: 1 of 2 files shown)


mkuratczyk (approved):
Please remove the `NodePort` section. Otherwise, it's great.

docs/examples/import-https-url-definition/rabbitmq.yaml
```diff
@@ -0,0 +1,19 @@
...
+  name: import-https-url-definitions 
+spec:
+  replicas: 1
+  resources:
+    requests:
+      cpu: 2
+      memory: 8Gi
+    limits:
+      cpu: 2
+      memory: 8Gi
```
ChunyiLyu, line 13: Are the resource request and limit required for this example to work? If not, lets remove it to keep the example simple and just about the importing definition itself.

docs/examples/import-https-url-definition/rabbitmq.yaml
```diff
@@ -0,0 +1,19 @@
...
+  rabbitmq:
+    additionalConfig: |
+      definitions.import_backend = https
+      definitions.https.url = https://raw.githubusercontent.com/rabbitmq/sample-configs/main/lot-of-queues/5k-queues.json
+      definitions.tls.versions.1 = tlsv1.2
+      vm_memory_high_watermark_paging_ratio = 0.85
```
ChunyiLyu, line 19: Similar question to above, is configuring memory high water mark necessary here?

ChunyiLyu (commented):
Please see comments about resource limit and request and the memory high water mark configurations. If they are not required for the example, let's remove it. If they are indeed required, may I suggest adding comments in the example or the README to explain what they are for?

DanielePalaia:
@ChunyiLyu thanks for your review, indeed I was planning to write the motivation in the README but I then forgot. The example takes a lot of memory to load a very big definition file, so I think better to put the resource requests/limits. I removed the vm_memory_high_watermark_paging_ratio as it wasn't really necessary for the example and I updated the README file.

ChunyiLyu (approved):

PR #1037, opened 2022-05-12
Document how to set up tls verify for prometheus servicemonitor

### LINKED ISSUE #1007: Validate TLS cert when scraping RabbitMQ metrics
@ansd commented on [Tue May 11 2021](https://github.com/rabbitmq/tanzu-cluster-operator/issues/65)

When scraping RabbitMQ metrics via TLS, we currently advise to skip TLS validation in the ServiceMonitor ([here](https://github.com/rabbitmq/cluster-operator/blob/71074e811bcf036ae9a89ca0b5eebfce60b32e14/observability/prometheus/monitors/rabbitmq-servicemonitor.yml#L18)) and in the Prometheus config file ([here](
https://github.com/rabbitmq/cluster-operator/blob/71074e811bcf036ae9a89ca0b5eebfce60b32e14/observability/prometheus/config-file.yml#L119)).

We should validate the TLS server certificate.
To do so, we need to define the `tlsConfig` in the ServiceMonitor by adding the CA cert which signed the RabbitMQ Prometheus TLS cert. I manually tested that the following example works:
```yaml
apiVersion: cert-manager.io/v1
kind: ClusterIssuer
metadata:
  name: selfsigned-issuer
spec:
  selfSigned: {}
---
apiVersion: cert-manager.io/v1
kind: Certificate
metadata:
  name: selfsigned-cert
spec:
  dnsNames:
    - "*.rabbit-nodes.default.svc.cluster.local"
    - "*.rabbit.default.svc.cluster.local"
  secretName: tls-secret
  issuerRef:
    kind: ClusterIssuer
    name: selfsigned-issuer
---
kind: RabbitmqCluster
metadata:
  name: rabbit
spec:
  replicas: 1
  tls:
    secretName: tls-secret
---
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  name: rabbitmq
spec:
  endpoints:
  - interval: 15s
    port: prometheus
    scheme: http
    scrapeTimeout: 14s
  - interval: 15s
    port: prometheus-tls
    scheme: https
    scrapeTimeout: 14s
    tlsConfig:
      ca:
        secret:
          key: ca.crt
          name: tls-secret
      insecureSkipVerify: false
      serverName: '*.rabbit.default.svc.cluster.local'
  namespaceSelector:
    any: true
  selector:
    matchLabels:
      app.kubernetes.io/component: rabbitmq
```

We should make the Cluster Operator deploy one ServiceMonitor object per RabbitMQ Cluster (if Prometheus Operator is installed).

**Notes:**
* Above YAML is just an example.
* The Operator should deploy the ServiceMonitor with a single endpoint: `prometheus` if no TLS is enabled or `prometheus-tls` if TLS is enabled.
* `serverName` is needed so that we don't need to include IP addresses into the certificate.

ChunyiLyu:
This closes #1007 

## Summary Of Changes

Instead of adding a new example yaml, the tls section is added as comment section since the rest of the configuration is the same.

### DIFF
--- a/observability/prometheus/monitors/README.md
+++ b/observability/prometheus/monitors/README.md
@@ -25,7 +25,7 @@ metadata:
 
 Given the `matchLabels` fields from the Prometheus spec above, you would need to add the label `release: my-prometheus` to the `PodMonitor` and `ServiceMonitor` objects.
 
-File [rabbitmq-servicemonitor.yml](./rabbitmq-servicemonitor.yml) contains scrape targets for RabbitMQ.
+File [rabbitmq-servicemonitor.yml](./rabbitmq-servicemonitor.yml) contains scrape targets for RabbitMQ. TLS verify will be skipped by default. To enable TLS verification for scraping, set `spec.endpoints[port=prometheus-tls].tlsConfig.insecureSkipVerify` to false and provide a Kubernetes Secret containing CA cert used for Prometheus.
 Metrics listed in [RabbitMQ metrics](https://github.com/rabbitmq/rabbitmq-server/blob/master/deps/rabbitmq_prometheus/metrics.md) will be scraped from all RabbitMQ nodes.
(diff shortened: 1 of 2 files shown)


DanielePalaia (approved):
Looks good!

PR #1051, opened 2022-06-01
This PR override pod security context in system tests in case we are …

DanielePalaia:
…running on an openshift environment.

This is due in order to create a pipeline to run our system tests on openshift

This closes #

## Local Testing

```
```

```
$ kubectx dev-bunny
$ make destroy deploy-dev
$ make system-tests
```

### DIFF
--- a/system_tests/utils.go
+++ b/system_tests/utils.go
@@ -23,10 +23,11 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	controllerruntime "sigs.k8s.io/controller-runtime"
 	"strconv"
 	"strings"
 	"time"
+
+	controllerruntime "sigs.k8s.io/controller-runtime"
 
 	"gopkg.in/ini.v1"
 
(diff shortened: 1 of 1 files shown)


system_tests/utils.go
```diff
@@ -428,6 +433,28 @@ func newRabbitmqCluster(namespace, instanceName string) *rabbitmqv1beta1.Rabbitm
 	return cluster
 }
 
+func overrideSecurityContextForOpenshift(cluster *rabbitmqv1beta1.RabbitmqCluster) *rabbitmqv1beta1.RabbitmqCluster {
```
coro, line 436: I think this doesn't need to return the pointer that's passed in - you can just modify the RabbitmqCluster object referenced via the pointer and return nothing.

coro (approved):
Just a small comment

Zerpet (changes requested):
Content is right, but you made your commit with your gmail account, which may make OSPO unhappy. In addition to that, I'm strongly against long subject lines in a commit. Would you mind amending this commit with a better commit message?

Some guidelines and tips to commit messages: https://cbea.ms/git-commit/

DanielePalaia:
@Zerpet: Amended the author, put a better commit message!

PR #1059, opened 2022-06-08
Make GiB global

jiuker:
This closes #

## Local Testing

```
```

```
$ kubectx dev-bunny
$ make destroy deploy-dev
$ make system-tests
``` 
It seems better.


jiuker:
> I'm unsure why is better to have the constant at module level? The variable `GiB` is used only in the `removeHeadroom()` function. What's the benefit of having it at module level?
First of all, I think this is a unit of measurement. Many packages can share it. Secondly, it is strange to define a const variable in a function. If it is not defined as const, I think it is acceptable. So either define it as a global variable or remove the definition of const. I chose the former.

jiuker:
Maybe it's ok

PR #1065, opened 2022-06-10
Update PR workflow with latest k8s

Zerpet:

## Summary Of Changes

Update PR workflow to run on latest available Kubernetes version in KinD.

## Additional Context

Given that we support the latest k8s version since we merged #1044 and Kubernetes 1.21 is out of support since 1.24 became GA, we should update our workflow to test on a newer Kubernetes version.

Edit: 100% I will squash merge

## Local Testing

N/A change to GitHub Actions.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -6,6 +6,7 @@ on:
 
 env:
   GO_VERSION: 1.17.x
+  K8S_VERSION: v1.24.1
 
 jobs:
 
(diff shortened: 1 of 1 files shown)


Zerpet:
The errors in [the system tests](https://github.com/rabbitmq/cluster-operator/runs/6833834411?check_suite_focus=true) seem to be a legitimate upstream failure in RabbitMQ Server. I reached out to the folks in the Core about this. I think it should not block merging this PR, given that released versions of RabbitMQ work well in the new Kubernetes version.

DanielePalaia (approved):

ChunyiLyu (approved):

ChunyiLyu:
Love it! @Zerpet Would you mind doing it for messaging topology operator as well? It runs `1.21` k8s as well.

PR #1090, opened 2022-07-07
Add full URL entry to generated secret resource

### LINKED ISSUE #1089: Full AMQP url inside generated secret
Hi!

First of, thanks for the awesome work here, lots of problems solved with managing RabbitMQ myself.

In most charts/operators (i.e. Crunchy Postgres or Bitnami Cassandra) when generating a secret for the connection to the resource, besides separate config entries (username, password, host, port, ...) there's also the full URI to the resource, encapsulating the authentication and server domain part. The current secret has each of the specific entry as separate secret entries, so it would be neat to have the full AMQP URL also, in cases where libraries (notably, in my project it's `amqp-ts`) expect it.

I've already added it to my fork, added also unit tests, tomorrow I'm testing it on our dev cluster. So I'll open a PR with the additions.

Cheers!

penumbra23:
This closes #1089 

## Summary Of Changes
- added `url` entry to the generated secret resource (for example, `amqp://username:password@localhost:5672`)
- extended default user unit test to include the `url` field check

## Local Testing
Unit and integration tests pass

### DIFF
--- a/internal/resource/default_user_secret.go
+++ b/internal/resource/default_user_secret.go
@@ -56,6 +56,8 @@ func (builder *DefaultUserSecretBuilder) Build() (client.Object, error) {
 		return nil, err
 	}
 
+	url := fmt.Sprintf("amqp://%s:%s@%s", username, password, builder.Instance.ServiceSubDomain())
+
 	// Default user secret implements the service binding Provisioned Service
 	// See: https://k8s-service-bindings.github.io/spec/#provisioned-service
 	secret := &corev1.Secret{
@@ -71,6 +73,7 @@ func (builder *DefaultUserSecretBuilder) Build() (client.Object, error) {
 			"provider":          []byte(bindingProvider),
 			"type":              []byte(bindingType),
 			"host":              []byte(builder.Instance.ServiceSubDomain()),
+			"url":               []byte(url),
 		},
 	}
 	builder.updatePorts(secret)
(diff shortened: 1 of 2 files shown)


ablease:
Hi @penumbra23 thanks for this contribution!

I have verified that tests are passing, and the full url is present in the generated secret resource.

@ChunyiLyu unless you can think of anything else, I think we can merge this now.

ablease:
Hey @penumbra23, 

After some more consideration, and discussion with the rest of the team. We have decided not to merge this contribution. While technically there is nothing wrong with your suggestion, we find that it creates more problems than it solves. For example, amqp is the correct protocol to use in the most basic case, but what about when TLS is enabled? (AMQPS). RabbitMQ also supports many different protocols, STOMP and MQTT, both of which have their own TLS counterparts, and all of which use different ports.

Although in your original issue #1089 you mention that other charts/operators typicaly provide a full url field, we  feel that for RabbitMQ, this isn't really feasible. 

I appreciate that this means more work for you on the client side as you will need to construct your own url from the other fields in the secret.

I hope this makes sense, If you have any more questions do let us know, and thanks very much for your contribution.

penumbra23:
No problem, we internally use it for our cluster since AMQP is the only one required and the traffic is internal.

If the operator supports all the protocols, I could add logic for determining the right protocol or add multiple URLs for each of the protocol served (`amqp(s)`, `mqtt`, etc.). Does that make sense?

laurivosandi:
I also suggest adding different variable for each protocol. Assembling the URI from different env vars is clunky...

PR #1107, opened 2022-08-04
Update crd after k8s api and controller-gen bump

ChunyiLyu:
This closes #

## Summary Of Changes

Update crd after k8s api and controller-gen bump. We haven't regenerated the CRD after recent k8s api bump to 1.24.


coro (approved):
⚛️

PR #1110, opened 2022-08-11
Stop testing rmq 3.8

ChunyiLyu:
This closes #

**Note to reviewers:** remember to look at the commits in this PR and consider if they can be squashed

## Summary Of Changes

Stop testing rmq 3.8 as it has reached end of support

- remove 3.8.8 github action tests
- update README to say the Operator supports 3.9 onwards

PR #1111, opened 2022-08-15
Remove RabbitMQ 3.8 support

ansd:
Since we dropped support for RabbitMQ 3.8 in https://github.com/rabbitmq/cluster-operator/pull/1110, we can now simplify the cluster-operator:
1. Remove deprecated `cluster_formation.randomized_startup_delay_range` configurations such that warning `[warn] <0.130.0> cluster_formation.randomized_startup_delay_range.min and cluster_formation.randomized_startup_delay_range.max are deprecated` won't be output anymore in RabbitMQ logs. See https://github.com/rabbitmq/rabbitmq-server/pull/3075.
2. Use `rabbitmqctl enable_feature_flag all` command introduced by @mkuratczyk and @michaelklishin in https://github.com/rabbitmq/rabbitmq-server/commit/ce655864050a78f46005b8aa5e047394a6178cd6.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -53,7 +53,7 @@ jobs:
         k8s:
         - v1.24.1
         rabbitmq-image:
-        - rabbitmq:3.9-management
+        - rabbitmq:3.9.0-management # minimum supported version
         - rabbitmq:3.10-management
         - pivotalrabbitmq/rabbitmq:master-otp-min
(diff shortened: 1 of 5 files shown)


coro (approved):
Love it. Let's make sure to be explicit in our release notes for this cluster operator version then since we're now explicitly relying on behaviour in 3.9 rather than implicitly dropping support for 3.8.

Issue #958, opened 2022-02-10
RabbitMQ Node sometimes fails with nxdomain on startup


mstoetzer:
## Describe the bug
Not sure if this is a bug or just a question.
Sometimes I get random restarts of the RabbitMQ pods. When they restart, they produce following log:

This log is from e.g. `rabbitmq-server-1`:
```
2022-02-03 16:22:01.153148+00:00 [warn] <0.130.0> cluster_formation.randomized_startup_delay_range.min and cluster_formation.randomized_startup_delay_range.max are deprecated

BOOT FAILED
===========
2022-02-03 16:22:01.563990+00:00 [erro] <0.130.0>
2022-02-03 16:22:01.563990+00:00 [erro] <0.130.0> BOOT FAILED
2022-02-03 16:22:01.563990+00:00 [erro] <0.130.0> ===========
2022-02-03 16:22:01.563990+00:00 [erro] <0.130.0> ERROR: epmd error for host rabbitmq-server-1.rabbitmq-nodes.mynamespace: nxdomain (non-existing domain)
2022-02-03 16:22:01.563990+00:00 [erro] <0.130.0>
ERROR: epmd error for host rabbitmq-server-1.rabbitmq-nodes.mynamespace: nxdomain (non-existing domain)

2022-02-03 16:22:02.564897+00:00 [erro] <0.130.0>     supervisor: {local,rabbit_prelaunch_sup}
2022-02-03 16:22:02.564897+00:00 [erro] <0.130.0>     errorContext: start_error
2022-02-03 16:22:02.564897+00:00 [erro] <0.130.0>     reason: {epmd_error,"rabbitmq-server-1.rabbitmq-nodes.mynamespace",
2022-02-03 16:22:02.564897+00:00 [erro] <0.130.0>                         nxdomain}
2022-02-03 16:22:02.564897+00:00 [erro] <0.130.0>     offender: [{pid,undefined},
2022-02-03 16:22:02.564897+00:00 [erro] <0.130.0>                {id,prelaunch},
2022-02-03 16:22:02.564897+00:00 [erro] <0.130.0>                {mfargs,{rabbit_prelaunch,run_prelaunch_first_phase,[]}},
2022-02-03 16:22:02.564897+00:00 [erro] <0.130.0>                {restart_type,transient},
2022-02-03 16:22:02.564897+00:00 [erro] <0.130.0>                {significant,false},
[...]

Issue #752, opened 2021-06-30
RabbitMQ Cluster failed with 0/2 nodes are available: 2 pod has unbound immediate PersistentVolumeClaims on Kubernetes


sathishc58:
I am trying to install RabbitMQ Cluster Operator and RabbitMQ Cluster on K8S Cluster (bare metal server) by following the steps mentioned below

OS Version:  CentOS Linux 7 (Core)
[root@re-ctrl01 tmp]# kubectl version
```
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.0", GitCommit:"cb303e613a121a29364f75cc67d3d580833a7479", GitTreeState:"clean", BuildDate:"2021-04-08T16:31:21Z", GoVersion:"go1.16.1", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.0", GitCommit:"cb303e613a121a29364f75cc67d3d580833a7479", GitTreeState:"clean", BuildDate:"2021-04-08T16:25:06Z", GoVersion:"go1.16.1", Compiler:"gc", Platform:"linux/amd64"}
```

[root@re-ctrl01 tmp]# kubectl apply -f https://github.com/rabbitmq/cluster-operator/releases/latest/download/cluster-operator.yml

```namespace/rabbitmq-system created
Warning: Detected changes to resource rabbitmqclusters.rabbitmq.com which is currently being deleted.
customresourcedefinition.apiextensions.k8s.io/rabbitmqclusters.rabbitmq.com configured
serviceaccount/rabbitmq-cluster-operator created
role.rbac.authorization.k8s.io/rabbitmq-cluster-leader-election-role created
clusterrole.rbac.authorization.k8s.io/rabbitmq-cluster-operator-role created
rolebinding.rbac.authorization.k8s.io/rabbitmq-cluster-leader-election-rolebinding created
clusterrolebinding.rbac.authorization.k8s.io/rabbitmq-cluster-operator-rolebinding created
deployment.apps/rabbitmq-cluster-operator created
```

[root@re-ctrl01 tmp]# kubectl describe pod/test-rabbitmq-cluster-server-0 -n rabbitmq-system

```
Name:           test-rabbitmq-cluster-server-0
Namespace:      rabbitmq-system
Priority:       0
Node:           <none>
Labels:         app.kubernetes.io/component=rabbitmq
[...]

Issue #269, opened 2020-08-25
Using RabbitMQ Cluster Kubernetes Operator step by step but failed to get pod ready


Darwinarch:
## Describe the bug
Hi, 
      When I followed [the instruction online](https://www.rabbitmq.com/kubernetes/operator/using-operator.html) step by step, I could not make the pods ready !
![11](https://user-images.githubusercontent.com/69334027/91129026-ddb3dd00-e6db-11ea-9a6d-7b435a495da7.jpg)
      After I described the pod, 
      `kubectl describe po definition-rabbitmq-server-0`  
      I got the following message :
![22](https://user-images.githubusercontent.com/69334027/91129743-1a340880-e6dd-11ea-8767-6992c0134656.jpg)
       and the log told me that :
![66](https://user-images.githubusercontent.com/69334027/91150555-1e692180-e6ef-11ea-853c-52bca7c5fbba.jpg)
      my definition.yaml :
![33](https://user-images.githubusercontent.com/69334027/91130116-ba8a2d00-e6dd-11ea-87de-778c6458329e.jpg)
      my cluster info : only devd02 has PV
![55](https://user-images.githubusercontent.com/69334027/91130935-07bace80-e6df-11ea-98ca-f7efa85e87ea.jpg)
       
     I don't know what's wrong in my setup and I need some help.  Thanks a lot !

ansd:
Hey @Darwinarch,

thanks for reaching out and for the detailed screenshots.

Did you create the file `rabbitmq-headless.yml` manually?
The selector `app: rabbitmq` looks suspicious therein.

What should happen instead is that the cluster operator creates a headless service (namely the `service/definition-rabbitmq-headless` output in your first screenshot) adding a label `app.kubernetes.io/name: definition` to it.

Darwinarch:
Hello @ansd  ,

I didn't create the rabbitmq-headless.yaml by myself. 
Should I create it according to [here](https://www.rabbitmq.com/blog/2020/08/10/deploying-rabbitmq-to-kubernetes-whats-involved/) ?

In addition, `service/definition-rabbitmq-headless` is generated automatically after I key in `kubectl apply -f definition.yaml`.
When I described `definition-rabbitmq-headless`, `app.kubernetes.io/name: definition` had been labeled as follow:
![1](https://user-images.githubusercontent.com/69334027/91250219-0f37b180-e78b-11ea-9539-fdf2a08ad8af.jpg)

[...]

PR #1156, opened 2022-10-07
Delay start by 30 seconds

ansd:
The rabbitmq/cluster-operator relies on https://github.com/rabbitmq/cluster-operator/blob/8f4735b5e0f9050bb1028e635c1fba080d57412d/internal/resource/headless_service.go#L72

in its headless service for peer discovery such that the RabbitMQ cluster gets formed before the Pods are being marked as `Ready`.

However, as explained in https://github.com/kubernetes/kubernetes/issues/92559 this promise is broken in Kubernetes, specifically in CoreDNS: By default DNS entries of the Pods are not updated correctly within the first 30 seconds.

These networking issues early during RabbitMQ boot cause a series of issues:
* https://github.com/rabbitmq/cluster-operator/issues/958
* https://github.com/rabbitmq/cluster-operator/issues/269
* https://github.com/rabbitmq/cluster-operator/issues/752#issuecomment-872118881
* https://github.com/rabbitmq/cluster-operator/issues/298#issuecomment-760153567
* https://github.com/rabbitmq/rabbitmq-server/issues/5322
* https://github.com/rabbitmq/rabbitmq-server/issues/6034

In some of above issues, the solution taken was to retry N times within RabbitMQ during the boot process.
In other cases simply retrying within RabbitMQ is not even possible because things get stuck due to bugs in Erlang's `global` module (more details in above issues).

Instead of using brittle workarounds within RabbitMQ, I suggest we only start RabbitMQ once we know that the Pods' DNS entries are up-to-date.

How can we ensure that the Pods' DNS entries are up-to-date?

One possibility is to query the DNS service in the init container as demonstrated in https://github.com/rabbitmq/rabbitmq-server/issues/5322#issuecomment-1195826135. However, since some DNS services use a K8s Service object in front, it's not certain that all the DNS backend pods are queried. So, the init container could falsely get positive results while during RabbitMQ boot process, other DNS servers with outdated caches are queried.

Therefore, to make things simple, I suggest the new default will be to artificially delay starting RabbitMQ by 30 seconds. I chose 30 seconds because that's CoreDNS default `cache` value. It's an inelegant solution which I've tried to avoid until now. However, **reliable** rolling updates (and cluster formations) are more important than 30 seconds more downtime when a node is being updated.

This new value will be configurable via `spec.DelayStartSeconds`. Users who never experienced RabbitMQ boot issues (for example because they use DNS backends which actually respect `publishNotReadyAddresses=True` or because they explicitly decrease CoreDNS default `cache` value) can safely reduce this artificial delay or disable it by setting the value to `0`.

I tested 50 5-node cluster formations and rolling updates. With the new 30 seconds delay between creating the Pod and starting RabbitMQ, I never saw any clustering or upgrade issues.


config/crd/bases/rabbitmq.com_rabbitmqclusters.yaml
```diff
@@ -10,7 +10,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.9.2
+    controller-gen.kubebuilder.io/version: v0.10.0
```
MirahImage, line 13: Did you mean to bump the kubebuilder version?

ansd: Not intentionally.
`kubebuilder` seems to come from `controller-gen` which in turn comes from `controller-tools` https://github.com/rabbitmq/cluster-operator/blob/8f4735b5e0f9050bb1028e635c1fba080d57412d/tools/tools.go#L11 which in turn is checked in with `v0.10.0` in https://github.com/rabbitmq/cluster-operator/blob/8f4735b5e0f9050bb1028e635c1fba080d57412d/go.mod#L28

I can revert that line?

ansd: That was bumped 3 days ago via https://github.com/rabbitmq/cluster-operator/pull/1146
It seems I'm the first one who ran `make manifest` since then. So, it seems correct to bump kubebuilder version as done in this commit?

ansd:
Without this PR, I sometimes see the following in the logs of a node being booted during a rolling update:
```
2022-10-10 09:51:52.533123+00:00 [debug] <0.606.0> Getting epmd port node 'rabbit', 10 retries left
2022-10-10 09:54:07.004227+00:00 [debug] <0.606.0> Getting epmd port node 'rabbit', 9 retries left
```
... more than 2 minutes delay. In that regards this PR makes booting even faster.

PR #1180, opened 2022-11-04
Update pod-disruption-budget.yaml

BostjanPisler:
Updated apiVersion to policy/v1


DanielePalaia:
@BostjanPisler thank you!

Issue #1003, opened 2022-03-29
manage rabbitmq version with operator


steinarox:
**Is your feature request related to a problem? Please describe.**
To simplify rabbitmq management for simple usecases, it would be nice it rabbitmq-operator could take control over the versions deployed.

This code: https://github.com/rabbitmq/cluster-operator/blob/b853bb66bf36bc49c7872c365c1e642dfb9e565b/controllers/rabbitmqcluster_controller.go#L120
Skips setting the image version if already set.  

**Describe the solution you'd like**
A feature flag that will leave the image versions for connected rabbitmq clusters up to the operator. It will be the admins resposibility that the new default version can work without any migration. 
Maybe a feature flag like `CONTROL_RABBITMQ_IMAGE=true`, false is default

It should not override any rabbit ocontrols like the `pauseReconciliation`annotation. 

**Additional context**
This will simplify operation for smaller/simpler deployments and demos. running operator v.1.12.1

mkuratczyk:
Do you mean the Operator would update the `image` value to automatically upgrade RabbitMQ? If you just want your instance to be always using the latest version, you can use an image tag that isn't too specific, for example there is `rabbitmq:3.9`. Add an `imagePullPolicy: Always` (through overrides) and you will always deploy the most recent 3.9.

Zerpet:
The challenge here is that Cluster Operator does not monitor new versions of RabbitMQ i.e. it does not poll the registry in `image`. This proposal sounds specific to Cluster Operator upgrades (as in it should also upgrade all rabbits). 

You can achieve something similar if you don't specify an `image` in your manifest e.g.

```yaml
apiVersion: rabbitmq.com/v1beta1
kind: RabbitmqCluster
metadata:
  name: custom-configuration
spec:
  replicas: 1
```

Everytime you apply this manifest, Kubernetes receives empty values for the properties you have not specified, therefore entering the conditional you quoted. Since we set a patch version of the image, it will apply the new default that comes with the Cluster Operator's specific version.

steinarox:
@mkuratczyk if you can manually restart/rollout the statefulset that will work, but i would like to avoid manual cluster operations.

@Zerpet  the code that i linked to should work the way you describe. We are not adding image field in our definitions, however the image field gets patched into the definitions so in practice this does not work at the moment. If the intent is to work as you say, it seems like there is a bug in the controller somewhere else that persists the image property even if it should not.

ChunyiLyu:
[...]

PR #1181, opened 2022-11-07
let rabbitmq operator to control images

farodin91:
Part of #1003 

## Summary Of Changes

Add experimental ENV to let rabbitmq operator update images. **No safety checks at the moment.**

## Local Testing

```
```

```
$ kubectx dev-bunny
$ make destroy deploy-dev
$ make system-tests
```

### DIFF
--- a/api/v1beta1/rabbitmqcluster_types.go
+++ b/api/v1beta1/rabbitmqcluster_types.go
@@ -449,8 +449,8 @@ func (cluster *RabbitmqCluster) ExternalSecretEnabled() bool {
 	return cluster.Spec.SecretBackend.ExternalSecret.Name != ""
 }
 
-func (cluster *RabbitmqCluster) UsesDefaultUserUpdaterImage() bool {
-	return cluster.VaultEnabled() && cluster.Spec.SecretBackend.Vault.DefaultUserUpdaterImage == nil
+func (cluster *RabbitmqCluster) UsesDefaultUserUpdaterImage(controlRabbitmqImage bool) bool {
+	return cluster.VaultEnabled() && (cluster.Spec.SecretBackend.Vault.DefaultUserUpdaterImage == nil || controlRabbitmqImage)
 }
 
 func (cluster *RabbitmqCluster) VaultDefaultUserSecretEnabled() bool {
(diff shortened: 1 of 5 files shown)


DanielePalaia:
Hi @farodin91  Can you explain to us more about your use case and how this functionality can help you?
Maybe opening a new issue and linking it to this PR will be nice too.

In general we are not against it but using it improperly can cause some unwanted consequences like a rolling restart of all the rabbitmq clusters whenever an operator upgrade happens.

I think the scope of this new variable should be well documented at least in the code if not maybe even in our doc.

johanneswuerbach:
@DanielePalaia we would also be interested in this functionality mainly to reduce operational burden and only keep the operator up-to-date instead of maintaining the operator and cluster(s).

> rolling restart of all the rabbitmq clusters whenever an operator upgrade happens

We generally have a 1:1 mapping, so for us that would be fine. An option might be to generally limit how many cluster are rolled in parallel, would this be sufficient?

farodin91:
@DanielePalaia I will try to write down a proposal. 

This should be just the first step. 
Ideas:
1. adding checks to update only proper version.
2. a map of support version with the newest patch level to support upgrade of multiple versions
3. rate limit cluster upgrade speed.

First of all, I will try to add a small documenation.

DanielePalaia:
Hi @johanneswuerbach @farodin91 yes thank you, I think this should be better documented on the source code as well as maybe describe this future as *experimental* somewhere in our doc and describe the purpose of it. As first step I suggest anyway to open an issue and link this PR to it so we can discuss it there.

farodin91:
@DanielePalaia the update to documentation have to be done in a different repo, i guess?

DanielePalaia (approved):

DanielePalaia:
Hi @farodin91 we have this page to document our environment variables: https://www.rabbitmq.com/kubernetes/operator/configure-operator-defaults.html#parameters.

But as this is just a first experimental approach just useful for "testing" purpose, maybe we can document it when the functionality will be fully ready (at least to provide a version validity check).

Anyway if you don't have other modifications to be done in this PR I will integrate it.

farodin91:
@DanielePalaia Should I open an PR to the website repository before this get merged?

farodin91:
We are done here.

DanielePalaia:
Thanks @farodin91 !

PR #1199, opened 2022-11-24
External Secret: Implementation to skip the creation of the default u…

DanielePalaia:
…ser credential secret and give the possibility to use an external one

This closes #

## Summary Of Changes

This changement is adding a new field in the SecretBackend struct of the rabbitmqcluster crd  to avoid creating the credential secret for the default user automatically. A secret can be provisioned externally to the operator and be used by the operator using this field.
Pods will remain in init mode till the secret will be provisioned.
It can be useful when an external secret tool (like external secret operator) is used to export a secret from a cloud provider secret management system (like google secret manager).

## Additional Context

This can be a first step in order to improve the external secrets management.

## Local Testing

Tested both the operator and topology operator system tests against this modification

```
```

```
$ kubectx dev-bunny
$ make destroy deploy-dev
$ make system-tests
```


api/v1beta1/rabbitmqcluster_types.go
```diff
@@ -104,7 +104,8 @@ type RabbitmqClusterSpec struct {
 // Future secret backends could be Secrets Store CSI Driver.
 // If not configured, K8s Secrets will be used.
 type SecretBackend struct {
-	Vault *VaultSpec `json:"vault,omitempty"`
+	Vault          *VaultSpec `json:"vault,omitempty"`
+	ExternalSecret string     `json:"externalSecret,omitempty"`
```
coro, line 108: Looks like this later gets translated into a LocalObjectReference. Could this type just be a LocalObjectReference upfront? I'd be keen to find out if we get more upfront validation if we were to do that.

coro: By upfront validation, I mean would the Kubernetes API reject the request, rather than making it all the way to the controller before rejecting it.

DanielePalaia: Hi @coro LocalObjectReference looks good! I modified the type!

coro (approved):

PR #1205, opened 2022-12-08
force boot rabbitmq pod because mnesia tables times out

lanzhiwang:
This closes #

## Local Testing

```
```

```
$ kubectx dev-bunny
$ make destroy deploy-dev
$ make system-tests
```

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -67,6 +67,14 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - pods/log
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - ""
+  resources:
   - secrets
   verbs:
   - create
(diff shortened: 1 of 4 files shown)


michaelklishin:
It's a really bad idea to force boot a node like that. Force boot must only be used when the operator (a human) really understands that it is the only way to make a node boot.

This message is [documented in the docs on node restarts](https://www.rabbitmq.com/clustering.html#restarting). When you see this on Kubernetes? Almost certainly
because a health check that requires a fully booted node [is used as a readiness probe](https://www.rabbitmq.com/cluster-formation.html#peer-discovery-k8s).

The solution is not to force boot nodes but to use a basic health check for readiness probe.

michaelklishin:
Why not force boot nodes? A force booted node won't sync any data from its peers, by definition. It is the last resort cluster recovery mechanism when a majority of nodes is permanently lost.

Force booting all cluster nodes is a sure way to eventually lose data.

PR #1237, opened 2023-01-23
Add StatefulSet.Spec.PersistentVolumeClaimRetentionPolicy Override.

### LINKED ISSUE #1231: Allow persistentVolumeClaimRetentionPolicy setting for RabbitMQ cluster statefulSet
This option has been available since the Kubernetes v1.23 and it would be really helpful with AWS installations, where PVs get automatically deleted when PVC gets deleted. And that's what RabbitMQ Operator does. It automatically deletes the PVCs. 

This is how it would be set up:
```
spec:
  replicas: 3     
  override:
    statefulSet:
      spec:
        persistentVolumeClaimRetentionPolicy: 
          whenDeleted: Retain
          whenScaled: Delete
```

If we use it like this though,  we get this error:
`Error: INSTALLATION FAILED: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(RabbitmqCluster.spec.override.statefulSet.spec): unknown field "persistentVolumeClaimRetentionPolicy" in com.rabbitmq.v1beta1.RabbitmqCluster.spec.override.statefulSet.spec`

The workaround is to set the StorageClass retention policy to "Retain" but then it's hard to re-attach such volume after the reinstallation of the RabbitMQ cluster as AWS doesn't allow the usage of  a claim Selector. Thus PV stays in the "released" state while the new empty one is created for the new PVC.
image

I think there are plenty of K8s AWS users to justify this being a bit more of a priority issue. I have already seen this being requested here but I tried explaining it more clearly why it's necessary.

MirahImage:
This closes #1231 

## Summary Of Changes
Adds persistentVolumeClaimRetentionPolicy to the StatefulSet.Spec override.


DanielePalaia (approved):

Dragane:
Thank you so much for working on this solution Mirah. 

I would like to point out the bug (typo) in the CRDS file `config/crd/bases/rabbitmq.com_rabbitmqclusters.yaml`

There is a letter t missing:
`persistenVolumeClaimRetentionPolicy` instead of `persistentVolumeClaimRetentionPolicy`

MirahImage:
Thanks for catching that!

PR #1251, opened 2023-02-01
Set patch version to avoid updating existing RMQs

ChunyiLyu:
This relates to #1247, which is only in `main` branch and now released yet.

## Summary Of Changes

With a tag like `3.11` which moves every time a new patch has been released, existing RMQ could be updated unintentionally. 

## Local Testing

```
```

```
$ kubectx dev-bunny
$ make destroy deploy-dev
$ make system-tests
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ under `site/kubernetes`.
 
 ## Supported Versions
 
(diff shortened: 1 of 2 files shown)


Zerpet:
Maybe we should set it to `3.11.7` for now, due to this https://github.com/docker-library/rabbitmq/issues/603

ChunyiLyu:
@Zerpet Updated to `3.11.7`

Zerpet (approved):

PR #1269, opened 2023-02-21
Move verbose log lines to debug

ChunyiLyu:

## Summary Of Changes

- Update generated CRD
- log cluster spec at debug instead of info
- log message about requeue request to enable plugin at debug instead of info
- log successfully queue rebalance like other cli command

'Start reconciling' log which used to contain `.spec` now looks like
`2023-02-21T17:35:40Z	INFO	Start reconciling	{"controller": "rabbitmqcluster", "controllerGroup": "rabbitmq.com", "controllerKind": "RabbitmqCluster", "RabbitmqCluster": {"name":"sample","namespace":"rabbitmq-system"}, "namespace": "rabbitmq-system", "name": "sample", "reconcileID": "b8013f6d-40a3-45db-9b57-77e21e56d2a2"}`

### DIFF
--- a/config/crd/bases/rabbitmq.com_rabbitmqclusters.yaml
+++ b/config/crd/bases/rabbitmq.com_rabbitmqclusters.yaml
@@ -10,7 +10,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.10.0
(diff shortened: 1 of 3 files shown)


MirahImage (approved):
Thanks! I was undecided at first whether the requeue bit should be debug, but after thinking about it more, it's definitely an internal detail and therefore debug level makes sense.

Zerpet (approved):

PR #1365, opened 2023-05-30
Bump controller runtime and k8s api 1.27

ChunyiLyu:

## Summary Of Changes

- Bump controller runtime to `v0.15` and k8s api to `1.27`
- Add kinD local system tests against k8s 1.19 cluster for PRs to ensure compatibility with k8s 1.19


.github/workflows/build-test-publish.yml
```diff
@@ -17,6 +17,7 @@ env:
   # Taken from https://github.com/kubernetes-sigs/kind/releases/tag/v0.18.0
   # The image here should be listed under 'Images built for this release' for the version of kind in go.mod
   KIND_NODE_IMAGE: "kindest/node:v1.26.3@sha256:61b92f38dff6ccc29969e7aa154d34e38b89443af1a2c14e6cfbd2df6419c66f"
+  KIND_119_NODE_IMAGE: "kindest/node:v1.19.16@sha256:81f552397c1e6c1f293f967ecb1344d8857613fb978f963c30e907c32f598467"
```
MirahImage, line 20: Would it make sense to call this `KIND_OLDEST_NODE_IMAGE` or something similar? Then, if we decide to make a change to support only K8s versions from some newer release, we won't have to rename this.

MirahImage (approved):

.github/workflows/build-test-publish.yml
ChunyiLyu: Updated

PR #1430, opened 2023-08-19
Renamed patches -> patchesStrategicMerge

preved911:
Hello. I can't use this repo kustomize templates, because "patches:" notation was obsolete. I get errors, when try to execute kustomize:
```
error: invalid Kustomization: json: cannot unmarshal string into Go struct field Kustomization.patches of type types.Patch
```
Could you merge this little PR to fix it?


Zerpet (approved):
Thank you! Will merge after CI passes.

PR #1455, opened 2023-09-27
feat(ipv6): add support for IPv6 env

simonwu-sn:

PR #1474, opened 2023-10-18
Support for Erlang INET configuration

Zerpet:

## Summary Of Changes

- Support for configuring Erlang INET configuration
- Support for configuring IP family policy in Service
- Example to configure IPv6

## Additional Context

Originally inspired by #1455. Main motivation is to ease the configuration of Erlang
to use IPv6

## Local Testing

There are unit and integration tests. There are no system tests as IPv6 requires infrastructure
support for IPv6, which is not guaranteed.


Zerpet:
The check that failed is a system test on the `main` branch of RabbitMQ server. This is fine because the main branch is going through a number of changes for 3.13 and some for 4.0, and it's expected to break some of our tests. The system tests pass in 3.9.9 (EOL'ed, but absolute minimum required by the Operator) and in `rabbitmq:management`, which is 3.12's latest patch.

PR #1494, opened 2023-11-23
imagePullSecrets is an array

mkuratczyk:
Without the fix, passing `--image-pull-secret` would lead to validation failure.


Zerpet:
We should change CI so that changes to the `kubectl` plugin don't trigger the entire system test suite... One day Soon™ 😏

PR #1506, opened 2023-12-12
Reverting docker/build-push-action version bump due to a known issue

PujaVad:
https://github.com/docker/build-push-action/issues/820


DanielePalaia:
Hi @PujaVad thanks. I think this can be fine as temporary solution in order to release new valid versions for amd64 and arm64 for the next commercial releases but I think in the long run we may need a definitive solution (like put provenance: false as suggested in the issue), otherwise we will never be able to upgrade and we will continue to receive the dependabot upgrade which could be integrated by mistake. Maybe we can start opening an issue about this that can be investigated next

DanielePalaia (approved):

PujaVad:
@DanielePalaia , you're absolutely right, this is only temporary, will create a separate issue to test with other options.

PR #1544, opened 2024-01-22
Correct annotations for certificates generated via vault intermediate CA

### LINKED ISSUE #1540: Vault secret backend templates break when using certs issued by an intermediate CA
## Describe the bug

When using the vault secret backend to retrieve server certificates, if the certificate is generated using an intermediate CA, the default templates do not produce the right annotations to render all the certificates required to validate the chain.

In this case, I can get the correct behaviour with the following two overrides:

```
        vault.hashicorp.com/agent-inject-template-ca.crt: |
          {{- with secret "<vault-root-ca-cert-path>" -}}
          {{- .Data.certificate -}}
          {{- end -}}
        vault.hashicorp.com/agent-inject-template-tls.crt: | 
          {{- with secret "<vault-intermediate-cert-issue-path>"
          "common_name=<template-common-names>"
          "alt_names=<template-alt-names>"
          "ip_sans=<template-ip-sans>-" -}}
          {{- .Data.certificate -}}
          {{- if .Data.ca_chain -}}
          {{- $lastintermediatecertindex := len .Data.ca_chain | subtract 1 -}}
          {{ range $index, $cacert := .Data.ca_chain }}
          {{ if (lt $index $lastintermediatecertindex) }}
          {{ $cacert }}
          {{ end }}
          {{ end }}
          {{- end -}}
          {{- end -}}      
```

However this is less than ideal because it includes other settings such as the requested alt_names which we do not want to override.

Out of these overrides, the updated tls.crt template will work whether the server cert is generated from an intermediate or not.
The ca,crt override is required to request the root certificate from a different path than the intermediate, which would mean an additional config value in the vault section that could be left blank in non-intermediate situations.

Would you be receptive to a PR to enable this scenario?

mr-miles:
This closes #1540 

## Summary Of Changes

 - Add an optional PKIRootPath field to RabbitMqCluster.SecretsBackend.Vault.VaultTLSSpec
 - Generate vault server certificate templates that handle intermediate certificates
 - Add to unit tests

## Local Testing

```
```

```
$ kubectx dev-bunny
$ make destroy deploy-dev
$ make system-tests
```


Zerpet:
hey @mr-miles, thank you for this contribution. This looks good to me as is. Do you have any further work in mind before marking this as ready for review?

Zerpet:
CI failed in `go vet ./...`

```
go vet ./...
# github.com/rabbitmq/cluster-operator/v2/api/v1beta1
Error: vet: api/v1beta1/rabbitmqcluster_types_test.go:454:27: fetchedRabbit.SecretBackend undefined (type *RabbitmqCluster has no field or method SecretBackend)
```

mr-miles:
HI @Zerpet 

Sorry, needed to rebuild my laptop halfway through so it was not an easy ride for such a small change!  Tests now all pass so I am marking this as ready for review.

Thanks for following up!

PR #1565, opened 2024-02-22
removing encoded rabbitmq-system namespace in system-tests

DanielePalaia:
This will allow us to run the system-tests in a different namespace than the default "rabbitmq-system"


Zerpet (approved):
Yes, please!

PR #1597, opened 2024-03-22
Deploy 3.13 by default

mkuratczyk:
Bump the default to 3.13.1. Do not mention some old RMQ and Kubernetes versions that likely work, but we surely don't test new Operator versions with them

### DIFF
--- a/.github/workflows/build-test-publish.yml
+++ b/.github/workflows/build-test-publish.yml
@@ -234,7 +234,7 @@ jobs:
     strategy:
       matrix:
         rabbitmq-image:
-        - rabbitmq:3.9.9-management
+        - rabbitmq:3.11.0-management
         - rabbitmq:management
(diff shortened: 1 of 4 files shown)


MirahImage (approved):

Zerpet (commented):
We test in Kubernetes 1.19 and 1.26. We can bump the "oldest supported" Kubernetes version in CI here:

https://github.com/rabbitmq/cluster-operator/blob/a7f46d9bb0455e117fddce5325ec20becbe45f7f/.github/workflows/build-test-publish.yml#L17-L20

I'm happy to drop support for older versions of Kubernetes to N-3, or whatever TKG ships, or something similar.

mkuratczyk:
I've bumped the tested Kubernetes versions from 1.19+1.26 to 1.26+1.29

PR #1603, opened 2024-04-04
small fix on test_and_publish_OLM_bundle

DanielePalaia:
Small fix on test_and_publish_OLM_bundle action workflow.

There is a small issue during the last release: 
https://github.com/rabbitmq/cluster-operator/actions/runs/8539313818/job/23393756084

Basically the sh shell doesn't recognize properly the line:
BUNDLE_VERSION=${BUNDLE_VERSION:1}

giving a bad substitution error.

This should fix this issue.

This PR is also bumping GO to 1.22 and golang.org/x/net to v0.23.0 in order to avoid the new vulnerability issue detected in https://github.com/rabbitmq/cluster-operator/actions/runs/8552381205/job/23433359006

It is also adding some missing crd fields to our CRD after a make manifests.
It maybe related to  https://github.com/rabbitmq/cluster-operator/commit/be674d6b3e5632ccd335360ee87821eaeea0e828

### DIFF
--- a/.github/workflows/testing_and_publishing_OLM_bundle.yml
+++ b/.github/workflows/testing_and_publishing_OLM_bundle.yml
@@ -42,6 +42,7 @@ jobs:
           run: | 
             BUNDLE_VERSION=${GITHUB_REF#refs/*/} 
             echo "BUNDLE_VERSION=${BUNDLE_VERSION:1}" >> $GITHUB_ENV
+          shell: bash
 
         - name: Set tag image for test version
           if: startsWith(github.ref, 'refs/tags/v') == false


Zerpet (commented):
I don't oppose explicitly to these changes, but a few observations:

Bumping to Go 1.22 may break some people. We found a user in the past who was consuming this repo as a Go library. The same applies for the Topology Operator, it may break it, as Go is becoming more strict with the `go` directive in `go.mod`, so that you don't consume modules from higher library versions. The fix seems possible in Go 1.21, so I don't think it's necessary to bump to 1.22.

There are 3 arguably related changes in this PR. Whilst all the changes are valid, I think it's better to submit separate PRs for each of them. Not only it helps with understanding the reason behind each change, it also makes it easier to revert a specific change, shall the need arise. As it stands, reverting this PR would revert 3 changes, whereas we may only want to revert one (in the hypothetical scenario that we wanted to revert)

DanielePalaia:
@Zerpet Yes I agree with your point. I was thinking the same. I will submit 3 PRs then and I will keep this one for the original fix related to OLM. For what concerns the vulnerability check I originally tried to bump just golang.org/x/net to v0.23.0 and not golang but apparently it wasn't enough. I will try to do other tests on the new Prs

PR #1604, opened 2024-04-04
updating cluster operator crd

DanielePalaia:
The cluster operator crd is currently not updated. 

make manifests is adding the pkiRootPath field. 
This was probably a step missing in:

https://github.com/rabbitmq/cluster-operator/commit/be674d6b3e5632ccd335360ee87821eaeea0e828


Zerpet (approved):
Good catch!

mr-miles:
Good spot (also just bitten by it and found same cause) - apologies it was missing from that PR, I didn't know to add it there!  I am wondering if there is a step or test we could add to generate the crds and confirm the files match what's checked in?  That would avoid it coming up again

PR #1605, opened 2024-04-04
bumping golang to 1.21.9

DanielePalaia:
Bumping golang to 1.21.9 to fix di vulnerability check

https://github.com/rabbitmq/cluster-operator/actions/runs/8552381205/job/23433359006


Zerpet:
I recall `govulncheck` complaining about `golang.org/x/net`, you may have to bump this dependency explicitly. In #1603 it also failed on `golang.org/net/http` IIRC, although that may be fixed simply with the Go patch update.

Zerpet:
If you want faster feedback loops, you can install Go 1.21 locally, install `govulncheck` and run the same command locally. There should be no reason to get a different behaviour than in CI.

DanielePalaia:
@Zerpet yes that's strange, this branch works fine for me locally with just bumping the x/net library to 0.23.0 with go 1.21.x (I tried 1.21.5 and the last 1.21.9). Not sure why in the action is complaining (even though it seems then working with 1.22 as we say in the last PR)

DanielePalaia:
From the error message:
  Standard library
    Found in: net/http@go1.21.8
    Fixed in: net/http@go1.21.9

and the vuln report https://pkg.go.dev/vuln/GO-2024-2687 it seems like it is affecting previous version of 1.21.9.

Maybe the github action is not taking the last 1.21.9 and still using 1.21.8?

DanielePalaia:
Yes it was a caching issue apparently. Cleaning up the cache solved the issue.

I will close this one and merge this identical one just arrived then: https://github.com/rabbitmq/cluster-operator/pull/1606

PR #1612, opened 2024-04-09
Don't restart the cluster after scale-out

### LINKED ISSUE #1610: Adding replicas triggers a rolling restart
Scaling out a cluster, eg. changing `replicas` from `3` to `5` should not trigger a rolling restart. It didn't in the past, most likely the restart is caused by the introduction of target cluster size hint:
https://github.com/rabbitmq/cluster-operator/pull/1113

If the only thing that changed is the `replicas` field, we should ship the restart.

mkuratczyk:
This closes #1610

### DIFF
--- a/internal/resource/configmap.go
+++ b/internal/resource/configmap.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/rabbitmq/cluster-operator/v2/internal/metadata"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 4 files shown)


internal/resource/configmap.go
```diff
@@ -247,6 +250,43 @@ func (builder *ServerConfigMapBuilder) Update(object client.Object) error {
 		return fmt.Errorf("failed setting controller reference: %w", err)
 	}
 
+	updatedConfigMap := configMap.DeepCopy()
+	if err := removeConfigNotRequiringNodeRestart(previousConfigMap); err != nil {
+		return err
+	}
+	if err := removeConfigNotRequiringNodeRestart(updatedConfigMap); err != nil {
+		return err
+	}
```
Zerpet, line 259: I'm a fan of wrapping the error with `fmt.Errorf("some context: %w", err)`, and provide some context as what we were trying to do; in this case, removing a property from the config map. Otherwise, a low level error from `Write` or `ini.Load` will bubble up and get logged in the operator logs, with no context as where this error comes from. It's a great practice to provide a bit of context around what we were trying to do, before bubbling the error.

mkuratczyk: Good point. Done

Zerpet (approved):

PR #1631, opened 2024-05-10
Revert docker/build-push-action version due to known issue

PujaVad:
Version 5 of docker/build-push-action produces `unknown` architecture and os in manifest


Zerpet (approved):
Should we leave a comment in the YAML so that this bump doesn't happen again?

PR #1647, opened 2024-06-03
force string for resource attributes in examples

nicoangelo:

## Summary Of Changes

This fixes the datatype of the resource attributes in the production ready example. Otherwise this example will cause CD tools like Flux, to constantly update the resource and change it back to a numeric value.
Since the `production-ready` example is rather prominently mentioned in the docs, I think it might make sense to provide an example that doesn't trigger changes by CD tools (even though using a numeric is allowed).


mkuratczyk:
Thanks. This sounds like a bug in Flux to be honest, but 🤷

PR #1690, opened 2024-08-03
Support override default topologySpreadContraints

### LINKED ISSUE #1687: Support override default topologySpreadContraints

If Kubernetes nodes do not have the topology.kubernetes.io/zone label, pods are scheduled incorrectly (all can be put in the same node). Antiaffinity can be used as a workaround, but it will be preferable to allow overwriting the topologySpreadConstraints variable.

Setting spec.override.statefulSet.topologySpreadConstraints allows adding but not removing topology constraints.

If I put:

``` yaml
kind: RabbitmqCluster
metadata:
  name: production-ready
spec:
  override:
    statefulSet:
      spec:
        template:
          spec:
            containers: []
            topologySpreadConstraints: []
```

I got in the Statefulset topology:

``` yaml
            topologySpreadConstraints:
            - maxSkew: 1
              topologyKey: "topology.kubernetes.io/zone"
              labelSelector:
                matchLabels:
                  app.kubernetes.io/name: production-ready
```

I will gladly contribute a PR if you agree with this solution or tell me an alternative one.

I would prefer that topologySpreadConstraints values in the RabbitmqCluster do not merge adding but overriding.

If I put:

``` yaml
kind: RabbitmqCluster
metadata:
  name: production-ready
spec:
  override:
    statefulSet:
      spec:
        template:
          spec:
            containers: []
            topologySpreadConstraints: []
```

The Statefulset will not have any topologySpreadConstraints.

DanielDorado:
Allow a user to override the default TopologySpreadConstraints. Appending to the default constraint is not supported, so the user should put it explicitly in the override. In a rare case, when appending the default is needed, putting it explicitly in the override is not a big deal.

This closes #1687 

## Local Testing

```
```

```
$ kubectx dev-bunny
$ make destroy deploy-dev
$ make system-tests
```

### DIFF
--- a/internal/resource/statefulset.go
+++ b/internal/resource/statefulset.go
@@ -298,6 +298,14 @@ func patchPodSpec(podSpec, podSpecOverride *corev1.PodSpec) (corev1.PodSpec, err
 		}
 	}
 
+	// A user can may wish to override the default TopologySpreadConstraints. To remove the default constraints without adding new ones,
+	// set this to an empty array. To append to the default constraint is not supported, the user should put it explicitly in the override.
+	// As the merge key for the TopologySpreadConstraints is the TopologyKey there is no way to remove it with the `strategicpatch.StrategicMergePatch`.
+	// When a user wants to overwrite these topology constraints it can be assumed that the default will not be used, besides, in a rare case
+	// when appending the default is needed, then, to put it explicitly in the override is not a big deal.
+	if podSpecOverride.TopologySpreadConstraints != nil {
+		patchedPodSpec.TopologySpreadConstraints = podSpecOverride.TopologySpreadConstraints
+	}
 	return patchedPodSpec, nil
 }
 
--- a/internal/resource/statefulset_test.go
+++ b/internal/resource/statefulset_test.go
@@ -2010,16 +2010,6 @@ default_pass = {{ .Data.data.password }}
 								},
 							},
 						},
-						corev1.TopologySpreadConstraint{
-							MaxSkew:           1,
-							TopologyKey:       "topology.kubernetes.io/zone",
-							WhenUnsatisfiable: corev1.ScheduleAnyway,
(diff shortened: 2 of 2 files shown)


Zerpet:
Our CI is having a rework at the moment. I will verify that tests pass locally before approving. The code changes look correct to me.

mkuratczyk:
It doesn't seem to address the main issue. Am I missing something?

```
> cat /tmp/a.yml
kind: RabbitmqCluster
metadata:
  name: a
spec:
  override:
    statefulSet:
      spec:
        template:
          spec:
            containers: []
            topologySpreadConstraints: []

> kubectl apply -f /tmp/a.yml
rabbitmqcluster.rabbitmq.com/a created

> kubectl get -o json sts a-server | jq .spec.template.spec.topologySpreadConstraints
[
  {
    "labelSelector": {
      "matchLabels": {
        "app.kubernetes.io/name": "a"
      }
    },
    "maxSkew": 1,
    "topologyKey": "topology.kubernetes.io/zone",
    "whenUnsatisfiable": "ScheduleAnyway"
  }
]```

Zerpet:
I think the API server may not differentiate between empty array and nil, once encoded by the server: https://github.com/kubernetes/kubernetes/issues/70281

In which case, this PR won't address the issue 😢

DanielDorado:
Yes, both of you are right.

It does not allow putting an empty topologySpreadConstraints, but it does allow us to overwrite it with a topology spread constraint referencing a node label that exists in our Kubernetes nodes.

As this behavior is a (little) better than not overriding at all, I put the PR anyway.

Our problem is solved, and anybody who needs to overwrite the topologySpreadConstraints because of lack of the "topology.inditex.io/zone label or because they do not want to use it will be satisfied.

But, anyway, if you do not want to accept this solution, I will do a PR with a better solution.

DanielDorado:
Closed on behalf of #1694

PR #1705, opened 2024-08-19
Disable ANONYMOUS login

ansd:
See https://github.com/rabbitmq/rabbitmq-server/pull/11999 for full context.

Starting with RabbitMQ 4.0, anonymous login should be disabled in production.

`anonymous_login_user` and `anonymous_login_pass` both default to `guest` in RabbitMQ 4.0. The rabbitmq/cluster-operator complies already with best practices by provisioning a new `default_user` and `default_pass` instead of using RabbitMQ's default `guest` user.
Instead of having RabbitMQ advertise the ANONYMOUS mechanism, this commit disables anonymous logins.

Because `anonymous_login_user` is a new RabbitMQ 4.0 `rabbitmq.conf` setting and the cluster-operator doesn't know what RabbitMQ version it deploys and setting `rabbitmq.conf` key `anonymous_login_user` in RabbitMQ 3.13 would make booting RabbitMQ fail, this commit modifies the `auth_mechanisms.*` settings in `rabbitmq.conf`:
If the user provided a conscious choice on what `auth_mechanisms` RabbitMQ should advertise, this configuration will be respected. If the user did not configure `auth_mechanisms`, the cluster-operator will disable ANONYMOUS logins by setting only:
```
auth_mechanisms.1 = PLAIN
auth_mechanisms.2 = AMQPLAIN
```

### DIFF
--- a/internal/resource/configmap_test.go
+++ b/internal/resource/configmap_test.go
@@ -36,7 +36,10 @@ cluster_formation.peer_discovery_backend   = rabbit_peer_discovery_k8s
 cluster_formation.k8s.host                 = kubernetes.default
 cluster_formation.k8s.address_type         = hostname
 cluster_formation.target_cluster_size_hint = 1
-cluster_name                               = ` + instanceName)
+cluster_name                               = ` + instanceName + `
+auth_mechanisms.1                          = PLAIN
+auth_mechanisms.2                          = AMQPLAIN
+`)
 }
 
 var _ = Describe("GenerateServerConfigMap", func() {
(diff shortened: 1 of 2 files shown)


ansd:
Note to reviewers: You could deploy different configurations in `spec.rabbitmq.additionalConfig` or `spec.rabbitmq.advancedConfig` and check the effective configuration via `kubectl exec myrabbit-server-0 -c rabbitmq -- rabbitmqctl eval 'application:get_env(rabbit, auth_mechanisms).'`

internal/resource/configmap_test.go
```diff
@@ -168,6 +171,22 @@ var _ = Describe("GenerateServerConfigMap", func() {
...
+					instance.Spec.Rabbitmq.AdditionalConfig = userDefinedConfiguration
+					expectedConfiguration := iniString(userDefinedConfiguration)
+
+					Expect(configMapBuilder.Update(configMap)).To(Succeed())
+					Expect(configMap.Data).To(HaveKeyWithValue("userDefinedConfiguration.conf", expectedConfiguration))
+					operatorDefaults, exists := configMap.Data["operatorDefaults.conf"]
+					Expect(exists).To(BeTrue())
+					Expect(operatorDefaults).NotTo(ContainSubstring("auth_mechanisms"))
+					Expect(operatorDefaults).NotTo(ContainSubstring("PLAIN"))
+					Expect(operatorDefaults).NotTo(ContainSubstring("ANONYMOUS"))
```
Zerpet, line 187: This is fine, it works as expected. However, I would recommend the matcher [HaveKey()](https://onsi.github.io/gomega/#havekeykey-interface) instead of "expect bool to be true", because a failed assertion would give as little as "expected false to be true", whilst the other matcher would give a more helpful error.

Zerpet (approved):

PR #1720, opened 2024-09-02
Updates to workflows

Zerpet:
## Summary Of Changes

- Use supported method to generate step output
- Add missing permission for Upload GCS job
- Use go.mod to determine Go version to install
- Use floating major tag
- Switch to GChat notification

## Additional Context

Deprecation announcement for set-output command:
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


.github/workflows/prometheus-rules.yml
```diff
@@ -6,18 +6,18 @@ on:
...
     steps:
+      - uses: actions/checkout@v4
+        name: Checkout code
+
       - name: Install Go
         uses: actions/setup-go@v5
         with:
-          go-version: ${{ env.GO_VERSION }}
-      - uses: actions/checkout@v4
+          go-version-file: "go.mod"
```
MirahImage, line 19: Nice, I didn't know this was an option, definitely going to grab this for other places

MirahImage (approved):

PR #1740, opened 2024-10-03
Ignore mirroring related shutdown errors

### LINKED ISSUE #1734: Remove deprecated mirrored queue majority check
https://github.com/rabbitmq/cluster-operator/blob/82415be7ae2e972d8badbf099679b4764a34762c/internal/resource/statefulset.go#L616-L618

The command `rabbitmq-upgrade await_online_synchronized_mirror` is removed in 4.0. In order to support 4.0, we need to remove this check from the pre-stop hook, otherwise, the hook will always fail, and Kubernetes will ungracefully kill the Pod.

From: https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#hook-handler-execution

> If either a PostStart or PreStop hook fails, it kills the Container.

mkuratczyk:
RabbitMQ 4.0 doesn't include the mirroring-specific command `rabbitmq-upgrade await_online_synchronized_mirror`. Therefore, the preStop hook would fail on 4.0, leading to the maintenance mode not getting enabled at all (the moment a command fails in preStop, the shutdown process continues as if preStop completed).

This approach should make the Operator compatible with 3.x and 4.x.

This closes #1734

### DIFF
--- a/internal/resource/statefulset.go
+++ b/internal/resource/statefulset.go
@@ -614,7 +614,7 @@ func (builder *StatefulSetBuilder) podTemplateSpec(previousPodAnnotations map[st
 								Command: []string{"/bin/bash", "-c",
 									fmt.Sprintf("if [ ! -z \"$(cat /etc/pod-info/%s)\" ]; then exit 0; fi;", DeletionMarker) +
 										fmt.Sprintf(" rabbitmq-upgrade await_online_quorum_plus_one -t %d &&"+
-											" rabbitmq-upgrade await_online_synchronized_mirror -t %d &&"+
+											" rabbitmq-upgrade await_online_synchronized_mirror -t %d || true &&"+
 											" rabbitmq-upgrade drain -t %d",
 											*builder.Instance.Spec.TerminationGracePeriodSeconds,
 											*builder.Instance.Spec.TerminationGracePeriodSeconds,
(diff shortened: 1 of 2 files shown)


internal/resource/statefulset.go
```diff
@@ -614,7 +614,7 @@ func (builder *StatefulSetBuilder) podTemplateSpec(previousPodAnnotations map[st
 								Command: []string{"/bin/bash", "-c",
 									fmt.Sprintf("if [ ! -z \"$(cat /etc/pod-info/%s)\" ]; then exit 0; fi;", DeletionMarker) +
 										fmt.Sprintf(" rabbitmq-upgrade await_online_quorum_plus_one -t %d &&"+
-											" rabbitmq-upgrade await_online_synchronized_mirror -t %d &&"+
+											" rabbitmq-upgrade await_online_synchronized_mirror -t %d || true &&"+
```
MirahImage, line 617: Does this actually do what we expect? There's an `&&` following the `|| true`. I think we need to somehow group `rabbitmq-upgrade await_online_synchronized_mirror || true` into one command, so that the logic works out right.

Zerpet: Apparently, not in Bash or other shells. TIL
https://unix.stackexchange.com/questions/88850/precedence-of-the-shell-logical-operators#:~:text=In%20many%20computer%20languages%2C%20operators,%7C%7C%20have%20the%20same%20precedence.

mkuratczyk: Yeah, I kind of expected this not to work but I did try. Information about the node going into maintenance mode is back. Also, a manual tests like this:
```
bash -c 'echo foo && ehco bar || echo meh && echo baz'
foo
bash: line 1: ehco: command not found
meh
baz
```

PR #1761, opened 2024-10-31
Remove single-arch builds

Zerpet:
## Summary Of Changes

- Updates to Actions: remove single arch builds
- Update minimum version of rabbit and k8s in CI


Zerpet:
Vault TLS example is failing without apparent reason. Running the same test locally works well 🤷‍♂️

MirahImage (approved):

PR #1859, opened 2025-04-22
Add support for custom Service labels via spec.service.labels

### LINKED ISSUE #1821: Adding labels to spec.service
I need to attach labels to my service for cilium to do it's job. Because there is only a `spec.service.annotations` i have to set this via `spec.override.service.metadata.labels`

I'd like to be able to set labels the same way we do it for annotations via `spec.service.labels`

There's the workaround with the override function, but this isn't the most beautiful solution, if annotations and labels cannot be defined in the same place

**Additional context**
I know that there was an issue with pods and labels: https://github.com/rabbitmq/cluster-operator/issues/442 
Please correct me, but this shouldn't be an issue here, right?

EladAviczer:
This closes #1821 

## Summary Of Changes
This PR introduces support for applying custom labels to the Service created by the RabbitmqCluster resource.
in the practicesuggested by the issue opener spec.service.labels

## Local Testing
unit and integration tests succeeded.
Deployed to local cluster using the makefile.

##  Sample usage

```
kind: RabbitmqCluster
metadata:
  name: example
spec:
  replicas: 1
  service:
    labels:
      foo: "bar"

```


EladAviczer:
For the reviewer: please consider moving the `if` clause further down in the function for aesthetic reasons

mkuratczyk:
Thanks for looking into this. It does work, but a few things are missing:
1. There are [internal labels](https://www.rabbitmq.com/kubernetes/operator/using-operator#internal-labels) that currently can be overwritten by `.spec.service.labels`. I don't think they should be - we use these labels to find child objects in some cases, so I think we should guarantee they are set as we expect and not overwritten by the user
2. There's no test validating this behaviour. We should add one similar to https://github.com/rabbitmq/cluster-operator/blob/main/internal/resource/service_test.go#L875-L904

EladAviczer:
@mkuratczyk 
took some time to understand your test file but i think it is good now
please feel free to recheck now : )

mkuratczyk:
Thank you!

Issue #960, opened 2022-02-15
`initContainerMemory` and cpu should can be customize


jinleileiking:
In the code:

```
initContainerMemory string = "500Mi"
```

This should can be modified in rabbitmq.yaml for using in test env.

mkuratczyk:
How much lower do you expect to set it? While we could probably lower this value or make it configurable, I'm honestly not sure we'd like RabbitMQ deployed in environments with less than 0.5GB available. It's a recipe for very poor performance and overall experience of using RabbitMQ. Can you elaborate on your use case?

jinleileiking:
In a test env，I only test for the function. In aws k8s 500M is a lot of money, so, I think configure this size is useful.

BTW, I can set the memory in rabbit container, why I can not set in rabbit-setup container.

Moreover， the init container will exit when it fininshed.

mkuratczyk:
As you mentioned, this container exists for a very short time, so I really doubt you can save any significant amount of money by assigning less memory for it. If you really think it's valuable, you can actually customize it right now:
```
apiVersion: rabbitmq.com/v1beta1
kind: RabbitmqCluster
metadata:
  name: my-rabbitmqcluster
spec:
  override:
    statefulSet:
      spec:
        template:
          spec:
            containers: []
            initContainers:
            - name: setup-container
              resources:
                limits:
                  memory: 250Mi
                requests:
                  memory: 250Mi
```
Just keep in mind that this container could need more memory in the future (I hope not, but we can't guarantee that).

If you know how much you can save by lowering this value, please let us know.

jinleileiking:
It worked , Thanks!

jinleileiking:
save a 4cpu 8G machine :)

mkuratczyk:
Can you explain how this change allows you to save a 4CPU/8GB machine? I don't get it. Thanks,

zvlb:
Hi @jinleileiking @mkuratczyk 

I’d like to reopen this issue and clarify why configurability of the initContainer resources (CPU/memory) is critical for users like me.

**Scalability Across Many Clusters**:
We manage hundreds of RabbitMQ clusters via the operator. Requiring per-cluster modifications to override initContainer resources is impractical. A global configuration (e.g., via Operator settings) would avoid repetitive overrides in every RabbitmqCluster custom resource. While mutation webhooks are a workaround, they add unnecessary complexity for such a basic need.

**Excessive Defaults for Simple Tasks**:
[...]

PR #1863, opened 2025-04-29
PROPOSAL: Reduce init container memory from 500Mi to 64Mi

mkuratczyk:
It'd be great to get some feedback from users who changed the default init container resources - what did you change them to? I've tested successfully deployed RabbitMQ with a `50Mi` init container using:
* `rabbitmq:4.1-management`
* `bitnami/rabbitmq:4.1`
* `rabbitmq.packages.broadcom.com/vmware-tanzu-rabbitmq:4.1.0`

Given that an init container failure would be a terrible user experience, I'd rather have some headroom, hence `64Mi`.

When originally developed, it wasn't clear what would need to be done in the init container. With hindsight, there's very little to do and the init phase hasn't changed in a long time.

We also assumed that the resources requested for the init container would effectively be reused by the actual `rabbitmq` container later. Turns out that's not the case: https://github.com/kubernetes/kubernetes/issues/124282 init container resources are considered even after the init container completes its task.

This closes https://github.com/rabbitmq/cluster-operator/issues/960

### DIFF
--- a/internal/resource/statefulset.go
+++ b/internal/resource/statefulset.go
@@ -35,7 +35,7 @@ import (
 
 const (
 	stsSuffix           string = "server"
-	initContainerCPU    string = "100m"
+	initContainerCPU    string = "20m"
 	initContainerMemory string = "64Mi"
 	defaultPVCName      string = "persistence"
 	DeletionMarker      string = "skipPreStopChecks"
--- a/internal/resource/statefulset.go
+++ b/internal/resource/statefulset.go
@@ -36,7 +36,7 @@ import (
 const (
 	stsSuffix           string = "server"
 	initContainerCPU    string = "100m"
-	initContainerMemory string = "500Mi"
+	initContainerMemory string = "64Mi"
 	defaultPVCName      string = "persistence"
 	DeletionMarker      string = "skipPreStopChecks"
 )
--- a/internal/resource/statefulset_test.go
+++ b/internal/resource/statefulset_test.go
@@ -783,9 +783,9 @@ var _ = Describe("StatefulSet", func() {
 
 			resources := statefulSet.Spec.Template.Spec.InitContainers[0].Resources
 			Expect(resources.Requests["cpu"]).To(Equal(k8sresource.MustParse("100m")))
-			Expect(resources.Requests["memory"]).To(Equal(k8sresource.MustParse("500Mi")))
+			Expect(resources.Requests["memory"]).To(Equal(k8sresource.MustParse("64Mi")))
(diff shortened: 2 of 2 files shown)


zvlb:
We tested with this resources:
```
resources:
  limits:
    cpu: 10m
    memory: 50Mi
  requests:
    cpu: 10m
    memory: 50Mi
```

All working good.

I think we can install:
```
initContainerCPU    string = "20m"
initContainerMemory string = "64Mi"
```
without any potential issues for users

Zerpet:
I like this proposal 👍 The init container doesn't do anything special, and it uses very basic shell commands like `cp` and `mv`.

mkuratczyk:
@Zerpet I've also pushed a change in CPU from 100m to 20m. Still ok with this?

Zerpet:
`20m` is 0.02 CPU 🤔 I understand the intention of lowering this value so drastically, after reading again https://github.com/kubernetes/kubernetes/issues/124282 and revisiting [Pod QoS documentation](https://kubernetes.io/docs/concepts/workloads/pods/pod-qos/). I'm not opposed to this change, since there are claims that `20m` CPU JustWorks™

mkuratczyk:
Yeah, I even tried what happens if I try to use `20m` to start RabbitMQ itself. I didn't wait long enough to see it running but nothing was crashing, it was just starting very very slowly... :)
Should not be a problem at all to just move some files around.

zvlb:
@mkuratczyk @Zerpet Thank you!

Сan I know when to expect a release with these changes?
The last release was on January 24th and I'm a little scared to guess how long to wait for the next one. Maybe there is some kind of release plan?

mkuratczyk:
There's just not that much to release usually. The Operator as it is, seems to be working well for a lot of people.
We can discuss within the team tomorrow and let you know.

One thing you I'd love to hear your input for is whether it's a problem that this change, as currently implemented, would restart all those clusters you have, since changes to the StatefulSet definition cause a restart. We could make some tweaks to prevent this I think

zvlb:
I understand that restarting all RabbitMQ clusters is an unpleasant situation and could indeed cause dissatisfaction among users. However, in my case, I don’t consider this a problem. If someone deployed a RabbitMQ cluster with a single instance and expects 100% SLA—that’s the problem of whoever set it up that way. Besides the reboot, such RabbitMQ installations cannot be considered stable, as a pod can be terminated for a multitude of reasons, and changes to the StatefulSet are just one of them!

If the RabbitMQ cluster is deployed with more than one node, updating the StatefulSet should proceed without performance issues.

PR #1864, opened 2025-05-05
Improve cluster deletion

mkuratczyk:
There were 3 issues:
* the STS could get deleted before deletion label propagated, leading to some pods terminating for a long time, because they didn't skip quorum checks
* there was an unnecessary explicit STS deletion
* addRabbitmqDeletionLabel was not filtering by namespace, potentially performing operations on wrong/too many pods (if there were multiple clusters with the same name in different namespaces)

### DIFF
--- a/controllers/rabbitmqcluster_controller_test.go
+++ b/controllers/rabbitmqcluster_controller_test.go
@@ -43,6 +43,7 @@ var _ = Describe("RabbitmqClusterController", func() {
 
 	var (
 		cluster          *rabbitmqv1beta1.RabbitmqCluster
+		suffix           string
 		defaultNamespace = "default"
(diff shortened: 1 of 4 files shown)


Zerpet (approved):
Nothing to object, the solution looks good to me 👍 Glad you figured out the problem!

Out of curiosity, did you consider using [CreateOrUpdate()](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#CreateOrUpdate) function in `removeFinalizer()` ?
`CreateOrUpdate` does essentially what you are doing of `Get` prior to `Update()`

Zerpet:
Ship it! :shipit:

PR #1873, opened 2025-05-14
Track tools using `tool` directive

Zerpet:
## Summary Of Changes

- **Track internal tooling using go.mod directive**
- **Add promtool as tool dependency in root go.mod**

## Additional Context

New in Go 1.24: https://tip.golang.org/doc/go1.24#tools
https://tip.golang.org/doc/modules/managing-dependencies#tools

**Important context:** I decided to keep the internal `tools` module because `go get -tool` adds all the tool's dependencies into your project. That means, we will end up with a much larger `go.mod` with many `require` directives for the tool's dependencies. Whilst this in itself is not a problem (that's what `go.mod` is for), it has the potential to trigger false positives on security scanning tools.

For example, `promtool` pulls a version of `github.com/golang-jwt/jwt` that has a known vulnerability. In this case, I realised it was not a vulnerability on our project, but on the tool dependency. For this reason, I decided to keep the internal tools pattern: to minimise false positives from security scanners

There's still some merit to having the `tool` directive in `go.mod`, because it pulls more dependencies, which means is more accurately tracking what our project depends on (even if we don't link to it).

## Local Testing

Run `go install tool` and `cd internal/tools; go install tool`. All tools build successfully and are installed to GOBIN.


MirahImage (approved):
Excited for this change

Zerpet:
@MirahImage I added extra context to the OP. I have some mixed feelings with the result, because I encountered some dependency conflicts between our tools. As a result, one of the tools moved to the "main" `go.mod`, resulting in a bit of bloat. This is not necessarily a bad thing, unless/until we start receiving false positives from security scanners 🙃

PR #1883, opened 2025-05-28
update prometheus rules for rabbitmq v4.1

### LINKED ISSUE #1871: Some PrometheusRules stopped working with rabbitmq 4 update
## Describe the bug

We used the rules from `observability/prometheus/rules/rabbitmq` in Grafana and since the update to rabbitmq 4, some of those rules stopped working and produce errors.

For example `UnroutableMessage`: 

```
found duplicate series for the match group {instance="..."} on the right hand-side of the operation: 
  [{__name__="rabbitmq_identity_info", cluster="prod2-user-workload", container="rabbitmq", endpoint="prometheus", instance="...", job="my-rabbitmq", namespace="my-namespace", pod="my-rabbitmq-server-0", rabbitmq_cluster="my-rabbitmq", rabbitmq_cluster_permanent_id="rabbitmq-cluster-id-SxlB3ZdnzxD6pkY2U8jwLg", rabbitmq_endpoint="detailed", rabbitmq_node="rabbit@my-rabbitmq-server-0.my-rabbitmq-nodes.my-namespace", service="my-rabbitmq"}, 
   {__name__="rabbitmq_identity_info", cluster="prod2-user-workload", container="rabbitmq", endpoint="prometheus", instance="...", job="my-rabbitmq", namespace="my-namespace", pod="my-rabbitmq-server-0", rabbitmq_cluster="my-rabbitmq", rabbitmq_cluster_permanent_id="rabbitmq-cluster-id-SxlB3ZdnzxD6pkY2U8jwLg", rabbitmq_endpoint="aggregated", rabbitmq_node="rabbit@my-rabbitmq-server-0.my-rabbitmq-nodes.my-namespace", service="my-rabbitmq"}];many-to-many matching not allowed: matching labels must be unique on one side
```

This is caused by the duplicate `rabbitmq_build_info` for the different endpoints.

Related but slightly different: all rules in `cluster-alarms.yml` don't handle rabbitmq clusters with multiple replicas.

awoimbee:
This closes #1871

## Summary Of Changes

Update alermanager rules to work with the latest rabbitmq version (v4.1).

* Remove rule `rabbitmq-tcp-sockets-near-limit` that depends on removed metrics:
  *  `rabbitmq_process_open_tcp_sockets`
  * `rabbitmq_process_max_tcp_sockets`
* Fix joins with `rabbitmq_identity_info` due to the `rabbitmq_endpoint` label being added
* Since I had to change a bit of everything I standardized some patterns (eg joining solely on `instance` everywhere)

## Local Testing

This fixes all rule evaluation errors in my setup:
image

I did not run the tests since I'm missing `setup-envtest` but I guess is doesn't matter ?


mkuratczyk:
@awoimbee are you sure about using `instnace` as the only key for joins? If I'm not mistaken, that will break if someone scapes the same node with multiple jobs (eg. `/metrics` every 30s + `/metrics/per-object` every 2 minutes or something like that). In such cases the `instance` will be the same

Issue #1240, opened 2023-01-25
Enable all feature flags on upgrade


MirahImage:
The default behavior of the cluster operator should be to enable all feature flags after an upgrade as an additional PostDeploy step.

Currently, all feature flags are enabled when a cluster is created, but they are never again enabled. Should a new feature flag be added during an upgrade, then that feature flag will not currently automatically be enabled. This could cause future upgrades to fail without manual intervention to enable the feature flags.

This behavior could be disabled by disabling the PostDeploy steps, much like the queue rebalance.

ansd:
I think it's safer for a human operator to decide when to enable what kind of feature flag.
Enabling a feature flag could - depending on the migration function of the feature flag - pause certain operations in RabbitMQ:
> As an operator, the most important part of this procedure to remember is that if the migration takes time, some components and thus some operations in RabbitMQ might be blocked during the migration.

However, having on opt-in (or opt-out) option sounds reasonable.

h0jeZvgoxFepBQ2C:
Hm, how do you enable feature flags afterwards? I just tried to enable oauth2 plugin, but I get following error after I connect to one of my nodes and submit following command:

```
rabbitmq-cluster-server-0:/$ rabbitmq-plugins enable rabbitmq_auth_backend_oauth2
Enabling plugins on node rabbit@rabbitmq-cluster-server-0.rabbitmq-cluster-nodes.rabbitmq-cluster:
rabbitmq_auth_backend_oauth2
Error:
{:plugins_not_found, [:rabbit_auth_backend_oauth2]}
```

JSurf:
We just ran into that issue while trying to upgrade from 3.11.17 to 3.12.x

From the documentation (https://www.rabbitmq.com/feature-flags.html)

> When a node starts for the first time, all supported feature flags are enabled by default. When a node is upgraded to a newer version of RabbitMQ, new feature flags are enabled by default if it is a single isolated node, or remain disabled by default if it belongs to a cluster.

We have a single node rabbitmq-server which was previously upgraded from an earlier version to 3.11. While upgrading from 3.11 to 3.12 we experienced that not all feature flags were enabled on 3.11, failing the upgrade. 

So 3.11 did not enable all feature flags on upgrade automatically as I would have expected from the docs

ftdcn:
Sometimes it's not as easy as "I think it's safer for a human operator to decide when to enable what kind of feature flag.".

[...]

PR #1892, opened 2025-06-17
Optionally enable all feature flags automatically

mkuratczyk:
If the cluster definition contains:
```
spec:
  autoEnableAllFeatureFlags: true
```

then after each cluster restart, the Operator will call `rabbitmqctl enable_feature_flag all`, which enables all stable feature flags. The property is `false` by default, so this feature is opt-in.

Resolves https://github.com/rabbitmq/cluster-operator/issues/1240

### DIFF
--- a/controllers/reconcile_cli_test.go
+++ b/controllers/reconcile_cli_test.go
@@ -179,7 +179,7 @@ var _ = Describe("Reconcile CLI", func() {
 							cluster.Spec.AutoEnableAllFeatureFlags = false
 						}
 					})).Should(Succeed())
-					Eventually(fakeExecutor.ExecutedCommands).Within(time.Second * 5).WithPolling(time.Second).
+					Consistently(fakeExecutor.ExecutedCommands).Within(time.Second * 5).WithPolling(time.Second).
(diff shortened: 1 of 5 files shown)


MirahImage (approved):

controllers/reconcile_cli_test.go
```diff
@@ -94,46 +91,114 @@ var _ = Describe("Reconcile CLI", func() {
...
+					// all replicas ready to execute commands. We have to reset the "registry"
+					// of commands.
+					fakeExecutor.ResetExecutedCommands()
+
+					Eventually(k.Update(cluster, func() {
+						if cluster != nil {
+							cluster.Spec.AutoEnableAllFeatureFlags = false
+						}
+					})).Should(Succeed())
+					Eventually(fakeExecutor.ExecutedCommands).Within(time.Second * 5).WithPolling(time.Second).
```
Zerpet, line 182: I think we should assert with `Consistently` instead of `Eventually`, because you may assert too soon, and eventually will pass, even if the reconcile loop in the background runs the command later.

If I understand the intention of this test, correctly, we want to ensure that the command is not run; you can't guarantee that without `Consistently`.

PR #1938, opened 2025-08-28
Bump tools and deps

Zerpet:
For some reason, dependabot has not created PRs for this yet


Zerpet:
Do not merge: there is a legitimate failure in kubectl tests

Zerpet:
kubectl tests are now passing 🤷‍♂️ There was a legitimate failure in system tests because one test relied on an "undocumented feature" related to https://github.com/rabbitmq/rabbitmq-server/pull/14359

Our test was setting the rabbitmq authentication backend to LDAP, but it did not enable the LDAP plugin. Prior to 4.2.0, this was valid configuration. After the mentioned PR, rabbit fails at boot. Our test now simply uses a different configuration in advanced config.

PR #1939, opened 2025-08-28
Use github packages for our container image

Zerpet:
## Summary Of Changes

- Push our container image to GitHub packages
- Generate additional release manifests for ghcr

## Additional Context

N/A

## Local Testing

You can verify the new make target with:

```
make generate-installation-manifest
```

And observe a manifest file specific for ghcr, and with an image reference to `ghcr.io`

Job showing a successful push to ghcr: https://github.com/rabbitmq/cluster-operator/actions/runs/17293728273/job/49087638691


.github/workflows/build-test-publish.yml
```diff
@@ -13,7 +13,7 @@ on:
     branches: [ "main" ]
 
 env:
-  GO_VERSION: ~1.24
+  GO_VERSION: 'stable'
```
MirahImage, line 16: We should presumably include some sort of equivalent functionality in the Dockerfile or we'll run into problems in the build stage

Zerpet:
Test should pass after merging #1938

Issue #1960, opened 2025-09-30
set cluster-operator securitycontext


3deep5me:
**Is your feature request related to a problem? Please describe.**
The current deployment of the operator is not Pod Security Standards:restricted compatible. 
Which is why i get this error
```
  Warning  FailedCreate  5m1s (x8 over 10m)  replicaset-controller  (combined from similar events): Error creating: pods "rabbitmq-cluster-operator-6b5b87cdc6-vr96l" is forbidden: violates PodSecurity "restricted:latest": allowPrivilegeEscalation != false (container "operator" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (container "operator" must set securityContext.capabilities.drop=["ALL"]), runAsNonRoot != true (pod or container "operator" must set securityContext.runAsNonRoot=true), seccompProfile (pod or container "operator" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
```
**Describe the solution you'd like**
Setting the security context to this:
```
    securityContext:
      allowPrivilegeEscalation: false
      capabilities:
        drop:
        - ALL
      runAsNonRoot: true
      runAsUser: 1001
      seccompProfile:
        type: RuntimeDefault
```
**Describe alternatives you've considered**
I patched it manually - but i think it would be also nice to have it in the upstream.

3deep5me:
Thank you!!!

PR #1961, opened 2025-10-01
Add security context

### LINKED ISSUE #1910: improve RabbitMqCluster default securitycontext
By default, the rabbitmq cluster `StatefulSet` gets created with this pod security context:

```
securityContext:
    runAsUser: 999
    fsGroup: 0
```

I tested this configuration today:

```
kind: RabbitmqCluster
metadata:
  name: rabbitmq
  namespace: rabbitmq
spec:
  override:
    statefulSet:
      spec:
        template:
          spec:
            securityContext:
              runAsNonRoot: true
              runAsUser: 999
              runAsGroup: 999
              fsGroup: 999
              seccompProfile:
                type: RuntimeDefault
            initContainers:
            - name: setup-container
              securityContext:
                allowPrivilegeEscalation: false
                capabilities:
                  drop:
                    - ALL
                privileged: false
                readOnlyRootFilesystem: true
            containers:
            - name: rabbitmq
              securityContext:
                allowPrivilegeEscalation: false
                capabilities:
                  drop:
                    - ALL
                privileged: false
                readOnlyRootFilesystem: true
```

I think a configuration like this can be included in the operator. I understand if not all of these can be included for all types of kubernetes clusters by default, but there is room for improvement.

MirahImage:
This closes #1910 and #1960 

Adds a securityContext to the cluster operator deployment context. 
Adds securityContext to RabbitMQ Pods, containers, and init containers.

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -51,4 +51,12 @@ spec:
         - containerPort: 9782
           name: metrics
           protocol: TCP
+        securityContext:
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop:
+            - All
+          runAsNonRoot: true
+          seccompProfile:
+            type: RuntimeDefault
       terminationGracePeriodSeconds: 10
--- a/internal/resource/statefulset.go
+++ b/internal/resource/statefulset.go
@@ -570,8 +570,12 @@ func (builder *StatefulSetBuilder) podTemplateSpec(previousPodAnnotations map[st
 		Spec: corev1.PodSpec{
 			TopologySpreadConstraints: builder.defaultTopologySpreadConstraints(),
 			SecurityContext: &corev1.PodSecurityContext{
-				FSGroup:   ptr.To(int64(0)),
-				RunAsUser: &rabbitmqUID,
+				FSGroup:      ptr.To(int64(0)),
+				RunAsUser:    &rabbitmqUID,
+				RunAsNonRoot: ptr.To(bool(true)),
+				SeccompProfile: &corev1.SeccompProfile{
+					Type: corev1.SeccompProfileTypeRuntimeDefault,
+				},
 			},
 			ImagePullSecrets:              builder.Instance.Spec.ImagePullSecrets,
 			TerminationGracePeriodSeconds: builder.Instance.Spec.TerminationGracePeriodSeconds,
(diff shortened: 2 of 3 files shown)


MirahImage:
Given the security context appears in 4 different places (once in the operator deployment itself, then in three places in the cluster statefulset), I suspect we may need to improve our documentation about overrides for OpenShift.

Zerpet:
Docs regarding openshift will need some updating. I'm going to run this locally in CRC, but we likely need to override the `RunAsUser` bit in favour of the FSGroup

Zerpet:
I had to switch to something else on Friday. I'm looking into it now.

Zerpet:
Indeed it doesn't like the `RunAsUser: 999` nor the `FSGroup: 0`

<details><summary>STS condition</summary>
<p>

```
Warning  FailedCreate      8m27s (x16 over 11m)  statefulset-controller  create Pod rabbitmq-server-0 in StatefulSet rabbitmq-server failed error: pods "rabbitmq-server-0" is forbidden: unable to validate against any security context constraint: [provider "anyuid": Forbidden: not usable by user or servieaccount, provider restricted-v2: .spec.securityContext.fsGroup: Invalid value: [[]int64{0}: 0 is not an allowed group, provider restricted-v2: .initContainers[0].runAsUser: Invalid value: 999: must be in the ranges: [1000660000, 1000669999], provider restricted-v2: .containers[0].runAsUser: Invalid value: 999: must be in the ranges: [1000660000, 1000669999], provider "restricted": Forbidden: not usable by user or serviceaccount, provider "nonroot-v2": Forbidden: not usable by user or serviceaccount, provider "nonroot": Forbidden: not usable by user or serviceaccount, provider "hostmount-anyuid": Forbidden: not usable by user or serviceaccount, provider "hostmount-anyuid-v2": Forbidden: not usable by user or serviceaccount, provider "machine-api-termination-handler": Forbidden: not usable by user or serviceaccount, provider "hostnetwork-v2": Forbidden: not usable by user or serviceaccount, provider "hostnetwork": Forbidden: not usable by user or serviceaccount, provider "hostaccess": Forbidden: not usable by user or serviceaccount, provider "hostpath-provisioner": Forbidden: not usable by user or serviceaccount, provider "privileged": Forbidden: not usable by user or serviceaccount]
```

</p>
</details> 

The error describes the problem:

```
provider restricted-v2: 
  .spec.securityContext.fsGroup: Invalid value: [[]int64{0}: 0 is not an allowed group, 
provider restricted-v2: 
  .initContainers[0].runAsUser: Invalid value: 999: must be in the ranges: [1000660000, 1000669999], 
provider restricted-v2: 
  .containers[0].runAsUser: Invalid value: 999: must be in the ranges: [1000660000, 1000669999]
```

This can be workaround with the suggestion from [our docs](https://www.rabbitmq.com/kubernetes/operator/using-on-openshift#arbitrary-user-ids) that basically removes the Pod security context. With the security contexts in the containers, it's enough to pass the "restricted" validation.

A potential improvement would be to set to `nil`, or not set at all, the problematic fields. It left unset, Openshift does the right thing and assigns a valid value. We could customise the behaviour for Openshift if an annotation e.g. `rabbitmq.com/platform-is-openshift: some-value-doesnt-matter` is present in `RabbitmqCluster` objtect.

bo0ts:
Quick feedback on this change: this broke all our RabbitMQ deployments in OpenShift and OKD.

We deploy rabbitmq instances by adding the anyuid permission to the rabbitmq serviceaccount (this is the workaround from back when the images couldn't run under an arbitrary user id) and after this change all rabbitmq pods are blocked because they set the seccomp profile.

I understand the OpenShift security model is a bit awkward, but that caught us by surprise in a minor version operator update.

PR #2033, opened 2026-01-12
Update BASELINE_UPGRADE_VERSION to v2.12.0

Zerpet:

## Summary Of Changes

Because version 2.1.0 is from 3 years ago. 2.12.0 is from 1 year ago. The delta between 2.1.0 and 2.19.0+ is simply too great, in Kubernetes versions and CRD changes, therefore, it's highly unlikely that upgrade 2.1.0 -> 2.19.0+ won't cause a rolling upgrade of all rabbits. The same may not be true for "closer" versions e.g. 2.12.0 -> 2.19.0+

## Additional Context

Since v2.12.1, all operator "upgrades" from the baseline v2.1.0 cause a rolling restart of all rabbits 🫠


Zerpet:
I'm merging this for now. I'll work soon in adding another upgrade test for N-1, and update the release header.

PR #2050, opened 2026-02-05
Add startup probe to RabbitMQ

### LINKED ISSUE #2016: Leverage cluster size hint for service readiness
When a cluster is forming, the nodes become available to the service immediately when they come online. If a quorum queue is created early on when only some of the nodes are online, then the quorum queue will not be spread across a full set of nodes, and each node will be quorum critical. This can block cluster upgrades, and requires manual intervention to add the remaining node(s) to the quorum set.

We set `cluster_formation.target_cluster_size_hint = ${number-of-replicas}`. If we can leverage this to ensure the cluster is not available via the service until the condition is met (or failing that, ensure that RabbitMQ itself does not accept connections until the condition is met), it would prevent this situation

Zerpet:
This closes #2016

## Summary Of Changes

- Add a startup probe to check that cluster is fully formed

## Local Testing

Added unit and integration test.

### DIFF
--- a/.github/workflows/build-test-publish.yml
+++ b/.github/workflows/build-test-publish.yml
@@ -13,7 +13,6 @@ on:
     branches: [ "main" ]
 
 env:
-  GO_VERSION: 'stable'
   # Taken from https://github.com/kubernetes-sigs/kind/releases/tag/v0.18.0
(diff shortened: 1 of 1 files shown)


MirahImage (approved):
We should attempt to remember to switch to the new API/cli call once the oldest supported RabbitMQ version has it.

PR #2074, opened 2026-03-18
Migrate kubebuilder v4

Zerpet:

## Summary Of Changes

- Migrated to Kubebuilder V4
- Added support to configure the Service type in system tests
- Moved tools from `go.mod` to Makefile
- Abandoned DockerHub image

## Additional Context

Migrating to Kubebuilder V4 is a long over-due chore. We were unable to use the `kubebuilder` cli at all. We were in a position where we couldn't add new types, nor new webhooks. Any new features from kubebuilder were very complicated to adopt. This PR unblocks the repo to use kubebuilder CLI again.

Adding support for different Service types in system tests allows to run system tests in environments (like my local Rancher Desktop) where NodePort (the default) doesn't route properly. In my setup, LoadBalancer routes correctly to the Pods.

Moving tools to Makefile is motivated to reduce the number of dependencies in `go.mod`. Tool dependencies in `go.mod` pollute the actual dependencies of our operator, and triggers false positives in CVE scanners.

For reasons outside of our control, we won't be producing images to DockerHub anymore. Our "dockerhub" manifest will use the image from GHCR instead. The new image does not require authentication, so we expect no disruption to users upgrading to the next version.

## Local Testing

- [x] Unit and integration tests passing
- [x] System tests passing locally in k3s
- [x] Upgrade test


.devcontainer/devcontainer.json
MirahImage: Did you mean to commit your devcontainer config?

PR #2116, opened 2026-04-24
Skip ImagePullSecrets update when DEFAULT_IMAGE_PULL_SECRETS is empty

### LINKED ISSUE #2098: `reconcileOperatorDefaults` performs unnecessary full Update on RabbitmqCluster CR, causing Helm SSA conflicts
## Describe the bug

`reconcileOperatorDefaults` calls `r.updateRabbitmqCluster()` on the `RabbitmqCluster` CR every reconciliation loop when `ImagePullSecrets` is nil, even if `DEFAULT_IMAGE_PULL_SECRETS` is empty and nothing was actually modified.

This causes two problems:

1. No-op Update: `strings.SplitSeq("", ",")` yields `""`, so nothing is appended, but `r.updateRabbitmqCluster()` is called anyway: <https://github.com/rabbitmq/cluster-operator/blob/d5d81c61f5f6a3d90819c8e0a7d0cec8543e2dcb/internal/controller/reconcile_operator_defaults.go#L25-L36>
2. Full PUT instead of targeted Patch: `r.updateRabbitmqCluster()` sends the entire CR, so the field manager "manager" takes ownership of all fields, not just `imagePullSecrets`. Additionally, Go's `resource.Quantity.MarshalJSON()` always serializes as a JSON string, silently converting e.g. `cpu: 1` (int) to `cpu: "1"` (string) on `x-kubernetes-int-or-string` fields.

The combination causes SSA conflicts on subsequent helm upgrade because the stored JSON type no longer matches what Helm sends.

## To Reproduce

1. Deploy the operator with `DEFAULT_IMAGE_PULL_SECRETS` unset or empty
2. Create a RabbitmqCluster via Helm with a bare integer resource value:

  ```yaml
  spec:
    resources:
      limits:
        cpu: 1
        memory: 1Gi
  ```

3. Wait for at least one reconciliation loop
4. Run helm upgrade on the same release
5. Observe:

  ```console
  UPGRADE FAILED: conflict occurred while applying object <ns>/<name> rabbitmq.com/v1beta1, Kind=RabbitmqCluster: Apply failed with 1 conflict: conflict with "manager" using rabbitmq.com/v1beta1: .spec.resources.limits.cpu
  ```

Only `limits.cpu` conflicts because it's the only value that is a bare YAML integer. Values like 1Gi or 500m are YAML strings and survive the round-trip unchanged.

**Expected behavior**

helm upgrade should succeed. The operator should:

1. Not update when nothing was appended.
2. Not take field ownership of fields it did not intend to modify.

- RabbitMQ: 4.1.3
- RabbitMQ Cluster Operator: 2.19.0
- Kubernetes: 1.32
- Cloud provider or hardware configuration: Azure AKS

## Additional context

The current workaround for our specific use case is to quote integer values in Helm values (`cpu: "1"` instead of `cpu: 1`).

MirahImage:
Avoid unnecessary API updates when no default image pull secrets are configured. Only reconcile ImagePullSecrets if there are changes or the DEFAULT_IMAGE_PULL_SECRETS env var is non-empty and results in actual secrets being added to the spec.

ai-assisted=yes

This closes #2098

### DIFF
--- a/internal/controller/reconcile_operator_defaults.go
+++ b/internal/controller/reconcile_operator_defaults.go
@@ -22,7 +22,7 @@ func (r *RabbitmqClusterReconciler) reconcileOperatorDefaults(ctx context.Contex
 		}
 	}
 
-	if rabbitmqCluster.Spec.ImagePullSecrets == nil {
+	if rabbitmqCluster.Spec.ImagePullSecrets == nil && r.DefaultImagePullSecrets != "" {
 		// split the comma separated list of default image pull secrets from
 		// the 'DEFAULT_IMAGE_PULL_SECRETS' env var, but ignore empty strings.
 		for reference := range strings.SplitSeq(r.DefaultImagePullSecrets, ",") {
@@ -30,8 +30,10 @@ func (r *RabbitmqClusterReconciler) reconcileOperatorDefaults(ctx context.Contex
 				rabbitmqCluster.Spec.ImagePullSecrets = append(rabbitmqCluster.Spec.ImagePullSecrets, corev1.LocalObjectReference{Name: reference})
 			}
 		}
-		if requeue, err := r.updateRabbitmqCluster(ctx, rabbitmqCluster, "image pull secrets"); err != nil {
-			return requeue, err
+		if len(rabbitmqCluster.Spec.ImagePullSecrets) > 0 {
+			if requeue, err := r.updateRabbitmqCluster(ctx, rabbitmqCluster, "image pull secrets"); err != nil {
+				return requeue, err
+			}
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


Zerpet:
I'm ok with this PR, it adds extra checks and avoids an API call when we know it won't meaningfully update RabbitMQ. Earlier we relied on the idempotency of Kubernetes API, and I think it was correct. I also think this PR will only workaround the issue described in #2098, because the actual problem is Helm taking ownership of `RabbitmqCluster` fields, which are entirely owned by the Operator. Nonetheless, I'm ok to merge this PR. The tests adding a fakeClient are quite valuable and may serve to further expand other tests with the same pattern.

Zerpet (approved):
I leave the decision to you whether to merge or not to merge.

PR #2128, opened 2026-04-30
Conditional supporting RBAC creation

Zerpet:
## Summary Of Changes

- Conditionally create RBAC for RabbitMQ StatefulSet, based on rabbit version
- Add `delete` RBAC for `Role` and `RoleBinding` to the Operator ClusterRole

## Additional Context

Related to #1865. Starting in RabbitMQ 4.1, the k8s peer discovery plugin does not require any RBAC in Kubernetes. Earlier versions of this plugin queried k8s API to determine the addresses of its peers. That's not required anymore since 4.1. This PR is an optimisation that removes unnecessary RBACs (SA, Role/Binding).

The annotation based approach has a caveat: since the Operator does not know before hand the version of RabbitMQ, it creates the RBACs, and when it detects that rabbit is `>= 4.1.0`, it removes them. This causes a rolling restart, which slightly delays initial rabbit deployment. The same applies when rabbit is upgraded from 3.13 to 4.2. Once rabbit is upgraded to 4.2, and the annotation is updated, RBACs are deleted and the STS is rolled. This happens because the Pod Spec changes, to remove the SA reference.

Another downside is that the Operator ClusterRole now **requires** `delete` on `Role`, `RoleBinding` and `ServiceAccount`. This broadens the permissions of the operator with a justified reason, but it feels like the operator is effectively a super-admin-user.

I'm ok to not merge this PR, if the ClusterRole change is deemed unacceptable.

## Local Testing

Unit and integration tests.

```
make unit-tests integration-tests
```

Tested locally the following scenarios:

1. Deploy 3.13 -> Verify RBAC is created -> Upgrade 4.2 -> Verify RBAC is deleted
2. Deploy 4.1 -> Verify RBAC is eventually deleted
3. Deploy 4.2 -> Verify RBAC is eventually deleted

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -18,7 +18,6 @@ rules:
   - configmaps
   - persistentvolumeclaims
   - secrets
-  - serviceaccounts
   - services
   verbs:
   - create
@@ -57,6 +56,17 @@ rules:
   - pods/exec
   verbs:
   - create
+- apiGroups:
+  - ""
+  resources:
+  - serviceaccounts
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - update
+  - watch
 - apiGroups:
   - apps
   resources:
@@ -106,6 +116,7 @@ rules:
   - roles
   verbs:
   - create
+  - delete
   - get
   - list
   - update
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -18,6 +18,7 @@ rules:
   - configmaps
   - persistentvolumeclaims
   - secrets
+  - serviceaccounts
   - services
   verbs:
   - create
@@ -56,17 +57,6 @@ rules:
   - pods/exec
   verbs:
   - create
-- apiGroups:
-  - ""
-  resources:
-  - serviceaccounts
-  verbs:
-  - create
-  - delete
-  - get
-  - list
-  - update
-  - watch
 - apiGroups:
   - apps
   resources:
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/rabbitmq/cluster-operator/v2
 go 1.26.2
 
 require (
+	github.com/Masterminds/semver/v3 v3.4.0
 	github.com/cloudflare/cfssl v1.6.5
 	github.com/eclipse/paho.mqtt.golang v1.5.1
 	github.com/go-logr/logr v1.4.3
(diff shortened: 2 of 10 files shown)


Zerpet:
@copilot apply changes based on the comments in [this thread](https://github.com/rabbitmq/cluster-operator/pull/2128#pullrequestreview-4227201900)

config/rbac/role.yaml
```diff
@@ -57,6 +56,17 @@ rules:
   - pods/exec
   verbs:
   - create
+- apiGroups:
+  - ""
+  resources:
+  - serviceaccounts
+  verbs:
+  - create
+  - delete
```
MirahImage: It doesn't look like copilot actually implemented the changes that it claimed, since the delete serviceaccounts rbac is still present.

MirahImage (commented):
Given that we still need the serviceaccount for vault integration, how complicated the logic ends up being, and the fact that we trigger additional rolling upgrades based on version, I'm not sure this is really worthwhile.

Also, did copilot actually implement the changes it claims? I'm not sure it did?

config/rbac/role.yaml
Zerpet: I believe it did, at code level. Updating this file is a matter of running a make target that generates it https://github.com/rabbitmq/cluster-operator/commit/abfca8a5e2a621ee0bd552490054c00e6bbf5beb

Zerpet:
I think it won't require a rolling restart anymore, since we are keeping the service account. The STS modification (to remove the SA reference) is what was causing the rolling restart. I'll test this in a bit.

MirahImage:
> I think it won't require a rolling restart anymore, since we are keeping the service account. The STS modification (to remove the SA reference) is what was causing the rolling restart. I'll test this in a bit.

Assuming it doesn't cause the rolling restart, and that we get the rbac right after the code generation, I'm good to go with this. It's still a bit fragile, but in a way that it takes the approach that works if it doesn't know.

Zerpet:
Confirmed the latest additions by Copilot don't cause a rolling restart. The operator initially creates the Role/Binding, because it doesn't know the version of RabbitMQ, but then it deletes them quietly. I also reviewed the Vault example. Indeed we require the SA for correct Vault integration.

Zerpet:
The [diff might look weird](https://github.com/rabbitmq/cluster-operator/pull/2128/changes#diff-6d4c4e29ec096149781536bb9118fde8242c1901df764e3eda3019ac8503d6c7R109). This is because I rebased and squashed some commits, so the only RBAC change required is adding `delete` to Role/Binding.

MirahImage (approved):

PR #2137, opened 2026-05-08
Expose deprecated features usage via Status

### LINKED ISSUE #1778: Leverage deprecated features subsystem and feature flags

RabbitMQ upgrades require certain number of feature flags to be enabled. Specially in the upgrade from 3.13.x to 4.0.x, we require all stable feature flags enabled, and khepri 3.13 disabled. This information is visible in the management UI, but it's not exposed in the CR status.

Expose in the CR status if a deprecated feature is used, or if not all stable feature flags are enabled.

Automatically enable all feature flags after a rolling update, or as post-deploy steps.

Zerpet:
This closes #1778

## Summary Of Changes

- Expose DeprecatedFeaturesUsed via RabbitmqCluster Status
- Use `Patch` to update RabbitmqCluster Status subresource

## Additional Context

The change to use `Patch` instead of `Update` is required because this PR introduces an additional
controller. After this PR, there will be two controllers managing `RabbitmqCluster`. The main controller
manages all logic regaring child resource lifecycle, setting defaults and reconciling changes. The
"deprecated features" controller manages polling RabbitMQ endpoint that reports this information,
and updating the Status.

The new controller introduced runs periodically every 5 minutes, or earlier if there are updates to
a rabbit CR. This controller does not update, or interact in any way, with the Spec or the metadata.
This controller uses the HTTP API, with the default user, to query the endpoint that checks if any
deprecated feature is in use.

The use of `Patch` is required to avoid controllers stepping on each other's changes.

## Local Testing

```
```

No new system tests as part of this PR.


Zerpet:
Somehow, `0 == 0` is false 🫠 

https://github.com/rabbitmq/cluster-operator/actions/runs/25555978954/job/75015655506?pr=2137#step:9:239

internal/rabbitmqclient/factory.go
```diff
@@ -14,7 +14,7 @@ import (
 	"context"
 	"fmt"
 
-	rabbithole "github.com/michaelklishin/rabbit-hole/v2"
+	rabbithole "github.com/michaelklishin/rabbit-hole/v3"
```
MirahImage, line 17: Amazed we hadn't updated to v3 before now.

MirahImage (approved):

PR #2151, opened 2026-05-26
feat: move defaulting logic to mutating webhook

### LINKED ISSUE #2121: Move defaulting logic to a mutating webhook
## Summary

Currently, defaulting logic (e.g., for `spec.imagePullSecrets`) is implemented in the reconciliation loop. This approach leads to several issues:

- Reconciliation-based defaulting can trigger unnecessary updates when no actual change is needed, resulting in full CR updates and possible Server-Side Apply (SSA) ownership conflicts, as reported in issue #2098.
- Users or external tools (like Helm) may wish to retain control over certain fields; the current implementation can unintentionally take field ownership from them, causing SSA conflicts, especially for CR fields with ambiguous YAML types (e.g., integer vs string in resource quantities).
- The best practice in Kubernetes is to perform defaulting at admission time via a mutating admission webhook, not during reconciliation.

## Proposal

Move all defaulting logic for `RabbitmqCluster` resources to a mutating admission webhook. 
- Defaulting should only happen when values are unset (e.g., `imagePullSecrets`).
- The webhook should read operator environment variables (e.g., `DEFAULT_IMAGE_PULL_SECRETS`) to determine defaults, as supported by operator-scaffolded webhooks.
- This will ensure defaults are only set when appropriate, and ownership of the field remains with the user if they explicitly set a value.

## Benefits
- Avoids unnecessary CR updates during reconciliation
- Eliminates field ownership conflicts and SSA merge errors with tools like Helm
- Aligns with Kubernetes best practices and Kubebuilder's webhook patterns
- Simpler, more predictable user experience: defaults are set at creation/admission time

## Context
Related to issue #2098 and recent discussions about SSA and field ownership. See the operator controller code in `internal/controller/reconcile_operator_defaults.go` for current defaulting implementation.

amaanx86:
This closes #2121

## Summary Of Changes

Defaulting of spec.image, spec.imagePullSecrets, and
spec.secretBackend.vault.defaultUserUpdaterImage has been moved from
reconcileOperatorDefaults into a mutating admission webhook. The webhook
sets a field only when it is unset, so any value the user provides is
preserved.

The reconciler no longer sends a full PUT on every loop to apply
defaults. It now only updates the cluster when ControlRabbitmqImage is
true, which is enforcement rather than defaulting and intentionally
keeps field ownership in that case. This fixes the SSA conflict
described in #2098, where the repeated PUT caused the manager field
manager to take ownership of all fields including resource quantities,
breaking subsequent helm upgrades.

The webhook runs in the same container as the operator, so
failurePolicy: Fail is safe and avoids the need for a fallback path in
the reconciler.

## Additional Context

The PROJECT file had its repo field set to
github.com/rabbitmq/cluster-operator while go.mod declares the module
as github.com/rabbitmq/cluster-operator/v2. This caused kubebuilder to
generate broken import paths when scaffolding the webhook. The fix was
confirmed with Zerpet before proceeding (see issue #2121).

Three issues were found and fixed while testing on a kind cluster:

- config/webhook/manifests.yaml was not generated by make manifests.
  Added a controller-gen webhook invocation to the Makefile.
- manager_webhook_patch.yaml used JSON Patch append on volumeMounts and
  volumes, which do not exist in manager.yaml. Changed to create them
  as new arrays.
- The webhook Service selector used kubebuilder default labels that do
  not match the operator pod. Fixed to use
  app.kubernetes.io/name: rabbitmq-cluster-operator.

## Local Testing

Tested on a kind cluster with cert-manager v1.15.1. Created a
RabbitmqCluster with spec: {} and confirmed spec.image was set to the
default by the webhook at admission time. Unit tests pass with
make just-unit-tests.

### DIFF
--- a/internal/controller/reconcile_operator_defaults_internal_test.go
+++ b/internal/controller/reconcile_operator_defaults_internal_test.go
@@ -15,6 +15,7 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gstruct"
 	rabbitmqv1beta1 "github.com/rabbitmq/cluster-operator/v2/api/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -25,31 +26,18 @@ import (
 )
 
 var _ = Describe("reconcileOperatorDefaults", func() {
-	When("DEFAULT_IMAGE_PULL_SECRETS is empty", func() {
-		It("does not Update the cluster when ImagePullSecrets is nil", func() {
+	When("ControlRabbitmqImage is false", func() {
+		It("does not call Update", func() {
 			ctx := context.Background()
 
 			scheme := runtime.NewScheme()
 			Expect(corev1.AddToScheme(scheme)).To(Succeed())
 			Expect(rabbitmqv1beta1.AddToScheme(scheme)).To(Succeed())
 
-			presetUserUpdaterImage := "preset-uu-image:1.0"
 			cluster := &rabbitmqv1beta1.RabbitmqCluster{
 				ObjectMeta: metav1.ObjectMeta{
-					Name:      "test-no-defaults",
+					Name:      "test-no-update",
 					Namespace: "default",
-				},
-				Spec: rabbitmqv1beta1.RabbitmqClusterSpec{
-					// Pre-set Image and DefaultUserUpdaterImage so the only branch
-					// in reconcileOperatorDefaults that could trigger an Update is
-					// the ImagePullSecrets one we are testing.
-					Image: "preset-image:1.0",
-					SecretBackend: rabbitmqv1beta1.SecretBackend{
-						Vault: &rabbitmqv1beta1.VaultSpec{
-							DefaultUserPath:         "some-path",
-							DefaultUserUpdaterImage: &presetUserUpdaterImage,
-						},
-					},
 				},
 			}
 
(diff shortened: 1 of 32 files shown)


amaanx86:
@Zerpet happy to get your thoughts on this when you have a moment. Marking ready for review.

internal/controller/reconcile_operator_defaults_internal_test.go
```diff
@@ -70,25 +58,60 @@ var _ = Describe("reconcileOperatorDefaults", func() {
...
 			Expect(requeue).To(BeZero())
-
-			By("not calling Update on the cluster")
 			Expect(updateCallCount).To(Equal(0))
+		})
+	})
+
+	When("ControlRabbitmqImage is true", func() {
+		It("enforces the default image and user updater image", func() {
+			ctx := context.Background()
```
Zerpet: Instead of using `context.Background()`, use `SpecContext` from the `It` node. Like so:

```
It("...", func(ctx SpecContext) {
  // tests
}
```

See Ginkgo documentation for this: https://onsi.github.io/ginkgo/#spec-timeouts-and-interruptible-nodes

internal/webhook/v1beta1/rabbitmqcluster_webhook.go
```diff
@@ -0,0 +1,73 @@
...
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+
+	rabbitmqcomv1beta1 "github.com/rabbitmq/cluster-operator/v2/api/v1beta1"
+)
+
+var rabbitmqclusterlog = logf.Log.WithName("rabbitmqcluster-resource")
```
Zerpet: I'd change the log name to be more specific, like `rabbitmqcluster-webhook`

Zerpet (changes requested):
It's looking good. However, integration tests are broken due to the introduction of the webhook. It's probably a matter of adding the `WebhookInstallOptions` in the `controller/` suite, as it was done for the webhook suite.

Blocking merge until the tests are green.

Zerpet:
I noticed you do not seem to have the CLA signed.

Please note that since the 28th of July 2025, all external contributors to RabbitMQ will have to sign a CLA again.

You have detailed instructions in [1][2].

Note companies that contribute to RabbitMQ only need to sign the CLA once, there is no need to make every
contributing employee to do so separately.

https://github.com/rabbitmq/cla
https://github.com/rabbitmq/rabbitmq-server/blob/main/CONTRIBUTING.md

amaanx86:
thank you @Zerpet for your detailed review .. I will work on the fixes and meanwhile sign the CLA

internal/controller/reconcile_operator_defaults_internal_test.go
amaanx86: `764c363a test: use SpecContext in It nodes instead of context.Background()` fixed in this commit

internal/webhook/v1beta1/rabbitmqcluster_webhook.go
amaanx86: `47686741 fix: rename log name from rabbitmqcluster-resource to rabbitmqcluster-webhook` fixed in this commit

amaanx86:
Hi @Zerpet CLA signing done! I request you to please review the changes now

amaanx86:
Got the e2e failure notifications. Investigated and pushed fixes:

- Moved cert-manager install before operator deploy in all CI jobs (kubectl_tests, system_tests, system_tests_oldest_k8s, test_doc_examples)
- Added webhook and certmanager resources to config/installation kustomize overlay along with the cert volume patch and CA injection replacements
- Fixed wrong resource names in e2e assertions (webhook-service, mutating-webhook-configuration)
- InstallCertManager now waits for all three components and probes the webhook until it actually accepts connections, not just until the deployment is Available
- Added ClusterRoleBinding cleanup in AfterAll to avoid failures on re-runs

All 4 e2e specs pass locally.

Zerpet:
I approved the new changes, but I think it won't be effective since Actions use the PR code from the base branch. I'll verify locally.

amaanx86:
Thank you .. btw prev failing e2e tests are now passing - rest you can validate and let me know your feedback

amaanx86:
image

@Zerpet I pushed commit `efbe05ae` to prevent CodeQL failure

amaanx86:
image

This should also be fixed in the next run -- rest were all passing

amaanx86:
Hi @Zerpet, the "Local system tests (min k8s)" job is still failing: https://github.com/rabbitmq/cluster-operator/actions/runs/26626230938/job/78466999322

The `persists messages` spec times out in `waitForRabbitmqUpdate` after pod deletion. Same job passes on stable k8s, was passing on main before the webhook. The webhook adds a round-trip on every `r.Update()` call during reconcile which seems to push the pod restart wait over the limit on v1.29.12.

My plan is to add an `Eventually` in the persistence `BeforeEach` to handle the retry case (when `FlakeAttempts` retries, the previous cluster is still terminating and the create hits `409 AlreadyExists`). Does that sound right to you, or do you think we need to address the webhook latency itself (e.g. `failurePolicy: Ignore`)?

Zerpet:
This test is known to be flaky. To the point that I'm considering just deleting it. I re-triggered the job. It is true that the flake attempts is not very effective. Any improvements on this front are welcome!

amaanx86:
> This test is known to be flaky. To the point that I'm considering just deleting it. I re-triggered the job. It is true that the flake attempts is not very effective. Any improvements on this front are welcome!

I'll open an issue and separately raise PR

amaanx86:
@Zerpet are we good to merge this now ?

Zerpet:
The code overall looks good. I'll probably squash the commits during merge. I won't merge this for now, because we have a code freeze next Tuesday and a patch release later in the month.

This change requires changes in other repos, specifically, in the OLM packaging for Openshift native package. Updating OLM package requires further changes, that don't need to be part of this PR, but must be present before our next release.

Fear not, this PR will be accepted.

amaanx86:
Thanks for the update 🫡

Zerpet:
@amaanx86 we are ready to merge this PR. Can you rebase on top of the latest main and resolve the conflicts?

amaanx86:
@Zerpet done, rebased on top of latest main with a clean linear history. Ready when you are.

Zerpet (approved):
I'll merge after CI passes 👍 Thank you for contributing!

Zerpet:
I pushed a fix for CI. It should pass now.

amaanx86:
Let me take a look .. It's still failing

amaanx86:
Took a look at the failing json-log test. looks like wait-for-webhook only checks the CA bundle but doesn't actually verify the service is reachable. there's a small gap between the pod becoming ready and kube-proxy wiring up the endpoint, so we hit connection refused. 

```
fix: add a quick probe to the service inside wait-for-webhook so it only returns once we can actually connect.
```

lmk if you want me to push

PR #2191, opened 2026-07-03
Wait for OLM webhook service endpoints before creating RabbitmqCluster in OLM test

Zerpet:
## Summary
- The OLM workflow's `test-olm-package` job has been failing with `dial tcp ...:9443: connect: connection refused` when the apiserver calls the mutating webhook while creating the sample `RabbitmqCluster`.
- The test only waited for the `Subscription` to reach `AtLatestKnown`, which confirms the CSV install succeeded but not that the operator pod is up and reachable behind the webhook Service — a race similar to one already fixed in the e2e suite.
- Added a wait for the `rabbitmq-cluster-operator-service` endpoints (the Service OLM generates from the CSV's `webhookdefinitions.deploymentName`) to have a real address before creating the `RabbitmqCluster` object.

## Test plan
- [x] `go vet ./olm/...` passes
- [x] Confirm the `olm.yml` workflow's `test-olm-package` job passes on the next run (webhook race can't be reproduced without the full OLM/Kind pipeline)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/olm/test/deploy_test.go
+++ b/olm/test/deploy_test.go
@@ -69,6 +69,35 @@ var _ = Describe("Deploy", func() {
 		container := deploy.Spec.Template.Spec.Containers[0]
 		Expect(container.Image).To(ContainSubstring("rabbitmqoperator/cluster-operator"))
 
+		By("waiting for the cluster operator webhook service endpoints to be ready")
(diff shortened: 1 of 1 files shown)


Zerpet:
Workflow run passes: https://github.com/rabbitmq/cluster-operator/actions/runs/28665473968

PR #2207, opened 2026-07-22
Secure operator metrics and pprof endpoints by default.

MirahImage:
[ai-assisted=yes]

### DIFF
--- a/config/default/manager_metrics_patch.yaml
+++ b/config/default/manager_metrics_patch.yaml
@@ -2,3 +2,6 @@
 - op: add
   path: /spec/template/spec/containers/0/args/0
   value: --metrics-bind-address=:8443
+- op: add
+  path: /spec/template/spec/containers/0/args/1
+  value: --metrics-secure=true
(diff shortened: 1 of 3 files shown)


config/default/kustomization.yaml
```diff
@@ -24,7 +24,7 @@ resources:
 # Only Pod(s) running a namespace labeled with 'metrics: enabled' will be able to gather the metrics.
 # Only CR(s) which requires webhooks and are applied on namespaces labeled with 'webhooks: enabled' will
 # be able to communicate with the Webhook Server.
-#- ../network-policy
+- ../network-policy
```
Zerpet: I did not add network policies to the Top-op metrics endpoint [1], back when migrating to kubebuilder v4. I only added the manager metrics patch.

I think adding network policies can break a considerable amount of users, because after this change, any namespace missing the label `webhook: "enabled"` won't be able to create `RabbitmqCluster` objects.

I would prefer to add a examples in the website docs and document how to enable network policies.

[1] https://github.com/rabbitmq/messaging-topology-operator/blob/main/config/default/kustomization.yaml#L30-L34

Zerpet (approved):

PR #2212, opened 2026-07-28
Flake Fix persists messages system test

### LINKED ISSUE #2152: test: improve reliability of flaky "persists messages" system test

The `Persistence / persists messages` system test is known to be flaky (discussed in #2151). Two distinct failure modes have been identified:

1. **`SpecTimeout` race on slow clusters (e.g. min k8s v1.29.12):** The spec has a 3-minute `SpecTimeout` that covers the full lifecycle — `BeforeEach` (cluster creation + `waitForRabbitmqRunning`) and the `It` body (`waitForRabbitmqUpdate` after pod deletion). On slower environments, or when the mutating webhook adds a round-trip to every `r.Update()` call during reconcile, the pod-restart cycle inside `waitForRabbitmqUpdate` regularly exceeds what is left of the budget after cluster creation, causing a timeout.

2. **409 AlreadyExists on `FlakeAttempts` retry:** When `FlakeAttempts(3)` retries the spec, the `AfterEach` deletion of the previous cluster may not have completed by the time `BeforeEach` runs again. The `createRabbitmqCluster` call fails immediately with `409 AlreadyExists`, so the retry itself fails before the test body even runs — making `FlakeAttempts` ineffective.

Two targeted changes to `test/system/system_test.go`:

1. **Wrap cluster creation in `BeforeEach` with `Eventually`** so that on a `FlakeAttempts` retry the setup polls until the previous cluster is fully gone before proceeding:

```go
Eventually(func(g Gomega) {
    cluster = newRabbitmqCluster(namespace, "persistence-rabbit")
    g.Expect(createRabbitmqCluster(ctx, rmqClusterClient, cluster)).To(Succeed())
}, podCreationTimeout, 5*time.Second).Should(Succeed())
```

2. **Increase `SpecTimeout` from 3 minutes to 5 minutes** on the `persists messages` It node, consistent with other slower specs in the same file (e.g. the TLS spec uses `SpecTimeout(time.Minute*5)`), to give the reconcile loop sufficient headroom when webhook round-trips are in the path.

- **Delete the test entirely** — mentioned in #2151 as an option; avoids the flake but loses coverage of the persistence guarantee.
- **Set `failurePolicy: Ignore` on the mutating webhook** — reduces webhook-induced latency during reconcile but changes production behaviour and is a broader change than necessary.
- **Increase `FlakeAttempts`** — does not address either root cause; the 409 AlreadyExists still makes retries fail immediately.

**Additional context**

Discussed in #2151 (comment by @Zerpet): https://github.com/rabbitmq/cluster-operator/pull/2151#issuecomment-4574039429

The failure is consistently reproducible on the "Local system tests (min k8s)" CI job (k8s v1.29.12) and was not observed on stable k8s prior to the mutating webhook being introduced in that PR.

MirahImage:
Allow cluster creation time to succeed.

This closes #2152

### DIFF
--- a/test/system/system_test.go
+++ b/test/system/system_test.go
@@ -270,7 +270,7 @@ CONSOLE_LOG=new`
 
 		BeforeEach(func(ctx SpecContext) {
 			cluster = newRabbitmqCluster(namespace, "persistence-rabbit")
-			Expect(createRabbitmqCluster(ctx, rmqClusterClient, cluster)).To(Succeed())
+			Eventually(createRabbitmqCluster(ctx, rmqClusterClient, cluster)).Should(Succeed())
 
 			waitForRabbitmqRunning(cluster)
 
@@ -312,7 +312,7 @@ CONSOLE_LOG=new`
 				Expect(pvc.OwnerReferences).To(HaveLen(1))
 				Expect(pvc.OwnerReferences[0].Name).To(Equal(cluster.Name))
 			})
-		}, SpecTimeout(time.Minute*3))
+		}, SpecTimeout(time.Minute*5))
 	})
 
 	Context("Persistence expansion", Label("persistence_expansion"), func() {


Zerpet (commented):
I'm a bit unsure this actually fixes the flake. In workflow runs where the test flakes, the retries are ineffective becase the "old" `RabbitmqCluster` does not delete. It is pending deletion and blocked on an unknown condition. In fact, subsequent runs (thanks to Ginkgo's `FlakeAttemps` decorator) fail with `RabbitmqCluster xxx is being terminated` or `... already exists`.
Of course, logs showing the flake have already rotated 🙁

MirahImage:
Yeah, I haven't seen the flake recently, which of course made it harder to validate. I am happy to leave this sitting around as a draft for a while until the flake reoccurs and we can inspect what happened again.

Zerpet:
Closing in favour of #2217

PR #2220, opened 2026-08-07
Ensure StatefulSetSelector matches deployed labels.

### LINKED ISSUE #2218: StatefulSet selector override causes permanent reconciliation failure (Forbidden / selector mismatch)
---
title: "StatefulSet selector override causes permanent reconciliation failure (Forbidden / selector mismatch)"
labels: bug
---

## Describe the bug

Setting `spec.override.statefulSet.spec.selector` on a `RabbitmqCluster` causes the operator to fail reconciliation **permanently**, and the failure cannot be recovered by removing the override.

The operator applies the override selector when **building** a new StatefulSet object (`internal/resource/statefulset.go`, `Build()`), but that same object is then submitted in `Update()` payloads. Since `spec.selector` is immutable on StatefulSets, the API server rejects every update. Depending on the state of the StatefulSet at the time of the change, the operator hits one of two errors:

```
StatefulSet.apps "test-cluster-server" is invalid: spec.template.metadata.labels: Invalid value: map[string]string{...}: `selector` does not match template `labels`
```

or

```
StatefulSet.apps "test-cluster-server" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', 'updateStrategy', 'persistentVolumeClaimRetentionPolicy' and 'minReadySeconds' are forbidden
```

Both are manifestations of the same root cause: the operator carries an immutable field (selector) into StatefulSet update payloads without comparing it against the existing object.

## To Reproduce

1. Deploy the cluster-operator and create a `RabbitmqCluster`:

```yaml
kind: RabbitmqCluster
metadata:
  name: test-cluster
  namespace: rabbitmq-system
spec:
  replicas: 1
```

2. Set the override selector:

```bash
kubectl patch rabbitmqcluster test-cluster -n rabbitmq-system \
  --type merge \
  -p '{"spec":{"override":{"statefulSet":{"spec":{"selector":{"matchLabels":{"app.kubernetes.io/name":"test-cluster-CHANGED"}}}}}}}'
```

3. Delete the StatefulSet so the operator rebuilds it through `Build()` (the selector is only picked up on the create path; a plain patch on a healthy cluster does not trigger the bug because `applyStsOverride` does not touch the selector during updates):

```bash
kubectl delete statefulset test-cluster-server -n rabbitmq-system
```

4. Observe `ReconcileSuccess` flips to `False` and stays there:

```bash
kubectl get rabbitmqcluster test-cluster -n rabbitmq-system \
  -o jsonpath='{.status.conditions[?(@.type=="ReconcileSuccess")].status}'
# -> False
```

5. **Recovery attempts that fail**:
   - removing the override (`kubectl patch ... remove /spec/override/statefulSet/spec/selector`)
   - annotating the CR to force a requeue
   - `kubectl rollout restart deploy/rabbitmq-cluster-operator`

   The operator keeps reconciling the same broken object; `ReconcileSuccess` stays `False` indefinitely. The only recovery is deleting and recreating the `RabbitmqCluster` CR.

**Expected behavior**

A user should not be able to get the cluster into a state from which the operator cannot recover:
- either the admission webhook should reject `selector` changes on the StatefulSet override (as it is immutable), or
- the operator should exclude immutable fields from the update payload (compare against the existing object), or
- it should perform a delete-and-recreate of the StatefulSet when the selector actually changes (the code even has `UpdateMayRequireStsRecreate()` returning `true`, but the recreate logic still ends up submitting the stale selector).

- RabbitMQ: 3.11 
- RabbitMQ Cluster Operator: latest
- Kubernetes: v1.23.0

MirahImage:
This closes #2218

### DIFF
--- a/internal/controller/rabbitmqcluster_controller.go
+++ b/internal/controller/rabbitmqcluster_controller.go
@@ -204,6 +204,7 @@ func (r *RabbitmqClusterReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	for _, builder := range builders {
 		obj, err := builder.Build()
 		if err != nil {
+			r.setReconcileSuccess(ctx, rabbitmqCluster, corev1.ConditionFalse, "Error", err.Error())
 			return ctrl.Result{}, err
 		}
 
--- a/internal/controller/rabbitmqcluster_controller.go
+++ b/internal/controller/rabbitmqcluster_controller.go
@@ -204,7 +204,11 @@ func (r *RabbitmqClusterReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	for _, builder := range builders {
 		obj, err := builder.Build()
 		if err != nil {
-			r.setReconcileSuccess(ctx, rabbitmqCluster, corev1.ConditionFalse, "Error", err.Error())
+			reason := "Error"
+			if errors.Is(err, resource.ErrInvalidStatefulSetSelectorOverride) {
+				reason = "InvalidStatefulSetSelectorOverride"
+			}
+			r.setReconcileSuccess(ctx, rabbitmqCluster, corev1.ConditionFalse, reason, err.Error())
 			return ctrl.Result{}, err
 		}
 
--- a/internal/controller/rabbitmqcluster_controller_test.go
+++ b/internal/controller/rabbitmqcluster_controller_test.go
@@ -25,13 +25,20 @@ import (
 	. "github.com/onsi/gomega"
 	. "github.com/onsi/gomega/gstruct"
 	rabbitmqv1beta1 "github.com/rabbitmq/cluster-operator/v2/api/v1beta1"
+	controllers "github.com/rabbitmq/cluster-operator/v2/internal/controller"
 	"github.com/rabbitmq/cluster-operator/v2/internal/status"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	discoveryv1 "k8s.io/api/discovery/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	k8sresource "k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
+	ctrl "sigs.k8s.io/controller-runtime"
 	runtimeClient "sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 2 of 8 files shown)


Zerpet:
**🤖 AI code review** — this was generated by Claude Code (an AI coding agent) and is being posted by a human reviewer who is checking these findings before acting on them. Treat as input to review, not as an authoritative verdict.

Ran a high-effort automated review of this PR's new StatefulSet-selector validation guard. 5 findings, all independently verified:

### Correctness

1. **`internal/webhook/v1beta1/rabbitmqcluster_webhook.go:78`** — `ValidateUpdate` ignores the old object and rejects *any* update to a CR whose already-persisted StatefulSet selector override doesn't match its pod template labels, with no grandfathering. A CR created before this webhook that already has a mismatched selector override (previously accepted) would have every future update rejected — even unrelated ones like bumping replicas or changing the image tag — until the mismatch itself is fixed in the same request.

2. **`internal/metadata/label.go:58`** — `ValidateStatefulSetSelector` only runs when `spec.override.statefulSet.spec.selector` is explicitly set. A pod-template-label-only override (e.g. overwriting `app.kubernetes.io/name` via `spec.override.statefulSet.spec.template.metadata.labels`) breaks the default selector match but is never validated by either the webhook or `StatefulSetBuilder.Build()`, since both skip validation when `override.Spec.Selector` is nil. The StatefulSet update is then rejected by the API server, leaving the cluster stuck in `ReconcileSuccess=False` — the exact failure mode this guard is meant to prevent.

3. **`internal/metadata/label.go:62`** — A non-nil but empty selector override (`spec.override.statefulSet.spec.selector: {}`) converts to `labels.Everything()` via `metav1.LabelSelectorAsSelector`, so it always "matches" and passes validation, even though Kubernetes rejects empty StatefulSet selectors outright. Same permanent-failure outcome as #2.

### Cleanup

4. **`internal/webhook/v1beta1/rabbitmqcluster_webhook.go:176`** vs **`internal/resource/statefulset.go:96-99`** — the block extracting `templateLabelOverrides` from `override.Spec.Template` is duplicated verbatim between the webhook and the builder. A future change to how template label overrides are represented is likely to update only one copy, silently reintroducing a webhook/builder mismatch.

5. **`internal/controller/rabbitmqcluster_controller.go:207`** — the new `setReconcileSuccess` call for `builder.Build()` errors uses the generic reason `"Error"`, unlike every other failure path in this function (`FailedReconcilePVC`, `InvalidConfiguration`, `FailedVersionAnnotation`, `FailedCLICommand`). Dashboards/alerts filtering on `status.conditions[].reason` can't distinguish a selector-override misconfiguration from any other builder failure.

Findings 1 and 2 share the same root cause: validation only considers the *new* object's selector-vs-template state in isolation, and only when a selector override is explicitly present.

Zerpet (approved):

Issue #1319, opened 2023-04-10
Support directly mounted TLS certs


max-frank:
**Is your feature request related to a problem? Please describe.**
In some cases e.g, when using [cert-managers CSI driver](https://cert-manager.io/docs/usage/csi/) TLS keys+CA cert can be directly mounted to the pods. Also the now possibly outdated[ mtls inter node example](https://github.com/rabbitmq/cluster-operator/blob/main/docs/examples/mtls-inter-node/rabbitmq.yaml#L17) mounts secrets manually.  Right now such cases are not really supported since many feature rely on `tls.secretName` being set for tls to be considered enabled and `tls.caSecretName` for mtls e.g.,

- [TLS service ports](https://github.com/rabbitmq/cluster-operator/blob/main/internal/resource/service.go#L183)
- [Disabling non TLS ports](https://github.com/rabbitmq/cluster-operator/blob/main/internal/resource/service.go#L118)
- etc.

Meaning that while it is possible to configure the rabbitmq nodes themselves to run with TLS/mtls using such directly mounted secrets, by also configuring the TLS config directly as is done in the MTLS example, it is not possible to utilise the operators other features related to TLS.

**Describe the solution you'd like**

This could be solved with backwards compatibility by adding an extra configuration option to force TLS to be enabled even in the absence of `tls.secretName` e.g., `tls.enabled`, if necessary a second option could also be added as switch for the MTLS features handled by the operator.

**Describe alternatives you've considered**
Another option could be instead of having just a simple flag to also allow to pass file paths instead of just secret refs
e.g., `tls.secretPath` and `tls.caSecretPath` and then treat them essentially the same as their secret ref equivalents.

**Additional context**
Also note that using e.g., the cert-manager CSI driver allows you to easily create TLS secrets specific to the pod running. So each cluster node gets their own cert automatically making it a bit easier to scale. So supporting this use case would be beneficial overall.

Example config with cert-manager CSI driver
```yaml
apiVersion: cert-manager.io/v1
kind: Issuer
metadata:
  name: rabbitmq-ca-issuer
  namespace: test
spec:
  selfSigned: {}
---
apiVersion: cert-manager.io/v1
kind: Certificate
metadata:
  name: rabbitmq-ca
  namespace: test
spec:
  isCA: true
  secretName: rabbitmq-ca
  commonName: rabbitmq-ca
  issuerRef:
    name: rabbitmq-ca-issuer
    kind: Issuer
    group: cert-manager.io
[...]

PR #2231, opened 2026-08-24
Automate inter-node mTLS via cert-manager CSI driver

Zerpet:
## Summary

Closes part of #1319: automates inter-node (Erlang distribution) mTLS
using the cert-manager CSI driver, so `spec.tls.interNode` replaces
the hand-wired `envConfig` + `spec.override.statefulSet` + out-of-band
ConfigMap flow shown in `docs/examples/mtls-inter-node`.

```yaml
spec:
  tls:
    interNode:
      enabled: true
      issuerRef:
        name: some-ca
```

- **API**: `spec.tls.interNode.{enabled, issuerRef}` and
  `InterNodeTLSEnabled()`.
- **ConfigMap**: generates the fixed strict-mTLS Erlang distribution
  terms and appends the required `rabbitmq-env.conf` variables
  without clobbering any existing `spec.rabbitmq.envConfig`.
- **StatefulSet**: mounts a per-pod certificate from the
  `csi.cert-manager.io` CSI driver, with a correctly-scoped SAN per
  replica (no per-replica manifest edits needed to scale) and the
  `fs-group` attribute required for the mounted files to be readable.
- **Docs**: `docs/examples/mtls-inter-node` rewritten to the native
  flow; the old hand-wired files are removed.

Out of scope (left open on #1319): the client-facing half of the
issue (`tls.enabled` / `tls.secretPath`, TLS service ports without
`tls.secretName`).

## Test plan

- [x] `gmake fmt vet`
- [x] `gmake just-unit-tests` (322/322 in the Resource suite)
- [x] `gmake just-integration-tests` (73/73 envtest specs, including
      new CEL/defaulting coverage for the API)
- [x] `gmake manifests generate` produces no diff
- [x] Live verification on a Kubernetes cluster with cert-manager and
      cert-manager-csi-driver installed:
  - Per-pod `CertificateRequest`s issued by the CSI driver, one per
    replica, with correctly-scoped SANs
  - Mounted certificate file ownership/readability
  - `rabbitmq-env.conf` composed correctly, existing `envConfig`
    preserved
  - `rabbitmqctl eval 'init:get_argument(proto_dist).'` →
    `{ok,[["inet_tls"]]}`
  - `rabbitmq-diagnostics cluster_status` shows all nodes
    running, clustering listener reporting `protocol: clustering/ssl`
  - Scaling 3 → 5 replicas: new pods get correctly-scoped certificates
    automatically, no CR edit beyond `replicas`


MirahImage:
Overall, looks great to me. Claude pointed out one minor point that might be worth thinking about:

**API ergonomics**: disabling requires deleting the block, you can't just set `enabled: false`. `issuerRef.name` is required with minLength:1 unconditionally on the interNode struct, so `{enabled: false}` with no issuerRef is rejected by the CRD schema — you must delete the whole interNode key to turn it off, you can't leave a dormant, pre-staged config with enabled: false. This is a deliberate, well-tested tradeoff (avoiding a CEL cross-field rule), but it's an inconsistency in a CRD where "toggle enabled" is otherwise the norm; a one-line CEL rule (self.enabled == false || has(self.issuerRef)) would fix it if this bites users. Minor, not blocking.

MirahImage (approved):

Zerpet:
From the commit message 2ae32886b1a8a6fbe13cc94d68965e872d524eee, it looks like, since `enabled` has `omitempty`, a CEL rule like the above would fail with an error on a valid CRD object. For example, the following is valid at the moment (up to debate whether it should):

<details><summary>minimal.yaml</summary>

```yaml
kind: RabbitmqCluster
metadata:
  name: mtls-inter-node
spec:
  image: rabbitmq:4.3.4-management
  replicas: 3
  tls:
    interNode:
      // note: enabled is omitted, defaulting to false
      issuerRef:
        name: mtls-inter-node-ca-issuer
```

</details>

The idea of marking `issuerRef` with min-length 1 is to prevent an empty string as a valid value. The field is not optional, however, empty string would be a valid value without min-length 1, and empty string is not a valid value for `Issuer` name.

It is possible to disable inter-node TLS by setting `enabled: false` and leaving `issuerRef.name` with its original value, or with a place-holder string. Deleting the block entirely is also fine.

I personally don't see much value in leaving a half-set dormant value like as follows:

<details><summary>invalid.yaml</summary>

```yaml
kind: RabbitmqCluster
metadata:
  name: mtls-inter-node
spec:
  replicas: 3
  tls:
    interNode:
      enabled: false
```

</details>

It doesn't make much sense to me, because toggling `enabled: true` is not valid (missing issuer ref name). Moreover, what's that CR trying to express? A cluster with disabled inter-node TLS? That's already the default in RabbitMQ and the Cluster Operator. In case one would want explicit disabling inter-node, an `nil` would do:

<details><summary>nil-inter-node.yaml</summary>

```yaml
kind: RabbitmqCluster
metadata:
  name: mtls-inter-node
spec:
  replicas: 3
  tls:
    interNode: nil
```

</details>

PR #2242, opened 2026-09-08
Fix linter issues.

MirahImage:
Purely automated fixes generated by running `golangci-lint run --fix`.

### DIFF
--- a/internal/resource/statefulset.go
+++ b/internal/resource/statefulset.go
@@ -1096,7 +1096,7 @@ func podHostNames(instance *rabbitmqv1beta1.RabbitmqCluster) string {
 	var altNames strings.Builder
 	var i int32
 	for i = range ptr.Deref(instance.Spec.Replicas, 1) {
-		fmt.Fprintf(&altNames, ",%s", fmt.Sprintf("%s-%d.%s.%s", instance.ChildResourceName(stsSuffix), i, instance.ChildResourceName(headlessServiceSuffix), instance.Namespace))
+		fmt.Fprintf(&altNames, ",%s-%d.%s.%s", instance.ChildResourceName(stsSuffix), i, instance.ChildResourceName(headlessServiceSuffix), instance.Namespace)
 	}
 	return strings.TrimPrefix(altNames.String(), ",")
 }
--- a/internal/resource/statefulset.go
+++ b/internal/resource/statefulset.go
@@ -77,10 +77,8 @@ func (builder *StatefulSetBuilder) Build() (client.Object, error) {
 	}
 
 	sts := &appsv1.StatefulSet{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      builder.Instance.ChildResourceName(stsSuffix),
-			Namespace: builder.Instance.Namespace,
-		},
+		Name:      builder.Instance.ChildResourceName(stsSuffix),
+		Namespace: builder.Instance.Namespace,
 		Spec: appsv1.StatefulSetSpec{
 			ServiceName: builder.Instance.ChildResourceName(headlessServiceSuffix),
 			Selector: &metav1.LabelSelector{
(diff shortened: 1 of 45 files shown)


internal/scaling/scaling_suite_test.go
```diff
@@ -199,7 +195,7 @@ func (matcher *updateActionMatcher) Match(actual any) (bool, error) {
 	}
 
 	updatedObject := reflect.ValueOf(action.GetObject()).Elem()
-	objMeta, ok := updatedObject.FieldByName("ObjectMeta").Interface().(metav1.ObjectMeta)
+	objMeta, ok := reflect.TypeAssert[metav1.ObjectMeta](updatedObject.FieldByName("ObjectMeta"))
```
Zerpet, line 198: Cool!

internal/webhook/v1beta1/rabbitmqcluster_webhook.go
```diff
@@ -177,7 +177,7 @@ func validateStatefulSetSelectorOverride(cluster *rabbitmqcomv1beta1.RabbitmqClu
 
 	var templateLabelOverrides map[string]string
 	if override.Spec.Template != nil && override.Spec.Template.EmbeddedObjectMeta != nil {
-		templateLabelOverrides = override.Spec.Template.EmbeddedObjectMeta.Labels
+		templateLabelOverrides = override.Spec.Template.Labels
```
Zerpet, line 180: I think there was a reason for `EmbeddedObjectMeta` and not using the original `ObjectMeta` field, in the context of overrides. Let me dig up a bit.

internal/resource/statefulset.go
```diff
@@ -113,7 +111,7 @@ func (builder *StatefulSetBuilder) Build() (client.Object, error) {
 		// an inconsistent selector would otherwise fail forever.
 		var templateLabelOverrides map[string]string
 		if overrideSts.Spec.Template != nil && overrideSts.Spec.Template.EmbeddedObjectMeta != nil {
-			templateLabelOverrides = overrideSts.Spec.Template.EmbeddedObjectMeta.Labels
+			templateLabelOverrides = overrideSts.Spec.Template.Labels
```
Zerpet, line 114: I think there was a reason for `EmbeddedObjectMeta` and not using the original `ObjectMeta` field, in the context of overrides.

internal/resource/statefulset.go
```diff
@@ -1137,7 +1096,7 @@ func podHostNames(instance *rabbitmqv1beta1.RabbitmqCluster) string {
 	var altNames strings.Builder
 	var i int32
 	for i = range ptr.Deref(instance.Spec.Replicas, 1) {
-		altNames.WriteString(fmt.Sprintf(",%s", fmt.Sprintf("%s-%d.%s.%s", instance.ChildResourceName(stsSuffix), i, instance.ChildResourceName(headlessServiceSuffix), instance.Namespace)))
+		fmt.Fprintf(&altNames, ",%s", fmt.Sprintf("%s-%d.%s.%s", instance.ChildResourceName(stsSuffix), i, instance.ChildResourceName(headlessServiceSuffix), instance.Namespace))
```
Zerpet: This is a bit overcomplicated: `Fprintf` taking as input an `Sprintf` ? Why not just use `Fprintf` only with all the arguments of `Sprintf`? :D

Zerpet (commented):
Let me dig up the reason for an embedded object meta. The rest looks ok.

Zerpet:
My concerns regarding not referencing `EmbeddedObjectMeta` were unfounded. Only one minor efficiency comment to be addressed 🚀

internal/resource/statefulset.go
MirahImage: Indeed. I'll manually fix this

PR #2253, opened 2026-09-21
Add startup probe management path prefix when configured.

### LINKED ISSUE #2249: Startup HTTP probe ignores `management.path_prefix` causing 404s and restart loops on healthy nodes
## Describe the bug
Since v2.22.0 the default startup probe is an HTTP GET against a hardcoded path:

```go
// internal/resource/statefulset.go
return &corev1.Probe{
    HTTPGet: &corev1.HTTPGetAction{
        Path:   "/api/health/checks/reached-target-cluster-size",
        Port:   port,
        Scheme: scheme,
    },
    ...
}
```

The operator does not account for [`management.path_prefix`](https://www.rabbitmq.com/docs/management#path-prefix). When that setting is present in `spec.rabbitmq.additionalConfig`, the management API is served under the prefix, so the probe's path does not exist and returns **404**. With `failureThreshold: 30` and `periodSeconds: 10`, the kubelet kills and restarts every node about every five minutes — nodes that are fully booted, clustered, and serving AMQP traffic.

This affects anyone who serves the management UI under a path prefix, which is the standard approach when several RabbitMQ clusters share an ingress hostname.

## To Reproduce
1. Create a `RabbitmqCluster` on RabbitMQ 4.2.4+ with a path prefix:

```yaml
kind: RabbitmqCluster
metadata:
  name: rabbitmq
spec:
  replicas: 3
  rabbitmq:
    additionalConfig: |
      management.path_prefix = /rabbitmq
```

2. Let the operator roll the StatefulSet.
3. The nodes boot and cluster normally, but never pass the startup probe:

```
Warning  Unhealthy  pod/rabbitmq-server-2  Startup probe failed: HTTP probe failed with statuscode: 404
Normal   Killing    pod/rabbitmq-server-2  Container rabbitmq failed startup probe, will be restarted
```

while the same pod reports:

```
$ kubectl exec rabbitmq-server-2 -c rabbitmq -- rabbitmq-diagnostics -q check_running
RabbitMQ on node rabbit@rabbitmq-server-2.rabbitmq-nodes.rabbitmq is fully booted and running
```

**Expected behavior**
The generated startup probe targets the management API where the broker actually serves it. Either:

- the operator parses `management.path_prefix` out of `spec.rabbitmq.additionalConfig` and prepends it to the probe path; or
- the CRD gains an explicit field for it (e.g. `spec.rabbitmq.managementPathPrefix`) that feeds both the probe and any other management-API consumer.

The first is the smaller change and needs no CRD bump. The second is more explicit and avoids parsing free-form config, and would also cover `defaultUserCredentialUpdater`, which hardcodes `http://127.0.0.1:15672` and `https://$(HOSTNAME_DOMAIN):15671` in the same file — the same class of assumption, currently only reachable by Vault users.

A validating webhook warning when `management.path_prefix` is set and no override is present would also have caught this before the roll.

## Workaround

`spec.override` pins the probe to the prefixed path but is clunky as a complete override of the `startupProbe`:

```yaml
kind: RabbitmqCluster
metadata:
  name: rabbitmq
spec:
  override: # Override the startupProbe below
    statefulSet:
      spec:
        template:
          spec:
            containers:
              - name: rabbitmq
                startupProbe:
                  httpGet:
                    path: /rabbitmq/api/health/checks/reached-target-cluster-size # Add management path prefix here
                    port: management
                    scheme: HTTP
                  initialDelaySeconds: 10
                  timeoutSeconds: 5
                  periodSeconds: 10
                  failureThreshold: 30
```

The `rabbitmq.com/legacy-startup-probe` annotation also resolves it, but it's documented as a fallback for clusters below RabbitMQ 4.2.4 and relying on it for an unrelated reason seems likely to break whenever that shim is retired.
- **RabbitMQ:** 4.2.7 (management image)
- **cluster-operator:** 2.23.0 (first affected: 2.22.0)
- **Kubernetes:** v1.36.3
- **Cloud provider:** Azure / AKS
 
## Additional context
We host multiple instances of RabbitMQ behind the same hostname so the management url prefix is needed.

MirahImage:
This closes #2249


Zerpet:
Looks good overall. Let me try with a local build.

Zerpet (approved):

